PR #7, opened 2020-10-03
Changed install location to /usr/share of Discord app

mobilegmYT:


Botspot:
I agree. /usr/share is a better place to put it.

PR #8, opened 2020-10-03
Add pikiss to app list

mobilegmYT:


Botspot:
Testing complete and it works fine for me.
BTW I lost my Internet connection for most of the day - that's why I was suddenly inactive earlier.

PR #25, opened 2020-10-19
Changed MC java and MC bedrock

mobilegmYT:
MC Java- Changed install to hopefully fix the unable to launch error
Mc bedrock-  Changed from flatpak to something a little lighter


Botspot:
Let's not install libpulse0. Pulseaudio breaks Raspbian.  
Other than that it looks good.

PR #31, opened 2020-10-20
Giving credit 😉

mobilegmYT:


Botspot:
There is a problem with this: During an update, if an app's `install` script has been modified, Pi-Apps will uninstall that app, update it, and then install it again.  
However, if the `install` script has not been modified, but another file has been modified (like the description), then the app will be updated without un/installing it.  

By inserting credit in the `install` scripts, we are basically triggering a bunch of uninstalls & installs. (completely unnecessary)  
For that reason, putting credit (or other non-crucial stuff) in the `install` scripts is discouraged.

Instead, please consider giving credit in the description file, or maybe we could add a new file to each app entitled `credit`.

Botspot:
This looks good.

PR #40, opened 2020-10-29
Add todo list

mobilegmYT:
You can change these to your liking


Botspot:
There is no plans to do an app review system on Pi-Apps.

However, in the near future, I may make a second app store, that does allow reviews, and would mainly be intended for repository-based software.

Botspot:
As there is not much more I plan adding to Pi-Apps' core functionality, I'm afraid the to-do list would remain very short if I merged the PR now.  
This pull request will remain open until enough list items are added to make it worthwhile.

If you can think of some good substantial list items that could be added, feel free to edit them into this PR.

I will merge it when the time seems right.

Botspot:
Huh. Github isn't letting me edit it today. OK I'll merge this and then add my new list item I thought of.

Botspot:
Oh great, Github isn't letting me merge this because your repository was deleted. :/

mobilegmYT:
oop

PR #64, opened 2020-11-04
Make newline

mobilegmYT:


Botspot:
Good catch, but I'd rather add two space characters at the end of the line than an ugly `<br>`.

PR #69, opened 2020-11-06
Add stable and beta to shortcuts

mobilegmYT:


Botspot:
Nice idea! However, it's not important enough to push an update to everyone just for a preferential name change.

No, we'll wait until eDEX-UI has some other, more substantial update, then push both updates at once.

PR #71, opened 2020-11-06
Change the shortcut names

mobilegmYT:


Botspot:
Same as last time: we won't be pushing changes this small individually. Instead we'll wait until there's a more significant update to eDEX-UI, and then bundle this with it.

Botspot:
I'm just going to change them back to "eDEX-UI 3.0.0" and "eDEX-UI 2.2.2".

Now that both versions work well, it doesn't seem necessary to add adjectives to the menu launchers.

PR #80, opened 2020-11-09
optfine

mobilegmYT:


Botspot:
Making unnecessary changes to the install scripts is discouraged. (as Pi-Apps will reinstall the app if the install script was modified)

Also I noticed you did not add the message to `install-64`. So only 32bit users would see it. Was that on purpose?

I'm closing this. Will add the URL to the description.

PR #94, opened 2020-11-17
Update install script

oxmc:
change some text.


Botspot:
Are these changes really necessary?

If merged, 100% of Pi-Apps users would see the bothersome pop-up saying the `install` script has a update. And making changes this insignificant seems silly.

PR #95, opened 2020-11-17
Change update script

oxmc:
instead of using the autostart folder make a service file so it starts at boot and waits till it's finished.
example
```
[Unit]
Description=Pi-Apps-update-service
After=network.target
 
[Service]
WorkingDirectory=/home/pi/pi-apps/
ExecStart=/usr/bin/bash /home/pi/pi-apps/update
 
[Install]
WantedBy=multi-user.target
```

### DIFF
--- a/updater
+++ b/updater
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+#This file runs after networktarget in systemd at boot.
+
 DIRECTORY="$(readlink -f "$(dirname "$0")")"
 
 function error {


Botspot:
A systemd service file *could* work fine in theory, but the question is "Why?"  I'll consider it if there's a problem with my existing autostarted solution, which has these advantages:

- Really simple
- Does not require root permissions to add/remove
- Does not hold up the boot process
- Only runs after the desktop session is initialized

Issue #99, opened 2020-11-26
Pycharm Community Edition by JetBrains (both 32-bit & 64-bit respectively)


fabianmendes:
[for 32-bit only (version from 2018.3.7) & 64-bit updatable versions]]

Hi, again @Botspot. I want to open an app request, but mostly I would like to help to integrate it, or at least do the "dirty work" of coding, and then you could write the _easiest_ parts, I don't know. It seems to be very easy to install, there is not any problem to do so manually; but what about doing it faster and _automatic_ alongside the others Pi-apps.

**App request.** As I said, it is pretty easy to install manually, just download the version, unzip, and run _./pycharm.sh_ once.* 
Source: https://www.element14.com/community/community/raspberry-pi/blog/2019/09/12/installing-pycharm-on-raspberry-pi (really thanks to this guy). I've even already memorized the process, **also, it is not necessary to move the folder into the _/opt/_ as it says either**.

Just let me know how can I help, @Botspot. Should I create a fork or something??

Otherwise, and most important: due to having two versions for the Community Edition: the _..-2018.3.7_ version it's the only recent release available for 32-bits OS. Should I create an `install-32 | install-64` option respectively, right?

What may I do to help you for adding it to the Pi-apps?

* as you may see on the link, we only just may check that _pip_, _pip3_, and _openjdk-11-jdk_ are installed.

Botspot:
Pi-Apps uses `bash` scripts to install and uninstall its apps.

A bash script is really just a text file full of **terminal commands**.

So if you know the terminal commands necessary to install your app, then simply paste all the commands into a text file and that's about it.

Pi-Apps has an automated App maker wizard. Launch it by going to Menu -> Preferences -> Pi-Apps Settings. Then click Create app.  
As I said, the wizard should guide you through the whole process, and saves your work as you go.

Once the app is in it's early stages, feel free to send it to me by compressing the relevant folder in `~/pi-apps/apps` as a .ZIP file, then attach that file to this message box.

We'll work together to make the app, and when it's done, I think we'll add it to the official Pi-Apps repository.

fabianmendes:
how can I verify on a system if it's 32-bit or 64-bit? so, then if it is 32-bit do the next process, or if it is 64-bit do this thing. Because I have to specify which bit is the system to download a file or another from JetBrain's website page.

Botspot:
[...]

PR #101, opened 2020-11-26
Pycharm CE (created)

fabianmendes:
Scripts are created to install Pycharm Community Edition (both 32 and 64 bits) with the Pi-Apps's wizard.
#99 [**App request**]


Botspot:
We're going to use the zip file on #99 for now. I've found it's easier to integrate changes and test them before it goes live on the repository this way.

PR #103, opened 2020-11-27
Update description

mobilegmYT:


Botspot:
Good idea, but why not just tell the user directly what button to click, instead of a link?

Botspot:
I'm going to give the config file thing a shot.

PR #104, opened 2020-11-27
Update install-32

mobilegmYT:


Botspot:
Hmm, good idea, but poor execution.

I wonder if we could spoof a config file like we do for MC Java?

PR #130, opened 2020-12-09
Fixes to pigro and puffin internet 

Itai-Nelken:
### problem:
puffin wasn't uninstalling, issue: the package name was `puffin-internet-terminal-demo` and not `puffin-internet-terminal` and even though that seems to be taken into account in the uninstall script, it didn't work. 
### fix: 
splitting the `apt purge` command to 2 commands one after the other.

### problem:
pigro installer not making all scripts executable.
### fix:
adding the needed lines to install script.
### problem:
the uninstaller script trying to remove menu shortcut from `/usr/share/applications` instead of `~/.local/share/applications`.
### fix:
change line in uninstall script.

### DIFF
--- a/apps/PiGro/uninstall
+++ b/apps/PiGro/uninstall
@@ -8,7 +8,7 @@ function error {
 }
 
 #if your app installs any packages, keep this command here so those packages will be removed.
-"${DIRECTORY}/purge-installed"neofetch"$(dirname "$0")" || exit 1
+"${DIRECTORY}/purge-installed" "neofetch" "$(dirname "$0")" || exit 1
 
 rm -rf ~/PiGro-Aid-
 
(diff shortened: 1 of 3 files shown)


apps/PiGro/uninstall
```diff
@@ -8,8 +8,9 @@ function error {
 }
 
 #if your app installs any packages, keep this command here so those packages will be removed.
-"${DIRECTORY}/purge-installed" "$(dirname "$0")" || exit 1
+"${DIRECTORY}/purge-installed" "neofetch" "$(dirname "$0")" || exit 1
```
Botspot, line 11: purge-installed does not work like this.

Botspot:
Okay I think everything is good now. Thanks, @Itai-Nelken!

apps/PiGro/uninstall
Itai-Nelken: oops

PR #132, opened 2020-12-11
Retropie desktop app

mobilegmYT:
@Botspot

### DIFF
--- a/apps/Retropie/install
+++ b/apps/Retropie/install
@@ -20,3 +20,15 @@ sudo ./retropie_setup.sh
 
 #Delete setup script
 sudo rm -r ~/RetroPie-Setup
+
+#Desktop shortcut
+echo '[Desktop Entry]
+Type=Application
+Name=RetroPie
+GenericName=RetroPie
+Comment=EmulationStation
+Icon=/home/pi/.icons/retropie.png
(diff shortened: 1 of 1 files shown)


fabianmendes (changes requested):
I think this is not correct, or it actually is?

apps/Retropie/install
```diff
@@ -20,3 +20,15 @@ sudo ./retropie_setup.sh
...
+echo '[Desktop Entry]
+Type=Application
+Name=RetroPie
+GenericName=RetroPie
+Comment=EmulationStation
+Icon=/home/pi/.icons/retropie.png
+Exec=sh -c "sudo chvt 1 && sudo ttyecho -n /dev/tty1 'emulationstation; sudo chvt 7'"
+Categories=Emulator;Game;
+Keywords=Emulator
+Terminal=true' > ~/.local/share/applications/retropie.desktop
```
fabianmendes, line 34: =false

Botspot:
> I think this is not correct, or it actually is?

It's not correct. You're right.

mobilegmYT:
🤷‍♂️ I found this from TwisterOS, so it should work on rpi os. Might be wrong though.

Botspot:
The syntax used for echoing was not correct. In any case, this PR is being closed because somebody will get 5 dollars for improving the RetroPie app.

PR #138, opened 2020-12-16
Retro2png

mobilegmYT:


Botspot:
I thought we discussed this in the discord server.

We don't need two apps that do the same thing. We don't want to change the name from "raspi2png" to something even more obscure.

mobilegmYT:
Whoops lol forgot about it.

PR #142, opened 2020-12-18
Update MC Bedrock install with latest one

mobilegmYT:


Botspot:
Can you assure me you've tested this?

mobilegmYT:
uh....... I'll be back. I got these from phoenix so I assume this should work.

mobilegmYT:
Yep, it's the same version as the latest update and works nicely.

Botspot:
Okay then. Thanks.

PR #155, opened 2020-12-24
Add Minecraft Bedrock 64 bit install script

CactiChameleon9:
I haven't tested this yet (I haven't got piOS installed on a 64bit pi) but it should work fine - I based it a lot off the 32bit one:
- Added the architecture
- Changed the dependencies to the ones mentioned [here](https://github.com/ChristopherHX/linux-packaging-scripts/releases)
- Changed the download url

I was concerned about my use of `sudo` for adding the armhf architecture, but it seems that there is no "best practice"  implemented here yet and the 32 bit script already uses it to move the app image to `/usr/bin`


Botspot:
Let's give it a try. ;)

PR #160, opened 2020-12-25
box86 improvement/bug fix

Itai-Nelken:
every time I tried to install box86 (on a new RPiOS 32bit install), it got stuck.
the error was a typo: `make -j24` instead of `make -j4`.


Botspot:
I think I remember the `-j24` (24 compiling threads) was slightly faster than `-j4` on a fresh install.

Actually, thanks for drawing my attention to Box86 - I've been meaning to switch to [box86-updater](https://github.com/Botspot/box86-updater). This method uses pre-compiled binaries - taking seconds to install than hours.

Botspot:
Box86 has been updated to use my box86-updater repository.

PR #166, opened 2020-12-27
Update install-64

fabianmendes:
[official version update] Otherwise, the comment line error was fixed; I hadn't delete nor modified "2018" on this 64bit installation script (the 2018 version is only for 32bit).


Botspot:
Awesome.

PR #179, opened 2020-12-29
Deleted install file in Doom 3 app

techcoder20:
Deleted install file because there are two install files and only one of them works because the deleted install file only runs the install script from the RPIDoom3Installer github repository and thats not enough as we need to clone the repository which has the deb files required during installation. The install-32 file works perfectly as it clones the repo.


Botspot:
Thanks for spotting that.

PR #198, opened 2021-01-04
Add lighpad to TwisterOS hidelist

mobilegmYT:


Botspot:
This will work fine for installations that freshly install Pi-Apps on TwisterOS, but for everyone who already has Pi-Apps, this code (the install script) won't be run.

I will merge this pull request, then add a trigger to the `gui` script, that runs the `install` script, if 'Lightpad' is missing from the hidelist and if the OS is Twister.

PR #205, opened 2021-01-05
Change Bedrock uninstall

mobilegmYT:
just in case a user wants all data gone. I added a failsafe and also tested the code.

### DIFF
--- a/apps/Minecraft Bedrock/uninstall
+++ b/apps/Minecraft Bedrock/uninstall
@@ -8,3 +8,32 @@ function error {
 
 #Remove app
 sudo rm /usr/bin/MCBedrock.AppImage
+
+#Dialog
+HEIGHT=15
+WIDTH=40
+CHOICE_HEIGHT=4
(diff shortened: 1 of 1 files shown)


Botspot:
During an update, we don't want a user-input holding the update process up if we can't help it. So let's work out a middle-ground.

@mobilegmYT, could you figure out exactly which directory is responsible for holding the minecraft world and configuration? Let's keep that. Everything else non-essential can be removed without asking.

mobilegmYT:
> During an update, we don't want a question holding the process up if we can't help it. So let's work out a middle-ground.
> 
> @mobilegmYT, could you figure out exactly which directory is responsible for holding the minecraft world and configuration? Let's keep that. Everything else non-essential can be removed without asking.

Alright, only configs and game files are kept now.

Botspot:
Looks like we still need to get rid of the dialog.

Botspot:
Why is the dialog still there?

mobilegmYT:
> Why is the dialog still there?

wdym?

PR #208, opened 2021-01-05
MC Bedrock update 64-bit version

mobilegmYT:


Botspot:
How do we just get rid of the uninstall file from this commit altogether?

mobilegmYT:
Yes

PR #219, opened 2021-01-07
Configured error function to print text in red

techcoder20:


Botspot:
Doesn't the error function already print in red?

techcoder20:
Oh I don't remember that sry

PR #222, opened 2021-01-08
Updated new link and added alias for AIW

spectrumgamer75:
This update for AIW adds the new download link for the script along with adding an alias for launching directly from the terminal.


Botspot:
Make sure the uninstall script reverts all changes, including the alias.

spectrumgamer75:
That is being worked on, will add that soon in the uninstall script too.

Botspot:
OK. Then I can wait.

spectrumgamer75:
Have added that now.

PR #232, opened 2021-01-11
Add warning about signing in with google

mobilegmYT:
Same as https://github.com/Botspot/pi-apps/pull/231 but with zoom.


Botspot:
Sorry, can't accept this PR.

Updating Zoom now will get everyone's hopes up when they see the available update, when in fact nothing has really changed.

Send me a PR for adding that to the description.

Issue #224, opened 2021-01-09
Minecraft Java Install Error


AmBruhIdot:
Pi 4 8GB-BerryBoot RPI OS (64bit)
Terminal: [https://pastebin.com/HP9HbdPf](url)

Before this installation when trying to install terminal something along the lines of installation failed because /home/pi/Minecraft already existed so I deleted that folder, pressed uninstall in pi apps, and tried installing again and got those errors in the Terminal. If there is anything you can help me with, please let me know!

AmBruhIdot:
I have a Mojang account* forgot to add that

mobilegmYT:
@AmBruhIdot Try again now.

AmBruhIdot:
I'll try again when I get home, which will be about in 4 hours.

Botspot:
Closed due to inactivity.

Reply back and I can re-open it.

Issue #173, opened 2020-12-28
minecraft java install error


premkubade:
when i try to install minecraft java it gives me a error that it failed to install minecraft java launcher

Itai-Nelken:
could you send the terminal output or a screenshot of the error?

Botspot:
> when i try to install minecraft java it gives me a error that it failed to install minecraft java launcher

What version of Minecraft Java were you trying to install?

mobilegmYT:
@Botspot I guess close this issue due to inactivity???

premkubade:
Installing Minecraft Java with install-64 script
mkdir: cannot create directory ‘/home/pi/Minecraft’: File exists
Failed to install the launcher!

Failed to install Minecraft Java with install-64 script!

Close this window to exit.

premkubade:
this error i found while installing

Botspot:
> this error i found while installing

It looks like it was partially installed prior to you installing it this time. Please try uninstalling it, then install it again.

premkubade:
i tried it 3 times uninstalling and installing but the result is same

Botspot:
> i tried it 3 times uninstalling and installing but the result is same

Well then try manually deleting `/home/pi/Minecraft`.

premkubade:
ok i will try it

premkubade:
Setup 6/9
Setup 7/9
Extracting java ...
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package openjdk-11-jdk is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'openjdk-11-jdk' has no installation candidate
Failed to install the launcher!

Failed to install Minecraft Java with install-64 script!

Close this window to exit.

Botspot:
> ```
> Setup 6/9
> Setup 7/9
> Extracting java ...
> Reading package lists... Done
> Building dependency tree
> Reading state information... Done
> Package openjdk-11-jdk is not available, but is referred to by another package.
> This may mean that the package is missing, has been obsoleted, or
> is only available from another source
> 
> E: Package 'openjdk-11-jdk' has no installation candidate
> Failed to install the launcher!
> 
> Failed to install Minecraft Java with install-64 script!
> 
> Close this window to exit.
> ```

Okay that is quite weird. Try searching for the java JDK yourself:
```
apt search openjdk
```
(no need for `sudo` for `apt search`)

[...]

PR #233, opened 2021-01-11
Remove useless install file

mobilegmYT:
THis removes the install script which isn't needed after a install. Forgot to add it into to last pull request so merge it when another fix is issued.

### DIFF
--- a/apps/Minecraft Java/install-32
+++ b/apps/Minecraft Java/install-32
@@ -42,6 +42,8 @@ echo '{
   }
 }' > $HOME/.minecraft/launcher_profiles.json
 
+sudo rm -rf ~/Minecraft
+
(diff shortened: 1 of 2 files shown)


Botspot:
This probably will fix MC Bedrock install issues.

IIRC, MC Bedrock git cloned a repo to ~/Minecraft, and for users who had already installed MC Java, it would fail to git clone.

mobilegmYT:
> This probably will fix MC Bedrock install issues.
> 
> IIRC, MC Bedrock git cloned a repo to ~/Minecraft, and for users who had already installed MC Java, it would fail to git clone.

Ah ok, but I thought MC Bedrock was just a appimage???

Botspot:
These.  
https://github.com/Botspot/pi-apps/issues/173  
https://github.com/Botspot/pi-apps/issues/224

Oh OK, so Java MC, but my point still stands: This should fix both of those issues.

PR #239, opened 2021-01-14
Install mcpi-docker (WIP)

mobilegmYT:


Botspot:
Looks like you added a bunch of stuff in addition to running the `script.deb.sh` script.

Why? I was under the impression that `script.deb.sh` took care of everything.

mobilegmYT:
> Looks like you added a bunch of stuff in addition to running the `script.deb.sh` script.
> 
> Why? I was under the impression that `script.deb.sh` took care of everything.

That one only installs mcpil-r which is a launcher. All the mods happen in mcpi-docker, which isn't installed as a dependency since it isn't in the repos.

PR #251, opened 2021-01-16
tor Browser: Update install-32

fabianmendes:
2021-01-14 version update


fabianmendes (commented):
`wget` is missing!

fabianmendes (commented):
wget missing

PR #261, opened 2021-01-20
Removed uninstall snap line

techcoder20:
If we uninstall snapd the user won't be able to use any apps he or she installed with snap store


Botspot:
You have a good point.

I've come up with a different way of solving the issue though, and so am closing this PR.

PR #269, opened 2021-01-23
Add Minecraft apps to MC folder

mobilegmYT:
Lots of new MC apps to be planned to be added, better if they get separated into their own catagories.

### DIFF
--- a/data/categories/structure
+++ b/data/categories/structure
@@ -27,8 +27,8 @@ Intellij IDEA|Editors
 jGRASP IDE|Editors
 LibrePCB|Editors
 Lightpad|Eyecandy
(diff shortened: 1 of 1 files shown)


Botspot:
While such an action would be appropriate for a lesser-known app, I think for Minecraft it's best to leave it as-is.

At least 1/3 of the user-base got pi-apps to use Minecraft. I think hiding it away under another layer of folders will do more harm than good. Everyone would have to click one more time to reach it, and some users may miss it entirely. (Because folders don't have icons)

mobilegmYT:
Hrm true... maybe add icons soon then? 😉 Maybe in the future when the games section gets more crowded?

Issue #272, opened 2021-01-24
Snap shows twice on menu


Chulamin:
Hi. I tried a couple of times to add and delete Snap with Pi-Apps but it always creates two menu items under system using Twister OS.

Botspot:
I had to make a workaround for Raspberry Pi OS, for the menu buttons to show up at all.

That workaround wasn't needed on TwisterOS, as Snap works out-of-the box there apparently.  
Thanks for letting me know!

techcoder20:
@Botspot I proposed a fix here https://github.com/Botspot/pi-apps/pull/273

Botspot:
@Chulamin Could you send a screenshot of your menu that contains two Snap Store menu icons?

I haven't been able to repeat what you're reporting.

Chulamin:
________________________________

Botspot:
[...]

PR #273, opened 2021-01-25
Creating Symlink Only If Its Raspberry Pi OS

techcoder20:
According to this issue https://github.com/Botspot/pi-apps/issues/272 There is no need to create a symlink if its twister os. So I modified to code to only create a symlink if its raspberry pi os. I am checking if the os is twister os by checking if the binary twistver is there as its builtin in twiseros


Botspot:
We have yet to prove if Snap Store really shows twice on TwisterOS or not.

It would be a shame if some Twister users didn't see Snap Store at all if this PR was merged.

techcoder20:
Your right. Needs more testing.

PR #293, opened 2021-01-28
Hide cordless

mobilegmYT:


Botspot:
Instead of this which probably wouldn't work anyway, I'll add a line to the `gui` script which simply deletes `$DIRECTORY/apps/Cordless`.

All good?

mobilegmYT:
Sure.

Botspot:
Done.

PR #314, opened 2021-02-02
Option to use precompiled 3g/1g kernel + small update/fix to wine install script

Itai-Nelken:
I just tested wine on a pretty fresh install of RPiOS and when testing notepad++ I got the bug that wine doesn't report the virtual C: drive (it shows `%System Drive%`), that is fixed by running `wine wineboot -u` (u = update) after the first wineboot, if that bug doesnt happen it won't break anything.


Itai-Nelken:
small note: it will ask to install wine mono and gecko if the first wineboot doesn't do that.

Itai-Nelken:
I added the option to use a precompiled kernel I'll compile every month or so.

PR #321, opened 2021-02-03
Update to add warning

mobilegmYT:


Botspot:
As Fake KMS is the default on all systems, I don't think this message is necessary.

Especially with it waiting for user input.

Now, what would be good was if the script detected if OpenGL was turned off and warned the user. That would be good.

PR #322, opened 2021-02-03
wine uninstall script removes all desktop shortcuts

Itai-Nelken:
before this it tried to remove `~/.local/share/applications/winecfg.desktop` that doesn't even exist.


Botspot:
Thanks for catching this. :)

PR #323, opened 2021-02-04
Etcher install scripts always download the latest version.

Itai-Nelken:
instead of having to update the link every time I compile and package a new version.
@Botspot don't merge yet,  I still want to be 100% sure it works.


Botspot:
I'm not sure what I think of this.

On the one hand, you're right, it will slightly reduce the workload whenever Etcher receives an update. But on the other hand, Etcher would never auto-update once installed.

Auto-updating in this manner basically circumvents Pi-Apps's updating approach.

I think the occasional minute or two taken to update a url in the install script, consequently triggering an update for everyone, is the best way to go.

Itai-Nelken:
ok, I have no problem changing the link. I just thought it might be better seeing as thats the way with the discord app.

Botspot:
> ok, I have no problem changing the link. I just thought it might be better seeing as thats the way with the discord app.

That's because spacingbat3 updated discord so often... the updates for it got annoyingly frequent.

There's a balance, you've got to admit. ;)

Itai-Nelken:
>There's a balance, you've got to admit. ;)

you are right.

PR #332, opened 2021-02-07
change the install and uninstall script.

oxmc:


oxmc:
@Botspot update the install,uninstall and settings.

Botspot:
Reasons why I hesitate merging this:

- I think most people will run Pi-Apps from the main menu or from the terminal, not from the desktop.
- What if there is no line to begin with in libfm.conf? In that case, your sed lines wouldn't do anything.
- No, I will not allow the README's official install instructions to download and run stuff from your repository.
  - I may need to change the script later to fix something. But I wouldn't be able to because it's your script on your repository.

oxmc:
i have changed the README.md file to the same as yours, the new app was also added.

Botspot:
I've decided not to add this.

The risks are too high, the benefits too small, and too few people would notice.  
I'd happily add this if only Pi-Apps's desktop launcher was single-click, but making all scripts and all desktop files in all file managers single-click is a serious hazard.

PR #337, opened 2021-02-08
"improvement" to installation instructions

Itai-Nelken:
I saw a few people that didn't understand they have to press enter to run the install script (`~/pi-apps/install` line) so I put the git clone and the install command in a single line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,7 @@ Pi-Apps does exactly this - distributing over 50 useful app install scripts that
 ## To install Pi-Apps
 Make sure `git` is installed with `sudo apt install git`, then run:
 ```
-git clone https://github.com/Botspot/pi-apps
-~/pi-apps/install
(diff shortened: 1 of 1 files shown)


Botspot:
Fascinating.

I'll have to decide if a one-liner that's less understandable is better than two distinct lines that are a lot more self-explanatory.

Itai-Nelken:
maybe make the one line install and then use something like this (click on it):
<details>
<summary> for advanced users </summary>

### the 2 commands run for installation:
```bash
git clone https://github.com/botspot/pi-apps.git
~/pi-apps/install
```
</details>

the code:
```bash
<details>
<summary> Click me! </summary>

new line above and bellow are important

</details>
```

Botspot:
Thank you for the idea.

I've chosen to do something similar with that little drop-down thing you showed me. Check the readme.

PR #348, opened 2021-02-12
Fix broken uninstall

mobilegmYT:


Botspot:
I've been meaning to ask tbr about that error. Thanks for catching it.

PR #352, opened 2021-02-13
Replace repo signing code

mobilegmYT:
The ubuntu signing stuff just doesn't work for some people.


Botspot:
Glad I could help. ;)

PR #330, opened 2021-02-06
Get MCPI Modded to update.

RaspberryPiNews:
This is just a useless variable that makes pi-apps reinstall it and update to the latest version with a reinstall.

NOTE: The latest version didn't start for me, so wait until I say so before mergining

PR #353, opened 2021-02-13
Fix Some MCPI Stuff

TheBrokenRail:
- ~The ``MCPI_REBORN_VERSION`` variable wasn't used anywhere, so I removed it~
- I piped all output (including ``stderr``) from ``sudo apt-get remove -y minecraft-pi >/dev/null || :`` commands to ``/dev/null``, so people would stop reporting errors about ``E: Unable to locate package minecraft-pi-native``
- I removed the regex stuff from the ``gpg`` code because the ``buster-backports`` signing keys are never going to change and fixed the original problem (``gpg: keyserver receive failed: Server indicated a failure``) according to https://unix.stackexchange.com/a/399091

### DIFF
--- a/apps/Minecraft Pi (Modded)/install
+++ b/apps/Minecraft Pi (Modded)/install
@@ -4,6 +4,9 @@
 ## Config
 ##
 
+# Used To Trigger Updates
+MCPI_REBORN_VERSION='build_32'
+# MCPIL Version
 MCPIL_VERSION='0.1.8'
 
 set -e
--- a/apps/Minecraft Pi (Modded)/install
+++ b/apps/Minecraft Pi (Modded)/install
@@ -4,7 +4,6 @@
 ## Config
 ##
 
-MCPI_REBORN_VERSION='build_28'
 MCPIL_VERSION='0.1.8'
 
 set -e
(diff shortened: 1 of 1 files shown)


TheBrokenRail:
BTW, running a ``sed`` script on a command that specifically says its CLI is unstable is IMO not a good idea.

Botspot:
> The MCPI_REBORN_VERSION variable wasn't used anywhere, so I removed it

That was [added](https://github.com/Botspot/pi-apps/pull/330#issue-568832510) to keep mcpil updated. User @mobilegmYT makes a PR that changes the variable each time mcpil is updated. This technique is used to keep everyone's version up to date.

> running a sed script on a command that specifically says its CLI is unstable is IMO not a good idea.

@mobilegmYT just [made that fix](https://github.com/Botspot/pi-apps/pull/352). Several people had issues, and that PR was merged in an attempt to fix it.  
That `sed` command line automatically signs all unsigned repositories, and is an accepted way to do it automatically.

TheBrokenRail:
> That was added to keep mcpil updated. User @mobilegmYT makes a PR that changes the variable each time mcpil is updated. This technique is used to keep everyone's version up to date.

Oh, so rather than specifying the version to download, it is just triggering an update!

> @mobilegmYT just made that fix. Several people had issues, and that PR was merged in an attempt to fix it.
That ``sed`` command line automatically signs all unsigned repositories, and is an accepted way to do it automatically.

The issue was that some people's internet didn't like the Ubuntu keyserver as it was configured. This triggered an ``gpg: keyserver receive failed: Server indicated a failure`` error. The new code inadvertently fixed this by specifying the key server differently. However, the ``sed`` is superfluous because the ``buster-backports`` signing keys aren't going to change without breaking many people's installations.

PR #367, opened 2021-02-17
Update etcher to 1.5.114

Itai-Nelken:


techcoder20:
@Itai-Nelken Why don't you change the file name to etcher_latest.deb so that you don't need to change the name every time in the pi apps repository and you just have to update your repo

Itai-Nelken:
I did something similar in #323, but Botspot said he prefers it this way because then etcher is updated for everyone so users always have the latest version instead of having the latest version as of the time they installed Etcher.

PR #386, opened 2021-02-22
eDEX-UI add arm64 appimage

Itai-Nelken:
@Botspot don't merge yet, I haven't tested if it works yet.

also there is no need to update the uninstall script because using the `-f` option with `rm` makes it not show errors (unless it doesn't have the permission to remove the file).


SpacingBat3:
> @Botspot don't merge yet, I haven't tested if it works yet.
> 
> also there is no need to update the uninstall script because using the `-f` option with `rm` makes it not show errors (unless it doesn't have the permission to remove the file).

Just make this a Draft Pull Request?

apps/eDEX-UI/install-64
```diff
@@ -0,0 +1,26 @@
...
+Comment=eDEX-UI sci-fi interface
+Exec=bash -c "\""sudo modprobe fuse; $HOME/eDEX-UI-RPi/eDEX-UI-Linux-armv7l_2.2.6.AppImage"\"" %U
+Terminal=false
+Type=Application
+Icon=$(dirname "$0")/icon-64.png
+StartupWMClass=eDEX-UI
+X-AppImage-Version=2.2.6
+Categories=System;
+X-Desktop-File-Install-Version=0.23
+X-AppImage-Comment=Generated by /tmp/.mount_eDEX-UDQYwse/AppRun" > ~/.local/share/applications/appimagekit-edex-ui.desktop
```
SpacingBat3, line 26: Just to note, AppImages has usually a Desktop file inside of them, which you should rather use that the one generated by this script.

I believe AppImage runtime supports exporting files from the AppImages like this:
```sh
<path-to-appimage> --appimage-export [file]
```

SpacingBat3 (changes requested):

apps/eDEX-UI/install-64
Itai-Nelken: so I have to find the name of the desktop file inside the appimage (if present) and then run something like this: `edex-ui-2.2.6.appimage --appimage-export name-of-desktop-shortcut.desktop`?

Itai-Nelken: @SpacingBat3 this didn't work, its says that it isn't suppported yet on that appimage.

SpacingBat3: > @SpacingBat3 this didn't work, its says that it isn't suppported yet on that appimage.

Oh, I've also checked this and it seems there's an `--appimage-extract` flag, but no `--appimage-export`.

The syntax should be the same through.

Itai-Nelken: I won't use this because I'll have to edit the .desktop file anyway.

Itai-Nelken:
@Botspot should work now, but not tested yet.

Itai-Nelken:
added by @Botspot in commit d15498cb7a3bfe9e75f2c0636fdf9a54a557e6d5.

PR #392, opened 2021-02-25
Install details and Uninstall

slashtechno:
Since /usr/local/bin is modified it means that Pi-Apps does modify files/folders outside of the home directory. I put this in the README as well as a way to uninstall. For some reason I can not inline pi-apps in line 26.

**Also I believe the README should be shortened**


README.md
```diff
@@ -22,10 +22,11 @@ wget -qO- https://raw.githubusercontent.com/Botspot/pi-apps/master/install | bas
 git clone https://github.com/Botspot/pi-apps
 ~/pi-apps/install
 ```
-
 </details>
+The install script ensures YAD is installed, creates two menu buttons, and an autostarted updater. Nothing is modified outside your home directory execpt one file to allow `pi-apps` in a terminal.
+
+To uninstall Pi-Apps run `sudo rm //usr/local/bin/pi-apps && rm -r //home/pi/pi-apps`
```
Itai-Nelken, line 28: pi-apps has a uninstall script. `~/pi-apps/uninstall`.

README.md
```diff
@@ -22,10 +22,11 @@ wget -qO- https://raw.githubusercontent.com/Botspot/pi-apps/master/install | bas
 git clone https://github.com/Botspot/pi-apps
 ~/pi-apps/install
 ```
-
 </details>
+The install script ensures YAD is installed, creates two menu buttons, and an autostarted updater. Nothing is modified outside your home directory execpt one file to allow `pi-apps` in a terminal.
```
Itai-Nelken, line 26:
```suggestion
The install script ensures YAD is installed, creates two menu buttons, an autostarted updater, and a file (`/usr/local/bin/pi-apps`) that allows pi-apps to be run from terminal. Nothing is modified outside your home directory.
```

Itai-Nelken (changes requested):
I changed the part about /usr/local/bin/pi-apps:
as I understand it, the 'no files are modified outside of your home directory' part means no apps modify files outside of your home directory other then things like menu buttons and executables that go in /usr/bin and /usr/local/bin etc.

Pi Apps also has a uninstall script.

Itai-Nelken:
also I think the Readme isn't to long, it explains everything you need to know about pi-apps, maybe something better would be to add all the 'how it works' part in a different file or a dropdown details html thing like with the 'manual installation'.
but I still think its unnecessary to do any of those things.

README.md
slashtechno: I agree,

README.md
slashtechno: Oh okay.

slashtechno:
@Itai-Nelken I agree with you. I think that Pi-Apps needs a Github Pages site.

Botspot:
> Pi Apps also has a uninstall script.

Speaking of that, I just improved it so that it would remove the terminal command, and to notify the user that `~/pi-apps` still exists.

Issue #405, opened 2021-03-02
Tor Browser Needs updated on arm64


craigevil:
The newest is 10.0.11.
The correct link in the script should be :
wget https://sourceforge.net/projects/tor-browser-ports/files/10.0.11-arm64/tor-browser-linux-arm64-10.0.11_en-US.tar.xz/download

Botspot:
Thank you. We'll use the pull request RPiNews opened. #406

PR #406, opened 2021-03-02
Update tor to 10.0.11

mobilegmYT:
https://github.com/Botspot/pi-apps/issues/405


Botspot:
Glad I caught that. Looks like you didn't test the code. This is what multiple security layers is for: weeding out simple mistakes.

See it now?

PR #414, opened 2021-03-05
Force MCPI update to build 50

mobilegmYT:


Botspot:
How many of these changes are actually noticeable by end-users?

mobilegmYT:
> How many of these changes are actually noticeable by end-users?

What do you mean?

mobilegmYT:
Well he removed and added some important code, and removed something to do with root, but still better to update than to not 🤷‍♂️

Botspot:
> > How many of these changes are actually noticeable by end-users?
> 
> What do you mean?

I mean, what changes have been made between version 48 and 50? Significant changes, or severe bugs fixed, necessitating a Pi-Apps update?

Or one mere correction to a typo?

Botspot:
> Well he removed and added some important code, and removed something to do with root, but still better to update than to not

Okay. I'll push it then.

PR #422, opened 2021-03-05
Fixed bug in Pi-Apps Terminal Plugin where it wont work with apps which have space in their name

techcoder20:
The reason for doing this is that if we run a command like `pi-apps install 'Pi-Apps Terminal Plugin'` where the app name has spaces it wont work unless $@ is in ""


Botspot:
Thanks. I was thinking about that, but forgot if `$@` was the correct syntax, or `"$@"`.

PR #438, opened 2021-03-09
Echo debian version on MCPI unsupported version

mobilegmYT:


Botspot:
Let's wait until there's a necessary MCPI update, before pushing this. Pushing it now would cause an unnecessary reinstall for everyone.

mobilegmYT (commented):
Well yes, but no one checks the git repo lol

mobilegmYT:
> Let's wait until there's a necessary MCPI update, before pushing this. Pushing it now would cause an unnecessary reinstall for everyone.

Well actually, I just checked the jenkins and TBR added whitelist support to servers and clients, as well as a FPS counter flag 😉

PR #441, opened 2021-03-09
Snapdrop: Links invalid, also add install-64

chunky-milk:
@Botspot the install scripts contain broken links, I've fixed them, and also added install-64


Botspot:
Strange - I didn't encounter any broken links when testing. It worked fine for me...

PR #449, opened 2021-03-11
Force reinstall for mesa fix on MC Java

mobilegmYT:
Basically this forces a reinstall of MC java, which pulls the new desktop shortcut from https://raw.githubusercontent.com/mobilegmYT/pi-apps-resources/main/Minecraft/minecraft.desktop. It forces OpenGL 3.0, which a lot of people say gives Minecraft and other apps better performance. I use this override on my own install personally.

### DIFF
--- a/apps/Minecraft Java/install-32
+++ b/apps/Minecraft Java/install-32
@@ -7,11 +7,6 @@ function error {
   exit 1
 }
 
-#use the error function often!
-#If a certain command is necessary for installation to continue, then add this to the end of it:
-# || error 'reason'
-#example below:
-
 #Download java and launcher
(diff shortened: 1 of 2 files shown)


Botspot:
All I see is the deletion of comments. What part of the script actually pulls in your new .desktop file?

mobilegmYT:
> All I see is the deletion of comments. What part of the script actually pulls in your new .desktop file?

The desktop file is pulled from my git repo, which I updated https://raw.githubusercontent.com/mobilegmYT/pi-apps-resources/main/Minecraft/minecraft.desktop

Botspot:
Oh, I see.

Well, this may be okay, but triggering an update now would probably cause a bunch of people to get their hopes up. What do you suggest we do about that?

mobilegmYT:
> Oh, I see.
> 
> Well, this may be okay, but triggering an update now would probably cause a bunch of people to get their hopes up. What do you suggest we do about that?

Hm, I guess we could wait for ATLauncher to be done. I sent you a dm in discord a while ago about changes to the atlauncher script

PR #454, opened 2021-03-12
Add Heroes of Might and Magic 2 game

dimag0g:
This is another classic game which should be playable nicely on the Pi.
The shareware version has only one map, but here are lots of sites dedicated to old games where the full game can be downloaded for free. Not to mention GOG, where the game is sold for about $10.

[Heroes2.zip](https://github.com/Botspot/pi-apps/files/6132771/Heroes2.zip)


mobilegmYT:
@dimag0g Could you fix the problem of the conflicting file?

Itai-Nelken:
fixed the conflict.

mobilegmYT:
@Botspot

PR #456, opened 2021-03-13
Change nodejs to use nvm and not directly install from apt

mobilegmYT:
I've noticed a error with the apt repo nodejs, which is that a lot of apps have root issues while running. Switching to using the nvm manager fixes those issues since it does root magic while installing node.


Botspot:
> since it does root magic

Everything is awesome! (with magic)

PR #461, opened 2021-03-14
Terminal Fix for PyCharm

techcoder20:
The local terminal in pycharm doesn't work by default. I added some code which will fix that.


Botspot:
Has this been tested for both 64-bit and 32-bit?

techcoder20:
Yes. I am not sure why but arm64 doesn't need this fix while arm32 does need so I just made one script.

PR #464, opened 2021-03-14
`armhf` instead of `arm64` in OBS install-64

Itai-Nelken:


Botspot:
It appears the only changes you made were cosmetic, ie., in the `error` functions. Is that conclusion correct, or does the current script fail to work for some reason?

Itai-Nelken:
> It appears the only changes you made were cosmetic, ie., in the `error` functions. Is that conclusion correct, or does the current script fail to work for some reason?

the current script works just fine as far as I know, I didn't ever try it.

PR #479, opened 2021-03-19
Changed the description to VSCode

CleanMachine1:
I created an issue a little while ago regarding removing VSCode, since its been discussed and I am satisfied, I have just edited the description to let the user know that if they are using RPi-OS, they don't need to install it from MS servers, and rather just download from the raspberry pi repos 

Small change but a piece of useful information for someone who hasn't kept up with raspberry pi news


CleanMachine1:
Thank you very much, your wording is better than mine!

PR #488, opened 2021-03-20
discord not auto-get-newest-version to enable pi-apps updates

Itai-Nelken:


Botspot:
Our of curiosity, why do you keep putting `cd $HOME` in your scripts?

Is it necessary in some cases?

Itai-Nelken:
> Our of curiosity, why do you keep putting `cd $HOME` in your scripts?
> 
> Is it necessary in some cases?

just to make sure I know where the deb/files will be. I know its actually not needed because it doesn't matter where the script is run from, the deb will be downloaded, installed, and removed. but I just like to be extra sure that everything will work.

Botspot:
> > Our of curiosity, why do you keep putting `cd $HOME` in your scripts?
> > Is it necessary in some cases?
> 
> just to make sure I know where the deb/files will be. I know its actually not needed because it doesn't matter where the script is run from, the deb will be downloaded, installed, and removed. but I just like to be extra sure that everything will work.

[Line 139](https://github.com/Botspot/pi-apps/blob/master/manage#L139) of `manage` takes care of that for you.

But if you still feel more "secure" by explicitly specifying `cd $HOME`, go ahead.

PR #494, opened 2021-03-21
[WACUP] install wine only if not TwisterOS and not installed by pi-apps

Itai-Nelken:
# not tested, but should work

### DIFF
--- a/apps/WACUP (new WinAmp)/install-32
+++ b/apps/WACUP (new WinAmp)/install-32
@@ -7,7 +7,10 @@ function error {
   exit 1
 }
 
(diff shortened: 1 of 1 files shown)


Botspot:
I think it would be more straightforward if pi-apps changed the `status` of Wine if TwisterOS.

Botspot:
> I think it would be more straightforward if pi-apps changed the `status` of Wine if TwisterOS.

Should be implemented now with commit 762d526a1aa7a2e6dff5d5f7a221dd59cd704879.  
Untested.

PR #506, opened 2021-03-24
Loads of changes to descriptions and credit files

CleanMachine1:
I have worked for around 45 mins to try and standardise the files:

Capitalising first words
Lines when lines get too long/unrelated lines
And much more


Botspot:
Wow.
Looks like somebody has a taste for good grammar! Thanks!

I've made a couple tweaks in a few cases where you missed something, or where I felt that changing the sentence back would display better.

CleanMachine1:
The exclamation point should be after the (text).

Botspot:
> The exclamation point should be after the (text).

Yes I know: grammatically that's correct, but in this case the text was being wrapped. Additionally, if you read it aloud, the wrong part of the sentence is emphasized in my opinion.  
This seemed like a case where it's technically wrong, but to the average user, it improves the experience. Because in the end, that's what matters.

CleanMachine1:
Also I like perfect documentation, its annoying to read something that doesn't make sense!

Issue #171, opened 2020-12-27
Snap Store


Botspot:
I've heard mention of there being a GUI app store for `snap`. It would be cool to install this on RPi, and then add it to Pi-Apps.

techcoder20:
It is possible to install it on RPi but it might be a little hard as you need to reboot after installing snap so once we install snap with apt we need to reboot then only we can install snap store with snap.

Itai-Nelken:
I always installed snap this way:
`sudo apt install snap`
sudo snap install snap-store`
reboot
 
never had any issues doing it like that.
but it seems canonical have changed the way to install it:
https://snapcraft.io/docs/installing-snap-on-raspbian
https://snapcraft.io/docs/installing-snap-store-app

as @techcoder20 said, you have to reboot after installing snap (I'm not sure it's necessary after installing `snap`, maybe make the installer to ask at the end of installation if user wants to reboot) and now you have to install `core` snap package.

here are the instructions from the 2 links above:
```
sudo apt update
sudo apt install snapd
```
reboot here
```
sudo snap install core

# install snap-store
sudo snap install snap-store
```
**A VERY IMPORTANT NOTE!**
don't install `gnome-software`, it conflicts with add/remove software (`pi-package`) and anyway isn't as useful for a lot of people (I think?).
if you want to install flatpak apps with a gui app store, than try to use this: https://gitlab.gnome.org/haecker-felix/souk (never worked for me though).

Itai-Nelken:
I'll try installing it and see if a reboot is really needed after installing `snapd`

chunky-milk:
Just FYI, in order to get the Snap applications to appear in the menu, you may have to use this command: 
`ln -st ~/.local/share/applications /var/lib/snapd/desktop/applications/*.desktop `
due to the desktop files being misplaced.

Itai-Nelken:
@chunky-milk I never had any problem with menu shortcuts not appearing, and when it did happen a reboot fixed it.

chunky-milk:
Awesome! That must mean they fixed the bug then!

Itai-Nelken:
I just installed snap-store:
```sh
sudo apt update
sudo apt install snapd
sudo snap install core
sudo snap install snap-store
```
here I rebooted.

I can confirm that until you reboot, the menu shortcut for snap-store doesn't appear.
[...]

PR #516, opened 2021-03-30
why so complicated

srini-pro:
just
sudo apt update
 sudo apt install snap
sudo snap install snap-store


Itai-Nelken:
the snap core is needed, and a restart is needed, so @techcoder20 found a way around it.
read #171

techcoder20:
There were also a problem where the desktop entry would appear in Twister OS and not in Raspberry Pi OS, so botspot suggested using a symlink between two folders which resulted in more code

Botspot:
> why so complicated

Well, now you know! This is how people get smart, by asking questions. Thanks for asking. :)

srini-pro:
thanks a lot

PR #533, opened 2021-04-04
Changes to README.md - Privacy - Piping to Bash and Analytics 

CleanMachine1:
I personally feel its wrong to pipe code straight to bash, therefore I have added a statement about it, and given more of a reason to use the manual installation, my reasoning for this

Also fixed the word Autostarted as it should be auto-started 

Small commit, but still a useful one

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,23 +15,15 @@ Introducing **Pi-Apps**, basically a list of pre-made apps you can install with 
 ```
 wget -qO- https://raw.githubusercontent.com/Botspot/pi-apps/master/install | bash
 ```
+The install script ensures YAD is installed, creates two menu buttons, an auto-started updater, and a file (`/usr/local/bin/pi-apps`) that allows pi-apps to be run from terminal. Besides those, nothing is modified outside of the Pi-Apps folder.
 
 <details>
-<summary>Piping to Bash</summary>
-Piping scripts straight to bash can be considered controversial as you don't get to see what the script is running, therefore a manual installation is provided below.
-</details>
-<br>
-<details>
-<summary>Manual installation</summary>
-
-# Installation
+<summary>To install manually if you prefer to see what happens under the hood</summary>
 ```
 git clone https://github.com/Botspot/pi-apps
 ~/pi-apps/install
 ```
 </details>
-
-The install script ensures YAD is installed, creates two menu buttons, an auto-started updater, and a file (`/usr/local/bin/pi-apps`) that allows pi-apps to be run from terminal. Besides those, nothing is modified outside your home directory.
 
 <details>
 <summary>To uninstall Pi-Apps</summary>
@@ -124,13 +116,3 @@ Embed code:
 > - You can look through the [list of app requests](https://github.com/Botspot/pi-apps/issues) and help create the scripts for them.
 > - If you see someone complaining about a broken app on Youtube, Reddit, or Facebook, please encourage them to join the [Pi-Apps Discord](https://discord.gg/RXSTvaUvuu) or to [open an issue](https://github.com/Botspot/pi-apps/issues/new) to get the problem solved. (I'm not on those sites very often.)
 > - You could help troubleshoot/diagnose issues others are having.
-
-## Privacy regarding analytics
-
-Pi-Apps takes analytic data which gives Botspot access to the following data:
-Which Apps are being installed
-Countries which apps are being downloaded from
-
-However this data does not label you, only generic data to see which apps are favorite :)
-
-If you do not wish to participate in the data collection you are welcome to withdraw by going to Pi-Apps Settings and changing the `Enable Analytics` setting to off, this means the Pi-apps downloads straight from source rather than going through bit.ly.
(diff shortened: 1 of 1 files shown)


CleanMachine1:
The privacy noticed was also added as I feel everyone should have easy clear information for privacy

It shouldn't be hidden in the settings menu

CleanMachine1:
Installation didn't have a cap I

Botspot:
One can't argue against the convenience of using a single-command installation script.

While it can be controversial, it was necessary to add a one-liner. Before that, people were git cloning to the wrong folder, or trying to download the zip from the github website. Many people made mistakes during installation, but since adding the one-line installation command, the number of successful installs has increased, while the number of people experiencing issues has decreased to nearly zero.

Those who don't trust piping to bash, are almost always going to be Linux-savvy. Those people are the minority of users, and we kept them happy by providing manual installation instructions.

Providing a warning will only serve to confuse the noob users, while at the same time it will appear contradictory: "Why does the official Pi-Apps readme warn users against using its own installation script?"

CleanMachine1:
Regardless, it should be mentioned, safety shouldn't be hidden, it may stop someone piping a malicious script in the future!

CleanMachine1:
> Those who don't trust piping to bash, are almost always going to be Linux-savvy. Those people are the minority of users, and we kept them happy by providing manual installation instructions.

CleanMachine1:
> Those who don't trust piping to bash, are almost always going to be Linux-savvy. Those people are the minority of users, and we kept them happy by providing manual installation instructions.

No.
Why shouldn't people learn.

Also most Linux users (I expect) are fluent with things but may not see the issue with aspects

Botspot:
> Providing a warning will only serve to confuse the noob users, while at the same time it will appear contradictory: "Why does the official Pi-Apps readme warn users against using its own installation script?"

Plenty of other repos reach a compromise by including both a manual installation script, and a simpler one-liner option.  
Those repos simply allow the user to choose: those who distrust can install manually, while those who want to install it quickly and effortlessly can use the one-liner.

So far so good.

But what you're asking to do, is to educate users _against_ using one of the options. To say it's unsafe or whatever. For Pi-Apps to cast a doubt on itself as if it may contain malware. Such a move would not benefit anybody.

Instead, what do you think about something like this?  
2021-04-04-145455_1920x1080_scrot
I'd also like @Itai-Nelken's input on this.

CleanMachine1:
And what about the other edit I have made about analytics

CleanMachine1:
> But what you're asking to do, is to educate users against using one of the options. To say it's unsafe or whatever. For Pi-Apps to cast a doubt on itself as if it may contain malware. Such a move would not benefit anybody.

No, there needs to be a need to teach people to read code before just running it, if everyone never looked at it because they used the easy 1-liner, things can slip through
But what you're asking to do, is to educate users against using one of the options. To say it's unsafe or whatever. For Pi-Apps to cast a doubt on itself as if it may contain malware. Such a move would not benefit anybody.

People shouldn't feel that pi-apps has malware because its acknowledged that piping is controversial, its for the greater good overall.

CleanMachine1:
About the edit made in that picture

It shouldn't be confusing, it should be something like "What this line does"

CleanMachine1:
Also I am not saying remove it, rather tell people its not good practise

Botspot:
> Also I am not saying remove it, rather tell people its not good practise

Other repos don't do that. Why should this one?

> (about the picture)
> instead of confusing people with long words

I think you misunderstood the picture. I'm thinking of changing "Manual installation" with "Feel free to inspect the codebase first by installing Pi-Apps manually: (click here)".  
What do you think of that?

CleanMachine1:
I feel thats worse than the original.
It should be simple just to make it look nice 

Maybe have another summary between the simple and manual just called "Piping to Bash", then have what I have done in my initial commit

In fact I will commit that change now
Give me 5

CleanMachine1:
@Botspot Done.

Have you seen the notice about the analytics part. To be honest I thought that would be the thing to raise controversy

Botspot:
> @Botspot Done.
> 
> Have you seen the notice about the analytics part. To be honest I thought that would be the thing to raise controversy

Yes I have. One thing at a time though please.

CleanMachine1:
Are you satisfied with https://github.com/Botspot/pi-apps/pull/533/commits/6c197f671c2791d3936d627ffa0b337587b279e2 ?

Botspot:
There, I changed it to the way I wanted.

I do agree with you that mentioning the analytics on the readme is a good idea. But, any day now, Pi-Apps will publicly display analytics data within the app list. And then at that point it would be apparent and obvious that Pi-Apps keeps tallys of everything so I'd see little need to mention it in the readme too. So right now it would be fine to mention analytics, but in a few days it wouldn't be necessary because it would be obvious.

I'll merge the changes.

CleanMachine1:
Small error made, if you click on the manual installation, it doesn't go into code tags

Botspot:
> Small error made, if you click on the manual installation, it doesn't go into code tags

There, fixed it. 28763fc9e62de7b9a421b5432e18b0eeb7c8f0a7

Botspot:
> I do agree with you that mentioning the analytics on the readme is a good idea. But, any day now, Pi-Apps will publicly display analytics data within the app list.

Done. Pi-Apps now displays the number of users per app with commit 6f8dad5dafd4027b04c5a078f9e993a1ec4cb244.

PR #555, opened 2021-04-08
OpenSCAD

koendv:
Adds latest OpenSCAD as AppImage. The 64-bit version is from http://openscad.org; the 32-bit version from https://github.com/koendv/openscad-raspberrypi


Itai-Nelken:
it would be better if you made a zip of the folder, it makes it easier to test.

koendv:
OK

Itai-Nelken:
thanks

PR #560, opened 2021-04-09
Replace yad install by yad appimage

pi-dev500:


Itai-Nelken:
interesting! I would add the .AppImage file extension to them so people know what they are.

pi-dev500:
Done now.

Itai-Nelken:
did you test this?

pi-dev500:
I have tested the armhf appimage but for 64 bits,I can't test.

Botspot:
> Replace yad install by yad appimage

I don't have a problem with doing this per se, if there's a good reason, but what _is_ your reason for preferring a yad appimage over an apt package?

pi-dev500:
It is faster and simplest to install. and that can be also more portable after some code modifications.

pi-dev500:
I have tested the arm64 appimage whith debian-buster-64 installed with Pi-Gro. That works also well expectly some gtk-warnings.

Botspot:
I've decided not to add this.

1. All Botspot apps use YAD. Nobody has had a problem with using apt to install yad.
2. AppImages sometimes have issues on RPiOS.

In my opinion, the potential benefit is too small, and the potential risks are too high, to consider merging this.

Issue #44, opened 2020-10-30
Updates reset all data.


MTCK123:
Any update (on pi apps) requires uninstallation of application causing all data for (eg: worlds on Minecraft java) to de deleted.

Botspot:
For some of the apps, I made sure that uninstalling apps does not delete their config. For other apps, I forgot to make sure this happened.
Sorry about that - we'll fix it.

MTCK123:
So is that including Minecraft java saves?

MTCK123:
Ok cool. I will close problem now! Thanks for fixing!

Botspot:
> Ok cool. I will close problem now! Thanks for fixing!

No problem. If you find any other apps causing the same issue, please let me know. Thank you as well.

CircuitTear:
> worlds on Minecraft java) to de deleted.

oof i would be pissed if that happened

PR #571, opened 2021-04-13
Update uninstall to remove all files of Minecraft

Nippyfus:


Itai-Nelken:
That folder isn't being removed because it would mean peoples minecraft worlds would be deleted every update.

Itai-Nelken:
See #44

PR #578, opened 2021-04-14
Installing flameshot to /tmp/.deb instead

CleanMachine1:
Just means it gets wiped on reboot if the rm fails for any reason


CleanMachine1:
@Botspot ?

PR #605, opened 2021-04-24
don't update wine on TwisterOS

Itai-Nelken:


Itai-Nelken:
also steam.
in my opinion leaving box86 is ok though.

also @Botspot is it ok I'm making so much branches on the main repo? should I make a fork for that?

Botspot:
> also @Botspot is it ok I'm making so much branches on the main repo? should I make a fork for that?

I'm not sure. Is it a bad practice?

Itai-Nelken:
> > also @Botspot is it ok I'm making so much branches on the main repo? should I make a fork for that?
> 
> I'm not sure. Is it a bad practice?

I don't know, just wanted to know if it bothers you.

PR #615, opened 2021-04-27
Add a fullscreen style menu

pi-dev500:
here is a screenshot:
screenpi-apps

I have also removed bugs for the details window when using xlunch app list style (up an down arrows) by adding the details function and add yad list variable at the begin of details.

The menu is a fork of [splitbrain/pimenu](https://github.com/splitbrain/pimenu)


pi-dev500:
The pi-apps button open the pi-apps homepage.

Botspot:
Wow this is very impressive. @pi-dev500, you're the first person I'm aware of who has made a significant amount of changes to Pi-Apps' core functionality.

A couple questions before I try it:
- Why pimenu? Upon first glance, it does not appear there's anything pimenu can do that xlunch can't. (xlunch can go fullscreen too)
- What modifications did you make to Pimenu? (Does it include [my improvements](https://github.com/Botspot/pimenu)?)

pi-dev500:
That don't have your improvements because I have cloned splitbrain's menu but I have added the message and the pi-apps homepage button. I have also removed some bugs by adding the details function.

pi-dev500:
I've chosed pimenu because that's faster.

CleanMachine1:
How does this integrate to using a non max, I would imagine it would be pretty ugly, also I personally think the entire thing is ugly, colours and the icons don't go well and the icons are tiny, however with some changes it could/will be nice!

Overall good job though

pi-dev500:
I am adding colors and icons size settings.

techcoder20:
Just a suggestion, what about adding it as a plugin app instead of integrating it into pi apps.

pi-dev500:
Try it now . I have added settings for pimenu: icons size, buttons colors ,buttons style...

pi-dev500:
> Just a suggestion, what about adding it as a plugin app instead of integrating it into pi apps.

It's not possible because that modify curents scripts.

pi-dev500:
I'm going to add a search bar.

vedforealisqt:
@pi-dev500 i would like to know your discord :)

theofficialgman:
I saw this, and wasn't aware pi-apps had an xlunch backend for the GUI... but I couldn't run it, pi-apps would just not open after displaying the splashscreen
turns out this was the problem: https://github.com/Botspot/pi-apps/commit/5340e5b6ad815eb1e6440a0b776190005997ef4c

theofficialgman:
I will say, pi-apps xlaunch backend looks very nice, but its unusable on touchscreens. The scrollbar on the right is non-functional with touch or mouse selection input, it only responds to scroll wheel or arrow keys (arrow keys just move the selection which moves ends up scrolling)

theofficialgman:
there is going to be significant incompatibility between this and the new-gui https://github.com/Botspot/pi-apps/pull/1580

closing this for now, you are welcome to test that new GUI and make suggestions for changes after it has merged

pi-dev500:
I'm back working on it. I removed the old dark-edges buttons, and in a close future, I should allow it to have themes. Anything is integrated, so it could have his own desktop entry, like an independent app. For install, his folder is only placed in pi-apps one. I have a search working like xlunch's one, but no filter by category. I only need to add a touchscreen-compatible scrollbar(I'm thinking of buttons disposed verticaly corespondig to pages.) and a touchscreen keyboard if I don't find a lightweight one, but it will be a separate project, compatible with all applications...

pi-dev500:
I will also add settings GUI and integrated terminal (see pi-dev500/TkTerm)

PR #637, opened 2021-05-05
Fix typo in update-exclusion

Jai-JAP:


CleanMachine1:
Documentation fixing - MY FAVOURITE :smile:

Botspot:
Thanks for catching that.

This pull request will be marged, but beware that update-exclusion is stored in the `data/` directory. Consequently, it won't be updated on all pre-existing systems.

Itai-Nelken:
> it won't be updated on all pre-existing systems.

it should, because it fixes the wine installing on twisterOS when updating..

Botspot:
> > it won't be updated on all pre-existing systems.
> 
> it should, because it fixes the wine installing on twisterOS when updating..

This file is not modified directly **by the updater**. Therefore, changing this file in the repo now, won't change it for the people who already had this file. Make sense?
Nothing in the data directory is checked for updates. This is necessary in order to allow users to customize these files.

PR #642, opened 2021-05-06
Fix screensaver launch

mobilegmYT:
Apparently, the directory variable pointed to pi-apps, so when I launched Windows Screensavers it launched pi-apps 🤦‍♂️.


mobilegmYT:
Looks like it's been closed with https://github.com/Botspot/pi-apps/commit/5d808bef6cbd037b33e2bff6561ebec478815d33

PR #643, opened 2021-05-06
Update clamav version to 6.11

mobilegmYT:


Botspot:
@mobilegmYT Can you confirm that you really did create a new Discord account named `RPiNews#0816`?
Also, how come you aren't "constantly" online with this new account, like you were with your previous one?

One more thing: if you were token grabbed, you do realize that changing your password would have reset the token, right?

PR #651, opened 2021-05-10
Fix wine app path in `update-exclusion` (fixes #649)

### LINKED ISSUE #649: update-exclusion failing
I'm on Twister OS.

update-exclusion has this:
```
apps/Wine (x86)/
apps/Steam
```

I recently noticed that Wine got an update even when it's not supposed to. I tested this with manually triggering a reinstall with the update system, and it still reinstalled wine even though it's not supposed to.

Itai-Nelken:
## Not tested
fixes #649

### DIFF
--- a/gui
+++ b/gui
@@ -115,6 +115,13 @@ runonce "
     echo -e 'apps/Steam' >> '${DIRECTORY}/data/update-exclusion'
   fi
 "
+#fix wine in update-exclusion
+runonce "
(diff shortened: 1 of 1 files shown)


Itai-Nelken:
https://github.com/Botspot/pi-apps/compare/765275e38170...cdfe0ebd8f1d fixes this, so closing.

PR #665, opened 2021-05-16
Fix remarkable 64bit install script

Itai-Nelken:
# Not ready for merging yet
Installs successfully, but remarkable fails to start. tested on a fresh install of Ubuntu 21.04 desktop.

- added 2 missing dependencies.
- install all the packages in a single `pkg-install` line to prevent uninstalling the dummy deb for each dependency so at the end only the last package is installed.
- The fix in 3390bcec32e0b12b5c2c18a7c47b127ed688611b didn't work, so I changed it to `$(pwd)/xxx.deb` instead of `~/xxx.deb`.


Botspot:
As long as the script is `cd`'ed into `$HOME`, then I think all instances of `$(pwd)` could be replaced with `$HOME`.

theofficialgman:
remarkable uses a deb install now as of https://github.com/Botspot/pi-apps/commit/a46310c7f139b4ac44fa13c2d13818d21924821d

PR #678, opened 2021-05-21
Update Etcher to v1.5.120

chunky-milk:
also note that the deb name is now `balena-etcher` and not `balena-etcher-electron`.

### DIFF
--- a/apps/BalenaEtcher/install-32
+++ b/apps/BalenaEtcher/install-32
@@ -7,7 +7,7 @@ function error {
   exit 1
 }
 
-rm -f ./etcher*armv7l.deb
+rm -f ./etcher.deb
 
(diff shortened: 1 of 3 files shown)


Itai-Nelken:
Could you make the uninstall script something like this please (checks if the old etcher package is installed, if yes uninstall it else uninstall the new one). **This is not tested**
```bash
if dpkg -l "balena-etcher-electron" &> /dev/null ;then
	sudo apt purge -y balena-etcher-electron
else
	sudo apt purge -y balena-etcher
fi
```

Itai-Nelken:
# not ready for merging yet! the new etcher builds seem to be broken!

chunky-milk:
Should I close this for now?

Itai-Nelken:
If you want

PR #680, opened 2021-05-22
Rename Turbo Scratch to Turbowarp

wgyt:
Turbowarp is the actual name of this app, but Pi Apps refers to it as Turbo Scratch.


Botspot:
> Turbowarp is the actual name of this app, but Pi Apps refers to it as Turbo Scratch.

When this app was first added to the app store, it was decided that many users would not recognize it as an alternative, Scratch-like tool. Few people would realize that it was an improved version of Scratch.

So, [in the name of familiarity](https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=308834), I decided to name it Turbo Scratch in the hope that more people would find it.

@wgyt, after understanding my reasoning behind the name change, do you still feel like renaming this to Turbowarp would be a good idea, or not anymore?

wgyt:
> > Turbowarp is the actual name of this app, but Pi Apps refers to it as Turbo Scratch.
> 
> When this app was first added to the app store, it was decided that many users would not recognize it as an alternative, Scratch-like tool. Few people would realize that it was an improved version of Scratch.
> 
> So, [in the name of familiarity](https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=308834), I decided to name it Turbo Scratch in the hope that more people would find it.
> 
> @wgyt, after understanding my reasoning behind the name change, do you still feel like renaming this to Turbowarp would be a good idea, or not anymore?

I think it could be a good idea if we renamed it do something like "Turbowarp (Scratch Mod)" or something like that.

mobilegmYT:
> > > Turbowarp is the actual name of this app, but Pi Apps refers to it as Turbo Scratch.
> > 
> > 
> > When this app was first added to the app store, it was decided that many users would not recognize it as an alternative, Scratch-like tool. Few people would realize that it was an improved version of Scratch.
> > So, [in the name of familiarity](https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=308834), I decided to name it Turbo Scratch in the hope that more people would find it.
> > @wgyt, after understanding my reasoning behind the name change, do you still feel like renaming this to Turbowarp would be a good idea, or not anymore?
> 
> I think it could be a good idea if we renamed it do something like "Turbowarp (Scratch Mod)" or something like that.

I mean if you hover over the app or check the description it'll tell you that. Generally, it's best to avoid parentheses in app names since it messes up installing from CLI.

wgyt:
> I mean if you hover over the app or check the description it'll tell you that. Generally, it's best to avoid parentheses in app names since it messes up installing from CLI.

Fair point, I didn't think that out at all :P

wgyt:
@Botspot Because pull force pushed the branch none of the changes were merged :|

PR #695, opened 2021-05-27
Add root warning

mobilegmYT:


Itai-Nelken (approved):
I approve this change, but I want @Botspot to approve as well.

Botspot:
This warning may confuse beginner users. Also it won't catch 100% of people, as not everybody reads the directions. Many people never even see the pi-apps readme, but copy-n-paste the command from a YouTube video!

I propose we use a programmatic approach that stops the user from running the pi-apps `install` script if the root user is detected.

Issue #694, opened 2021-05-27
Command not found


Kirill2006183:
/root/ pi-apps/ api: line 15: file: command not found
/root/ pi-apps/ api: line 17: file: command not found 
how to fix it

mobilegmYT:
> /root/ pi-apps/ api: line 15: file: command not found
> /root/ pi-apps/ api: line 17: file: command not found
> how to fix it

Please do not use root to install/run pi-apps.

Botspot:
> /root/ pi-apps/ api: line 15: file: command not found
> /root/ pi-apps/ api: line 17: file: command not found
> how to fix it

I see two problems.

1. First, there has been absolutely zero testing for running Pi-Apps as the root user. Proceed at your own risk and brace yourself for endless permission errors.
2. Whatever system you are using does not include the `file` command. I've not seen this problem before, though I have known it's theoretically possible as the `file` command is not POSIX-compliant.

Kirill2006183:
> > /root/ pi-apps/ api: line 15: file: command not found
> > /root/ pi-apps/ api: line 17: file: command not found
> > how to fix it
> 
> Please do not use root to install/run pi-apps.

Now access is denied /root/pi-apps/gui I did without root

Botspot:
> > > /root/ pi-apps/ api: line 15: file: command not found
> > > /root/ pi-apps/ api: line 17: file: command not found
> > > how to fix it
> > 
> > 
> > Please do not use root to install/run pi-apps.
> 
> Now access is denied /root/pi-apps/gui I did without root

Exactly. If you use root even once with pi-apps, it changes the permissions for some files and breaks its usage for non-root usage.
To solve, I suggest you delete the whole pi-apps folder and start over.

Also, @Kirill2006183, what operating system are you using that doesn't include the `file` command?

Kirill2006183:
> > > > / root / pi-apps / api: строка 15: файл: команда не найдена 
> > > > / root / pi-apps / api: строка 17: файл: команда не найдена, 
> > > > как это исправить
> > > 
> > > 
> > > Пожалуйста, не используйте root для установки / запуска pi-приложений.
> > 
> > 
> > Теперь доступ запрещен / root / pi-apps / gui Я сделал без рута
> 
> Точно. Если вы используете root даже один раз с pi-приложениями, это изменяет разрешения для некоторых файлов и нарушает его использование для использования без полномочий root.
[...]

PR #696, opened 2021-05-27
print error and exit if install script is run as root

Itai-Nelken:
#695 
and should fix cases like #694

### DIFF
--- a/install
+++ b/install
@@ -7,7 +7,7 @@ function error {
 
 cd $HOME
 
-if [[ "$EUID" == 0 ]]; then
+if [[ "$(id -u)" == 0 ]]; then
 	error "You cant install pi-apps as root! please try again as a regular user."
 fi
 
--- a/install
+++ b/install
@@ -8,7 +8,7 @@ function error {
 cd $HOME
 
 if [[ "$(id -u)" == 0 ]]; then
-	error "You cant install pi-apps as root! please try again as a regular user."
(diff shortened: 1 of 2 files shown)


Kirill2006183:
> [# 695](https://github.com/Botspot/pi-apps/pull/695)
> и должен исправить такие случаи, как [# 694](https://github.com/Botspot/pi-apps/issues/694)

The question is solved I had to download the program file sudo apt install file only without root and started me

Kirill2006183:
> [# 695](https://github.com/Botspot/pi-apps/pull/695)
> и должен исправить такие случаи, как [# 694](https://github.com/Botspot/pi-apps/issues/694)

It starts with and without root without errors

Kirill2006183:
> [# 695](https://github.com/Botspot/pi-apps/pull/695)
> и должен исправить такие случаи, как [# 694](https://github.com/Botspot/pi-apps/issues/694)

It was necessary to install the sudo apt install file

Botspot:
Perhaps [this technique](https://stackoverflow.com/a/28776100) would be better? We need the script to detect `sudo` usage, *as well as* `root` usage.

Itai-Nelken:
It now uses `-id -u`, I tested and it does work: the error is printed when running with `sudo` and when running as the root user without `sudo`.

Itai-Nelken:
ready to merge in my opinion. what do you say @Botspot?

PR #703, opened 2021-05-31
Enable Github pages

slashtechno:
I thought that it would be cool if Pi-Apps had a website. Github Pages may need to be enabled though.

Also, [this is the pages site for my fork of Pi-Apps](https://slashtechno.github.io/pi-apps/)


mobilegmYT:
IMO I don't see much purpose in enabling GH pages since the README covers everything, and using GH pages just to show the readme is kinda useless.

slashtechno:
> IMO I don't see much purpose in enabling GH pages since the README covers everything, and using GH pages just to show the readme is kinda useless.

It should also appear on Google as a full website

chunky-milk:
> I thought that it would be cool if Pi-Apps had a website. Github Pages may need to be enabled though.

I agree, and I think it's long overdue..

Lots of open-source projects have "home page" websites for documentation, which can make them easier to read due to a better style IMO.

Itai-Nelken:
If we add a website, it should be a proper one. not just a glorified readme.
anyone here has good html and css skills?

slashtechno:
For the time being, I think Github Pages should work however.

Botspot:
There's no point in doing a half-good job.
Also I'm not sure how many of our users *want* a website. Nearly all our users find out about Pi-Apps because of word-of-mouth. Currently everyone simply points to https://github.com/Botspot/pi-apps, and this seems to work fine.

Github sites are a familiar sight to many people, and will especially be useful for those who wish to view the wiki, inspect files, see the issues or PRs, or check out our list of contributors.
With a website, the user-experience would be foreign.

I'm not against a Pi-Apps website, but I don't see the point of starting one unless we do it right. And even then, many people will prefer the familiarity of a github.com site, due to past experiences with the platform.

ryanfortner:
I've created a website: https://ryanfortner.github.io/pi-apps-website/

its source code is available [here](https://github.com/ryanfortner/pi-apps-website). Could someone have a look and let me know their thoughts on it?

ryanfortner:
> @ryanfortner My thoughts:
> 
> * I don't think the Main header text should be bold (as well as the body sub-text, and the `PI-APPS FOR RASPBERRY PI` part in the footer), and it would be cool to have a logo right there. This is just my personal preference though.
> * The install button in the footer leads to a 404
> * It doesn't really make sense for the `Donate` button to just lead to the Pi-Apps GitHub....
> * We could just tell the viewers how to install in the body, instead of pointing them to GitHub, although we could do that too.
> 
> I'll create a fork with some of my adjustments/fixes.
> 
> Despite the few problems, it's a way better website then I could have made! 🤣

Thanks for the feedback! 

Some of website code is heavily modified from [remarkable](https://github.com/remarkableapp.github.io)'s website and a few others, but it's been nice getting some experience in html/css/js.

Crilum:
My fork: https://crilum.github.io/pi-apps-website/

Warning: it's still under construction.. So, if something looks weird, I'm probably going to fix it...

ryanfortner:
@Crilum , I've updated the install button on the footer, so there's no more 404 error. There isn't a donation link, so I've just made the button redirect to the github page until there is a donation option.

ryanfortner:
@Botspot what do you think?

Crilum:
@ryanfortner After looking at the code, I realise that the all-caps stuff is part of `pure`. We could edit that to use a different font.

If there isn't really a way to donate, I think we should remove the `Donate` button entirely, at least until @Botspot sets up a way to donate.

ryanfortner:
> @ryanfortner After looking at the code, I realise that the all-caps stuff is part of `pure`. We could edit that to use a different font.
> 
> 
> 
> If there isn't really a way to donate, I think we should remove the `Donate` button entirely, at least until @Botspot sets up a way to donate.
> 
> 

Makes sense. Feel free to make a PR, I'm away from my computer for a bit.

Crilum:
@Botspot what font do you use for the Pi-Apps logo?

slashtechno:
> My fork: https://crilum.github.io/pi-apps-website/
> 
> Warning: it's still under construction.. So, if something looks weird, I'm probably going to fix it...

@ryanfortner Looks amazing!

I think there is also the option to make it with Wordpress and self host it or make it with [Publii](https://getpublii.com/).

Crilum:
> @Botspot what font do you use for the Pi-Apps logo?

Is it Cantarell? It looks really similar.. I had an idea that we could make the other links on the header the same font.

Crilum:
@ryanfortner Is there a reason why there are so many stylesheets, some with duplicate rules? Just wondering, because I'm trying to comment out some rules, and I have to check all the stylesheets to see if there are duplicates, or the page stays the same.. :laughing:

Crilum:
I've changed the website up a little bit, what does everybody think?

https://crilum.github.io/pi-apps-website/

I used `Cantarell` as the font for the header links, because it looked like it was the same font used in the Pi-Apps logo..

Crilum:
@slashtechno @Botspot I think this should be re-opened.

slashtechno:
> I've changed the website up a little bit, what does everybody think?
> 
> https://crilum.github.io/pi-apps-website/
> 
> I used `Cantarell` as the font for the header links, because it looked like it was the same font used in the Pi-Apps logo..

@Crilum The website looks amazing. In fact, is it okay if I use model some of my  websites after this using the source code of this site?

Crilum:
>@Crilum The website looks amazing. In fact, is it okay if I use model some of my websites after this using the source code of this site?

As far as I'm concerned, yeah! It's mostly @ryanfortner's code, so you should ask him too.

>On mobile the logo goes under the footer though.

Yeah, I noticed that in Chrome's device screen emulator thingy (Chrome Dev Tools are built in to Min), not sure how to fix it..

Botspot:
> @Botspot what font do you use for the Pi-Apps logo?

Google's Product Sans font. I acquired it legally and it's embedded into the vector svg, but for a website I suggest a font that's in the public domain so we avoid any copyright issues.

Crilum:
Cantarell is licenced under the Open Font Licence, so it's free for our type of use. It's also available via Google Fonts.

ryanfortner:
the website has been migrated to https://github.com/Pi-Apps-Coders/website, access it at https://pi-apps-coders.github.io/website

Crilum:
@Botspot Anything you thing should be added?

PR #707, opened 2021-06-02
Add 64bit support for sublime text (using official arm64 binaries)

Itai-Nelken:
Sublime text now have official Linux (and MacOS) arm64 binaries! unfortunately armhf OS's will have to stay with sublime text 2 running with box86 while 64bit OS's get sublime text 4 running natively.
**the script is not tested, testers needed.**

### DIFF
--- a/apps/Sublime Text/install-64
+++ b/apps/Sublime Text/install-64
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+DIRECTORY="$(dirname "$(dirname "$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )")")"
+
+function error {
+  echo -e "\\e[91m$1\\e[39m"
+  exit 1
+}
+
+#remove old archive if it exists
+rm -f sublime_text_arm64.tar.xz #should fail thats why the '-f' flag is being used (no error will be shown if it fails).
+#download new archive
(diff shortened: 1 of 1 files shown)


finitelife (changes requested):
Why install a static version when the Deb is available? 

```bash
wget -qO - https://download.sublimetext.com/sublimehq-pub.gpg | sudo apt-key add -

sudo apt install apt-transport-https

echo "deb https://download.sublimetext.com/ apt/stable/" | sudo tee /etc/apt/sources.list.d/sublime-text.list

sudo apt update

sudo apt install sublime-text
```

Itai-Nelken:
I have had problems with their repo in the past, but I'll take a look if it works fine now.

Itai-Nelken:
@Botspot what do you think is better:
1) not using the `apt` repo - we need to update manually but installation is more foolproof.
2) using the `apt` repo - automatic updates but there are chances of apt getting messed up.

chunky-milk:
I just tested the apt repo, works fine

Botspot:
> @Botspot what do you think is better:
> 
> 1. not using the `apt` repo - we need to update manually but installation is more foolproof.
> 2. using the `apt` repo - automatic updates but there are chances of apt getting messed up.

In this case, I see no problem with using the apt repo, because we've done that before.
Just make sure that it's done cleanly. (Example: use a sources.list.d file, and if apt signing fails, delete the .list file before the script exits with failure)

Botspot:
https://github.com/Botspot/pi-apps/commit/876722f1697fd5f05d7b6b36e5d0c51243738436

PR #708, opened 2021-06-02
Fix and improve node.js install script

Itai-Nelken:


Botspot:
It appears your pull request is only asking to delete two files. It does not replace them with anything.

Itai-Nelken:
oops thats not what was supposed to happen. I replaced them locally with a single `install` script, but it didn't ge pushed. I'll push  it later.

Itai-Nelken:
Added

Botspot:
> Added

Thank you.

Is there a way the function `install_node` could be made less cryptic? Those `if` statements and various `echo` usages seem to be way more complicated, and confusing, than necessary.

Itai-Nelken:
>Is there a way the function install_node could be made less cryptic? Those if statements and various echo usages seem to be way more complicated, and confusing, than necessary.

You are right, in the case of this script the message part is only used in one place and there is no need to make the `install_node` function so complicated.
I removed the `message` `if`'s, instead I added `echo "$3"` to the firs `if` (the one that sets `NVM_DIR` if `$1` == `set_nvm_dir`).
also `nvm` changed `NVM_DIR` to `~/.nvm` again, so I also fixed that.
I tested the install script and it works fine.

PR #732, opened 2021-06-14
Update Drawing

Itai-Nelken:


Botspot:
Successful test on my armhf system.

Itai-Nelken:
I guess it should work on arm64 as well. I don't think that the dependencies changed.

PR #734, opened 2021-06-15
Fix steam not launching

mobilegmYT:
Not tested, should work in theory. Steam update broke stuff, this changes box86 magic to use unbroken stuff.


Itai-Nelken:
If someone can confirm this works, I'll merge.

mobilegmYT:
> If someone can confirm this works, I'll merge.

This stuff came from ptitseb in pi labs discord. The syntax should be right but I'll get someone to check it.

theofficialgman:
`BOX86_NOGTK=1` is handled already by box86 as of one day after this PR creation: https://github.com/ptitSeb/box86/commit/72a896a95aee45549073534a338ecb537f0896e8

PR #769, opened 2021-06-28
Switch MC Java to lunar launcher

mobilegmYT:
Initial PR. Will need @gl91306 to have a look and make sure everything is good and all downloaded files are cleaned up properly. This is for the future when Lunar has full MS acc support fixed.


mobilegmYT:
Also I know that @gl91306 didn't want pi-apps to have lunar added, but perhaps a compromise could be worked out?

Botspot:
> Also I know that @gl91306 didn't want pi-apps to have lunar added, but perhaps a compromise could be worked out?

Probably so. And if not, how hard would it be to use the stock version of Lunar Client?

mobilegmYT:
> > Also I know that @gl91306 didn't want pi-apps to have lunar added, but perhaps a compromise could be worked out?
> 
> Probably so. And if not, how hard would it be to use the stock version of Lunar Client?

It's all written in electron and javascript so we'd need box64, as well as compiling some java libraries maybe.

apps/Minecraft Java/install-64
```diff
@@ -7,51 +7,97 @@ function error {
...
 
-#Create desktop shortcut
-cd ~/.local/share/applications/
+#Run modprobe fuse
+sudo modprobe fuse
+
+cd ~
+
+#Download client and assorted java stuff
+wget https://github.com/gl91306/lunar/raw/master/lunarclient-2.7.3c-armv7l.AppImage
```
Jai-JAP, line 49: This should be ` wget https://github.com/gl91306/lunar/raw/master/lunarclient-2.7.3c-aarch64.AppImage `

mobilegmYT:
Closing since it'll be added in a different way

PR #773, opened 2021-06-30
Update balenaEtcher to v1.5.120

chunky-milk:


Itai-Nelken:
This doesn't have the blanking issue when trying to flash something, right? Can someone test it please?

chunky-milk:
> This doesn't have the blanking issue when trying to flash something, right? Can someone test it please? 

The only time I had issues is with flashing .xz files, but I've had that issue in previous releases also.

PR #775, opened 2021-07-01
Update Cura to 20210629 and Webcord to 1.5.0

chunky-milk:


Botspot:
Doesn't Discord auto-update itself now? If so, it may not be necessary to use our discord version number anymore.

Itai-Nelken:
It only tells you when there is a new version as far as I know.

PR #804, opened 2021-07-14
Added new PiGro Icon

actionschnitzel:
Hi, I redesigned the icon would be great if you changed it. Of course, I kept the original dimensions.


actionschnitzel:
Hi, I redesigned the icon would be great if you changed it. Of course, I kept the original dimensions.

PR #823, opened 2021-07-21
Create CODE_OF_CONDUCT.md for betterment

CoderLogy:


Botspot:
Please convince me why **we** need this, while so many other repositories don't have one of these files.

CoderLogy:
we need this as its important by GitHub to make this repo better and rankable plus it would save you from any type of law issue and also it would protect you from consumer rights and last it would protect users of pi apps from any trolling, harassment, etc just read the file, you would get the point. Big repo like Ubuntu or Linux has it so why not prepare from now plus it does not have any disadvantages but only advantage.

Botspot:
Everyone here has historically been polite and used common sense.
But suppose that a troll comes along with the intent of causing trouble. Would they read this file? Probably not. Even if they did read it, would that dissuade them? I doubt it.

My reasons for not adding this file:

- It would be yet another non-functional file that takes up space on the main pi-apps page.
- It would probably not do any good: those who already practice **social common sense - don't need it**. And those who are looking for trouble won't read or abide by it.
- I don't want to further empower problematic github users with a code of conduct they could use to demand I treat them "fairly".
- Github already has its own [site-wide community guidelines](https://docs.github.com/en/github/site-policy/github-community-guidelines).
- Github codes of conduct do heve their downsides, loopholes, and implications. See [this blog post](https://dancerscode.com/posts/why-the-open-code-of-conduct-isnt-for-me/) for one author's opinion on the subject.

Bottom line: We haven't had problematic interaction here. (yet) And if we do in the future, I doubt a code of conduct would benefit the situation at all. If a troll shows up, I can simply [ban](https://docs.github.com/en/communities/maintaining-your-safety-on-github/blocking-a-user-from-your-personal-account) them from all Botspot repositories.

PR #824, opened 2021-07-21
Create SECURITY.md

CoderLogy:


CleanMachine1:
There are quite a few errors here and things that don't fit here at all.

Maybe something similar regarding 

Pi-apps has to install from external sources and also runs precompiled binaries. If you do not feel comfortable installing precomplied binaries then most of the software can be built from source very easily. If any software included is found to have vulnerabilities or similar, please report an issue to the pi-app's repository so that the app can be updated or removed.

CoderLogy:
Fixed now @CleanMachine1  tell me anything more to add it and @Botspot  pls reply and add my pull request mainstream

Botspot:
I don't think we need this file.

1. If someone ever finds malware in an app on Pi-Apps, they will spread that information far and wide. Everyone will know. We don't need to **_tell_** people to let us know - because they already are bound to tell everyone.
2. Curious people are _welcome and encouraged_ to inspect apps before installing them. This can be done manually with a file manager, or by clicking the "Scripts" button in an app's Details window.

CoderLogy:
@Botspot  I made it so like people would read it and also if not then they would know their rights which is important to let them know and also it would be supportive for opensource community and last not least it does not have any disadvantage, just it help people to know what they are supposed to do for security. Plus it increases more views and more popularity for pi-apps with GitHub if we had this

PR #839, opened 2021-07-24
Update install-64 to 4.0.2

conrackx:
Link is tested and working, version is vivaldi-stable and not snapshot


Itai-Nelken:
Thanks!

PR #845, opened 2021-07-25
Cura 64-bit: stl-thumb arm64 instead of armhf

chunky-milk:


Botspot:
Thank you for catching this.

PR #857, opened 2021-07-31
turbowarp 0.8.1

mobilegmYT:

PR #858, opened 2021-07-31
turbowarp fix

mobilegmYT:


mobilegmYT:
#857

Itai-Nelken:
Oops, missed that one...

Botspot:
Small typos like this are routine and common enough to be planned for. That's why the default update interval is "Daily", instead of "always" to give us a bit of a safety net.

PR #872, opened 2021-08-05
Add Puffin 64-bit install script

coder0107git:
Tested on x86_64 system.


Itai-Nelken:
You are downloading the x86_64 version, pi-apps is for arm 32/64 bit OS's on the Raspberry Pi.

Itai-Nelken:
if you want to create the install-64 for puffin, find a **arm64** (**aarch64**) package.

PR #890, opened 2021-08-11
Set upstream server in lokinet.ini

necro-nemesis:
Changes in lokinet dns settings require upstream server value to be provided.


necro-nemesis:
Thanks.

PR #909, opened 2021-08-15
Fix apt-key line

scamdotnet:


Botspot:
:( Looks like the server URL was misspelled. I [caught it](https://github.com/Botspot/pi-apps/commit/36f30226688562a9822c3cbbb64b826200931820) a few minutes ago.

PR #925, opened 2021-08-19
Force mcpi reinstall and add some notes

mobilegmYT:
DO NOT MERGE NOW I ACCIDENTLY CLICKED WRONG BUTTON MAKING PR. Merge when gmcpil 1.13.1 is released.

### DIFF
--- a/apps/Minecraft Pi (Modded)/install
+++ b/apps/Minecraft Pi (Modded)/install
@@ -31,6 +31,7 @@ done
 
 # Remove old MCPILs
 sudo apt-get remove -y mcpil-r mcpil &>/dev/null || true
+echo "Don't worry if above 2 lines look like errors, if they show up for you it means all is well!"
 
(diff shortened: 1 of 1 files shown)


Itai-Nelken:
mark as 'ready for review' when gmcpil 1.13.1 is released.

mobilegmYT:
@Itai-Nelken ready for merge

Botspot:
Why are we letting the user know that the above "errors" are normal, when these errors have been completely silenced with `&>/dev/null`?

mobilegmYT:
> Why are we letting the user know that the above "errors" are normal, when these errors have been completely silenced with `&>/dev/null`?

I think the /dev/null stuff is broken, if you look at pi-apps error logs you'll see that the errors still show up for packages not found.

Botspot:
Is this PR still relevant?

mobilegmYT:
> Is this PR still relevant?

Er no, I was hoping to switch it to use the custom buster debs and add sound support, but since it didn't work for you (segfault) I'm keeping it at the current version without sound.

PR #936, opened 2021-08-28
Use `pkg-install` in Mac OS Theme

techcoder20:
I added all the dependencies that are installed using apt but some dependencies are there where i manually download the deb and install it. I have seen deb files used with `pkg-install` but as it is running before the whole install scripts those debs wont be there. If you want I can only download and install those debs using the pi apps install script so that I can make use of `pkg-install`.

### DIFF
--- a/apps/Mac OS Theme/install-32
+++ b/apps/Mac OS Theme/install-32
@@ -7,6 +7,10 @@ function error {
   exit 1
 }
 
+# Get dependencies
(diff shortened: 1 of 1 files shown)


Botspot:
> I added all the dependencies that are installed using apt but some dependencies are there where i manually download the deb and install it. I have seen deb files used with `pkg-install` but as it is running before the whole install scripts those debs wont be there. If you want I can only download and install those debs using the pi apps install script so that I can make use of `pkg-install`.

Yes please. That way, all the `apt update` and `apt install` lines are in English, are diagnosed and in some cases, automatically repaired.
And yes, pkg-install is fully capable of installing local packages. Just be sure to use pkg-install only once in the install script. (running it a second time will undo changes made by the first time it was run)

PR #953, opened 2021-09-16
Add ability to force update pi-apps from terminal

oxmc:


Botspot:
Closing, because I'm actively re-writing the entire `updater` script and making it integrate with the pi-apps gui.

PR #963, opened 2021-09-22
Fix some WhatsApp issue

cycool29:
- Built with latest nativefier to solve `Old build detected` warning.

- Fix `To use WhatsApp, update Chrome or use Mozilla Firefox, Safari, Microsoft Edge or Opera.` by deleting service worker every time when WhatsApp launch. I know there are several JS injection solution on the web but they are often not working. 

**NOTE**: The service worker fix tested works perfectly in **normal version WhatsApp** but in **Multi-device beta** it will show a prompt like this:

update_available

Still exploring for fix of this bug...

### DIFF
--- a/apps/Whatsapp/install-32
+++ b/apps/Whatsapp/install-32
@@ -14,7 +14,7 @@ tar -xf WhatsApp-linux-armv7l.tar.xz || error "Failed to extract!"
 rm WhatsApp-linux-armv7l.tar.xz
 mv "WhatsApp-linux-armv7l" WhatsApp
 
-#Modify run-whatsapp
+#Modify run-whatsapp to remove Service Worker before launching WhatsApp
 echo "rm -rf $HOME/.config/whats*/Service* && $HOME/WhatsApp/WhatsApp" > WhatsApp/run-whatsapp
 
 #create menu shortcut
@@ -22,5 +22,5 @@ echo "[Desktop Entry]
 Name=WhatsApp
 Type=Application
 Exec=bash /home/pi/WhatsApp/run-whatsapp
-Icon=/home/pi/WhatsApp/resources/app/icon.png
+Icon=$(dirname "$0")/icon-64.png
 Categories=Network;Internet" >  ~/.local/share/applications/whatsapp.desktop || error "Failed to create menu button!"
--- a/apps/Whatsapp/install-32
+++ b/apps/Whatsapp/install-32
@@ -12,10 +12,12 @@ rm -f WhatsApp-linux-armv7l.tar.xz
 wget https://github.com/cycool29/whatsapp-for-linux/releases/download/rpi-v1.0/WhatsApp-linux-armv7l.tar.xz || error 'Failed to download!'
 tar -xf WhatsApp-linux-armv7l.tar.xz || error "Failed to extract!"
 rm WhatsApp-linux-armv7l.tar.xz
-mv "WhatsAppWeb-linux-armv7l" WhatsApp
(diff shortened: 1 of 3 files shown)


Botspot:
Thank you for seeing a problem and taking the time to solve it. We need more people like you.

Your scripts had a few problems (like not marking run-whatsapp as executable, not formatting the desktop file correctly, and not removing the right filenames during an uninstall)
I've modified the scripts so that they do work correctly, but am unable to fully test them because I don't have a WhatsApp account. Could you try it?
[WhatsApp.zip](https://github.com/Botspot/pi-apps/files/7211313/WhatsApp.zip)
Simply extract that zip file to your pi-apps/apps folder.

Botspot:
Also tagging @Itai-Nelken - he might be interested in testing the zip before it goes live.

cycool29:
> Thank you for seeing a problem and taking the time to solve it. We need more people like you.
> 
> Your scripts had a few problems (like not marking run-whatsapp as executable, not formatting the desktop file correctly, and not removing the right filenames during an uninstall)
> I've modified the scripts so that they do work correctly, but am unable to fully test them because I don't have a WhatsApp account. Could you try it?
> [WhatsApp.zip](https://github.com/Botspot/pi-apps/files/7211313/WhatsApp.zip)
> Simply extract that zip file to your pi-apps/apps folder.

Thanks for improving my script. I tested it locally and ran with pi-apps, it installed and launched perfectly.
The only problem is the uninstalling script as I changed the app name to `WhatsApp` instead of `WhatsAppWeb`. I made some commits adding the updated script and update uninstalling script. 😁

Botspot:
Tested again by me and it works fine. In the absence of @Itai-Nelken, I'm merging the PR without his test-results, but that's not a big deal.

Itai-Nelken:
I haven't tested it, but it looks good to me (not that it matters anymore).

Itai-Nelken:
@cycool29 the following JS injection code works fine for deleting the service worker stuff, though it sometimes requires to restart the webapp a few times after "updates". it's what the previous version of whatsapp in pi-apps used.
```js
if ('serviceWorker' in navigator) {
    caches.keys().then(function (cacheNames) {
        cacheNames.forEach(function (cacheName) {
            caches.delete(cacheName);
        });
    });
} 
```
(I have no idea how it works, I got it from here: https://github.com/nativefier/nativefier/issues/1112#issuecomment-795424961)
also using a firefox or safari user agent helps sometimes.

cycool29:
> @cycool29 the following JS injection code works fine for deleting the service worker stuff, though it sometimes requires to restart the webapp a few times after "updates". it's what the previous version of whatsapp in pi-apps used.
> 
> ```js
> if ('serviceWorker' in navigator) {
>     caches.keys().then(function (cacheNames) {
>         cacheNames.forEach(function (cacheName) {
>             caches.delete(cacheName);
>         });
>     });
> } 
> ```
> 
> (I have no idea how it works, I got it from here: [nativefier/nativefier#1112 (comment)](https://github.com/nativefier/nativefier/issues/1112#issuecomment-795424961))
> also using a firefox or safari user agent helps sometimes.

Some nativefier settings seems not working after second launch sometimes, including `inject` and `user-agent`. 
You can't know whether the JS code has been injected. But my script ensure the `Service Worker` folder deleted before launching WhatsApp ( using && operator ) .

But the weird thing is, the js script seems work in **multi-device beta**. It won't have the prompt like this:

image

Still finding out why.
If it really works, maybe add a prompt in installation script asking users whether they are using normal mode or multi-device beta ( however I think the multi-device function will be merged to the normal mode very soon ).

PR #966, opened 2021-09-23
Change ExecStart directory to /opt/sysmon/main.py

cycool29:
From the script, it cloned the repo to `/opt` but the ExecStart points to home folder.


Itai-Nelken:
Thanks!

Issue #156, opened 2020-12-24
Telegram (app request)


Aradionov366:
Yes, I know that you can install it using the command 'sudo install telegram-desktop', but the version you install will be too old, I would like to see a new version of this messenger. Thank you in advance :D

Itai-Nelken:
@Aradionov366 you can install it from snap (https://snapcraft.io/telegram-desktop there are armhf and arm64 versions), flatpak (https://flathub.org/apps/details/org.telegram.desktop not sure there are arm versions though) and if you can compile it yourself (https://github.com/telegramdesktop/tdesktop/blob/dev/docs/building-cmake.md) (as you said you can also install it with `sudo apt install telegram-desktop`).

I don't think that a app that's available on so many places can be on pi-apps

Botspot:
I still haven't decided on this. Issue will remain open until I make a decision.

Botspot:
It appears that you can install telegram via Flatpak:
```
sudo apt install flatpak
flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
flatpak install flathub org.telegram.desktop
```
To run:
```
/var/lib/flatpak/exports/share/applications/org.telegram.desktop.desktop
```
@Aradionov366, could you please compare the Flatpak version to the APT version and convince me that the Flatpak version is indeed better/necessary?

Itai-Nelken:
> It appears that you can install telegram via Flatpak:
> 
> ```
> sudo apt install flatpak
> flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
> flatpak install flathub org.telegram.desktop
> ```
> 
> To run:
> 
> ```
> /var/lib/flatpak/exports/share/applications/org.telegram.desktop.desktop
> ```
> 
> @Aradionov366, could you please compare the Flatpak version to the APT version and convince me that the Flatpak version is indeed better/necessary?

@Aradionov366 
reboot after installing flatpak, and the menu button will appear.
also if it doesn't, you can run using the command @Botspot sent, or this command:
```
flatpak run telegram-desktop
```

Botspot:
Closing this issue, as @Aradionov366 has not replied explaining **why** the `apt-get` version of Telegram is sub-optimal.

fabianmendes:
Please reopen this issue. Snap installation version has conflicts, also, Menu Icon is not set.
[...]

PR #971, opened 2021-09-25
Add Telegram app

cycool29:
Actually installable with `apt`, but just hope to have a more convenient way to install it.


Itai-Nelken:
Apt version is old and isn't fully functional.
Also as those scripts simply install the apt version, I don't think this should be added.
Also about telegram see #156

cycool29:
I don't know why, all `armhf` deb after version `1.5.11` [here](http://ftp.us.debian.org/debian/pool/main/t/telegram-desktop/) will popup error with a long dependencies hell. I think the only way to install Telegram desktop with full features and best performance is to install `snapd` as dependencies in pi-apps script.

Itai-Nelken:
@cycool29 could you try to use the flatpak version instead of the snap?
snapd seems to have a lot of problems installing on some systems, at least according to the pi-apps error reports.
so using flatpak is a better choice in my opinion. also I find flatpak's slightly faster than snaps.

cycool29:
> @cycool29 could you try to use the flatpak version instead of the snap?
> snapd seems to have a lot of problems installing on some systems, at least according to the pi-apps error reports.
> so using flatpak is a better choice in my opinion. also I find flatpak's slightly faster than snaps.

Done. Ya, it is slightly faster than snap one.

Itai-Nelken:
There is a problem with adding a desktop shortcut as flatpak already does that but it won't show until a reboot.
So maybe add a script to remove the one the `install` script creates on reboot (with cron maybe?) and then delete itself.

cycool29:
Actually flatpak store its desktop shortcut in `/var/lib/flatpak/app/org.telegram.desktop/current/active/export/share/applications/org.telegram.desktop.desktop` , but I don't know why it won't appear in the menu even after reboot.  But copying it to `~/.local/share/applications` and change the icon (as the system can't detect the icon without modifying `XDG_DATA_DIR` variable). So I will just add that in the script.

Itai-Nelken:
Thanks. make sure to move it so it won't appear twice for some systems like mine where after a reboot the flatpak one shows up.
if you do that check that the app still uninstall correctly.

cycool29:
Instead of keep `sed`ing the icon and name in the desktop entry, I decided to directly make a new file in `~/.local/share/applications` and echo the full desktop entry in it. 

> Thanks. make sure to move it so it won't appear twice for some systems like mine where after a reboot the flatpak one shows up.

Are you using Raspbian? Mine don't even show up.

cycool29:
Done.
@Itai-Nelken Can you test it?

cycool29:
Done. Works well for me.

Botspot:
@cycool29 I've made a few improvements to the scripts that address a few edge-cases.

PR #974, opened 2021-09-27
Fix reduceapt not working on sudo apt update in `pkg-install`

cycool29:
Just `reduceapt` first before `tee` to `/dev/stderr`.


cycool29:
@Botspot

PR #978, opened 2021-09-28
Use Nodesource to install Node.js

cycool29:
Using `nvm` to install nodejs is terrible for me. It slowed down my terminal loading time (there are few issues about this: nvm-sh/nvm#1277, nvm-sh/nvm#1261). Although there are many solution on the web like editing `.bashrc`, modifying nvm install script, etc, I don't want to make it so confusing...

Then Nodesource does it **perfectly**. 
- It won't slow down terminal loading speed.
- It will make pi-apps easy to update Node.js with only changing the nodesource link, but with `nvm`, pi-apps will probably hard to update as it need to find out what version of nodejs user have, what is the latest version does nvm have, etc. 
- Also, Node.js from nodesource there is a [proper way](https://github.com/nodesource/distributions/issues/512) to uninstall with just `sudo apt purge nodejs` and remove source list and apt-key. On `nvm`, you will need to delete the folders one-by-one.

However, the [snap one](https://snapcraft.io/node) is also good, but I prefer nodesource since it don't have any dependencies.

### DIFF
--- a/apps/Node.js/install
+++ b/apps/Node.js/install
@@ -24,8 +24,8 @@ export NVM_DIR="$HOME/.nvm"
 mkdir -p "$NVM_DIR"
 wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash || error "Failed to install nvm!"
 
-#determine if host system is 64 bit arm64 or 32 bit armhf
-if [ ! -z "$(file "$(readlink -f "/sbin/init")" | grep 32)" ];then
+source "${DIRECTORY}/api"
+if [ "$arch" == 32 ];then
   #armhf, so patch nvm script to forcibly use armhf
   sed -i 's/^  nvm_echo "${NVM_ARCH}"/  NVM_ARCH=armv7l ; nvm_echo "${NVM_ARCH}"/g' "$NVM_DIR/nvm.sh"
 fi
--- a/apps/Node.js/install
+++ b/apps/Node.js/install
@@ -9,10 +9,10 @@ function error {
 
 # Uninstall nodejs, node and npm from other sources first
 echo "Uninstall nodejs, node and npm from other sources first ..."
-sudo snap remove node --purge > /dev/null
-sudo rm -rf $NVM_DIR ~/.npm ~/.bower > /dev/null
-sed -i '/NVM_DIR/d' ~/.bashrc  > /dev/null
-sudo apt purge node npm nodejs -y  > /dev/null # I can't find a script in pi-apps to uninstall package in installation script 
+sudo snap remove node --purge 2&> /dev/null
+sudo rm -rf $NVM_DIR ~/.npm ~/.bower 2&> /dev/null
+sed -i '/NVM_DIR/d' ~/.bashrc  2&> /dev/null
+sudo apt purge node npm nodejs -y  2&> /dev/null
 
 # Add Nodesource  apt key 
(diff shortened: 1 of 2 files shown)


Itai-Nelken:
The version from the nodesource repo was used, but it bricked some systems including mine (broke permissions) so we changed to nvm.

cycool29:
> The version from the nodesource repo was used, but it bricked some systems including mine (broke permissions) so we changed to nvm.

How about using the [snap one](https://snapcraft.io/node)? I don't really think nvm is a good choice (especially in terminal loading speed).

I tried other method like [`nvs`](https://github.com/jasongin/nvs/),  [`nodenv`](https://github.com/nodenv/nodenv), [`fnm`](https://github.com/Schniz/fnm), all will slow down terminal speed as they will need to source their script before launching the terminal. But [`n`](https://github.com/tj/n) works pretty well, it won't slow down terminal speed. However, snap and nodesource is still best for me.

cycool29:
@Botspot What do you think?

Botspot:
> How about using the [snap one](https://snapcraft.io/node)? I don't really think nvm is a good choice (especially in terminal loading speed).

The slow terminal start-up time from `nvm` bothers me too. But it's reliable. We commonly receive snap error reports; we very rarely receive Node.js error reports. It would be a pity to switch to a less reliable installation method - sure it would benefit some users and their terminal start-up time, but it would completely break node-js on other user's systems.

So, while I'm open to a better alternative, I don't believe snap is the right option due to its reliably issues.

cycool29:
Since nodesource and snap is not reliable and stable, I am searching for ways to increase `nvm` initialize time.
Finally, this works for me:
```
# Defer initialization of nvm until nvm, node or a node-dependent command is
# run. Ensure this block is only run once if .bashrc gets sourced multiple times
# by checking whether __init_nvm is a function.
if [ -s "$HOME/.nvm/nvm.sh" ] && [ ! "$(type -t __init_nvm)" = function ]; then
  export NVM_DIR="$HOME/.nvm"
  [ -s "$NVM_DIR/bash_completion" ] && . "$NVM_DIR/bash_completion"
  declare -a __node_commands=('nvm' 'node' 'npm' 'yarn' 'gulp' 'grunt' 'webpack')
  function __init_nvm() {
    for i in "${__node_commands[@]}"; do unalias $i; done
    . "$NVM_DIR"/nvm.sh
    unset __node_commands
    unset -f __init_nvm
  }
  for i in "${__node_commands[@]}"; do alias $i='__init_nvm && '$i; done
fi
```

I got it from here : https://www.growingwiththeweb.com/2018/01/slow-nvm-init.html

I will add to the script now.

cycool29:
Done. @Itai-Nelken @Botspot 
Tested installing & uninstalling in pi-apps. Now the terminal loading time increased!

Itai-Nelken:
This is a good idea, the problem is that there are much more node dependant commands, and adding all of them makes no sense.
I think we should give the user the choice to add this, maybe someone uses a lot of node stuff and doesn't care how fast bash starts as long as he can use all of them?

Botspot:
> This is a good idea, the problem is that there are much more node dependant commands, and adding all of them makes no sense. I think we should give the user the choice to add this, maybe someone uses a lot of node stuff and doesn't care how fast bash starts as long as he can use all of them?

I did consider this factor, but figured that if someone asks for help we can solve it on a case-by-case basis, reconsidering as necessary going forward.

PR #979, opened 2021-09-29
Add reduceapt function 

cycool29:
Added `reduceapt` function in every app's install & uninstall script that using `apt` to remove unwanted lines from `apt` output.
Added in `purge-installed` script too.

### DIFF
--- a/apps/Angry IP scanner/uninstall
+++ b/apps/Angry IP scanner/uninstall
@@ -7,7 +7,11 @@ function error {
   exit 1
 }
 
-sudo apt purge -y ipscan
+reduceapt() { #remove unwanted lines from apt output
+  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
+}
+
+sudo apt purge -y ipscan 2>&1 | reduceapt
 
 #if your app installs any packages, keep this command here so those packages will be removed.
 "${DIRECTORY}/purge-installed" "$(dirname "$0")" || exit 1
--- a/apps/Downgrade Chromium/uninstall
+++ b/apps/Downgrade Chromium/uninstall
@@ -7,6 +7,10 @@ function error {
   exit 1
 }
 
+reduceapt() { #remove unwanted lines from apt output
+  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
+}
+
 if [ ! -z "$(ps aux | grep chromium | grep -v 'grep' )" ];then
   echo -n "Please close all Chromium windows first."
   yad --info --center --window-icon="$(dirname "$0")/icon-64.png" \
@@ -52,6 +56,6 @@ echo "chromium-codecs-ffmpeg-extra install" | sudo dpkg --set-selections
 
 echo "Upgrading Chromium..."
 "${DIRECTORY}/api" apt_lock_wait
-sudo apt upgrade -y --allow-downgrades chromium-browser chromium-browser-l10n chromium-codecs-ffmpeg-extra
+sudo apt upgrade -y --allow-downgrades chromium-browser chromium-browser-l10n chromium-codecs-ffmpeg-extra 2>&1 | reduceapt
 
 
--- a/purge-installed
+++ b/purge-installed
@@ -11,6 +11,10 @@ function error {
   exit 1
 }
 
+reduceapt() { #remove unwanted lines from apt output
+  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
+}
+
 DIRECTORY="$(readlink -f "$(dirname "$0")")"
 
 echo -e "\e[97m\nRunning purge-installed...\e[39m"
@@ -34,7 +38,7 @@ if dpkg -l pi-apps-$appnamehash &>/dev/null ;then
   #new pkg-install implementation - using dummy debs
   
   echo -e "\e[97m\nRemoving dummy deb for $app...\e[39m"
-  sudo -E apt purge -y pi-apps-$appnamehash --autoremove || error "apt failed to purge dummy deb (pi-apps-$appnamehash)!"
+  sudo -E apt purge -y pi-apps-$appnamehash --autoremove 2>&1 | reduceapt || error "apt failed to purge dummy deb (pi-apps-$appnamehash)!"
   
 elif [ -f "${DIRECTORY}/data/installed-packages/${app}" ];then
   #old pkg-install implementation
--- a/apps/Angry IP scanner/install
+++ b/apps/Angry IP scanner/install
@@ -7,9 +7,13 @@ function error {
   exit 1
 }
 
+reduceapt() { #remove unwanted lines from apt output
+  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
+}
+
 cd $HOME
 wget https://github.com/angryip/ipscan/releases/download/3.7.6/ipscan_3.7.6_all.deb
 # Get dependencies
 "${DIRECTORY}/pkg-install" "openjdk-11-jdk rpm fakeroot" "$(dirname "$0")" || exit 1
-sudo apt --fix-broken -y install ~/ipscan_3.7.6_all.deb
+sudo apt --fix-broken -y install ~/ipscan_3.7.6_all.deb 2>&1 | reduceapt
 rm ~/ipscan_3.7.6_all.deb
--- a/apps/Box86/install-32
+++ b/apps/Box86/install-32
@@ -7,8 +7,12 @@ function error {
   exit 1
 }
 
+reduceapt() { #remove unwanted lines from apt output
+  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
+}
+
 if dpkg -l box86 &>/dev/null ;then
-  sudo apt purge -y box86
+  sudo apt purge -y box86 2>&1 | reduceapt
 fi
 
 sudo rm -f /etc/apt/sources.list.d/box86.list
--- a/apps/Clam Antivirus/install
+++ b/apps/Clam Antivirus/install
@@ -9,17 +9,21 @@ function error {
   exit 1
 }
 
+reduceapt() { #remove unwanted lines from apt output
+  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
+}
+
 # Get package clamav
 "${DIRECTORY}/pkg-install" "clamav" "$(dirname "$0")" || exit 1
 
 wget https://github.com/dave-theunsub/clamtk/releases/download/v6.13/clamtk_6.13-1_all.deb
-sudo apt install --fix-broken ./clamtk_*.deb -y
+sudo apt install --fix-broken ./clamtk_*.deb -y 2>&1 | reduceapt 
 rm clamtk_*.deb
 
 #if thunar file manager installed then also install clamtk thunar extension
 if [ -f /usr/bin/thunar ];then
   wget https://bitbucket.org/davem_/thunar-sendto-clamtk/downloads/thunar-sendto-clamtk_0.06-1_all.deb
-  sudo apt install ./thunar-sendto-clamtk_0.06-1_all.deb -y
+  sudo apt install ./thunar-sendto-clamtk_0.06-1_all.deb -y 2>&1 | reduceapt
   rm ./thunar-sendto-clamtk_0.06-1_all.deb
 fi
 exit 0
--- a/apps/Clam Antivirus/uninstall
+++ b/apps/Clam Antivirus/uninstall
@@ -6,11 +6,15 @@ function error {
   exit 1
 }
 
+reduceapt() { #remove unwanted lines from apt output
+  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
+}
+
 #if your app installs any packages, keep this command here so those packages will be removed.
 "${DIRECTORY}/purge-installed" "$(dirname "$0")" || exit 1
 
 if [ -f /usr/bin/thunar ];then
-  sudo apt purge thunar-sendto-clamtk -y
+  sudo apt purge thunar-sendto-clamtk -y 2>&1 | reduceapt
 fi
 
 exit 0
--- a/apps/Downgrade Chromium/install-32
+++ b/apps/Downgrade Chromium/install-32
@@ -7,6 +7,10 @@ function error {
   exit 1
 }
 
+reduceapt() { #remove unwanted lines from apt output
+  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
+}
+
 #variable for all chromium versions supported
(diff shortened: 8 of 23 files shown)


Itai-Nelken:
Why not make pi-apps override the apt command? It would be simpler.

cycool29:
> Why not make pi-apps override the apt command? It would be simpler.

You mean like how pi-apps convert `wget` to `aria2c` ? 
I don't really understand how pi-apps convert `wget` to `aria2c`, if there is a way to do so on `apt` command, that will be great !
Maybe something like this:
```
apt() {
	actions = "$*"
	( sudo apt $actions ) 2>&1 | reduceapt

```

However, I think directly modifying the scripts will be more straight-forward way.

Botspot:
I've been gradually transitioning all apps from `apt` to the `pkg-install` script. Basically, if an app fails due to an apt error, I transition it.
@cycool29, Looks like you found all the ones I haven't transitioned yet!

Using reduceapt by itself will only **reduce** apt's output. That won't **diagnose** apt-errors for the user, which is the important thing.
By the way, unless I'm mistaken, none of these scripts will work, because the `reduceapt()` function is not in the `api` script, it's in the `pkg-install` script.

cycool29:
> I've been gradually transitioning all apps from apt to the pkg-install script. Basically, if an app fails due to an apt error, I transition it.
@cycool29, Looks like you found all the ones I haven't transitioned yet!

Lol, I just hate seeing those `WARNING: apt does not have a stable cli interface. Use with caution in scripts` and I just search in the repo, append `reduceapt` to all of them .

> By the way, unless I'm mistaken, none of these scripts will work, because the reduceapt() function is not in the api script, it's in the pkg-install script. (though that's easy to change)

Well, I have defined `reduceapt` function in those script.

Botspot:
> Lol, I just hate seeing those `WARNING: apt does not have a stable cli interface. Use with caution in scripts` and I just search in the repo, append `reduceapt` to all of them .

That warning appears because Pi-Apps appends the script's output to a logfile, and to the terminal. APT thinks stdout is not a terminal. @cycool29, if you know how to solve this (how to make apt think stdout is a terminal), that would be great.

cycool29:
@Botspot I think there should be a `pkg-purge` script, just like `pkg-install`, but just to purge packages in installation script to avoid conflicts. ( like removing `steamlink` when installing Steam because it conflicts with steam-launcher )

>  @cycool29, if you know how to solve this (how to make apt think stdout is a terminal), that would be great.

I am searching a whole day for this and I found `unbuffer` seems able to do that. You can try it by comparing `sudo apt update | grep -v abc` and `unbuffer sudo apt update | grep -v abc`. Make sure you have `expect` installed: `sudo apt install expect`.

Botspot:
> I found `unbuffer` seems able to do that. Make sure you have `expect` installed: `sudo apt install expect`.

Interesting. That worked for me. Now the question becomes, "Do we really want to another dependency to Pi-Apps for *this*?"

cycool29:
> Interesting. That worked for me. Now the question becomes, "Do we really want to another dependency to Pi-Apps for this?"

That's why, if that is capable with `reduceapt` (grep -v), there should not be another dependency to pi-apps.
Hence directly add `reduceapt` to the install & uninstall script should be the more straight-forward and effective way.

Maybe add `reduceapt` function in `api` script and append these text in the [Create app wiki](https://github.com/Botspot/pi-apps/wiki/Creating-an-app):

> If you are using `apt` in your script for some reason, add ` | reduceapt` at the end of your line. This will remove the unwanted lines from `apt`.
If you are using `sudo apt purge {package}`, use `sudo apt purge {package} | reduceapt` instead.
If you want to have a error message, use `sudo apt purge {package} | reduceapt || error "Your error message here"`

Jai-JAP:
Adding `apt` function and `reduceapt` function to api will work.

```
reduceapt() { #remove unwanted lines from apt output
  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
}
    
apt() {
    if [ $(id -u) -ne 0 ]; then
      sudo -E apt $@ 2>&1| reduceapt
    else
      "apt" $@ 2>&1 | reduceapt
    fi
}
```

Jai-JAP:
But using `pkg-install` is the best option as it also diagnoses apt errors
(Moreover It has `reduceapt` lol)

cycool29:
> But using `pkg-install` is the best option as it also diagnoses apt errors (Moreover It has `reduceapt` lol)

+1.
So just change all the `apt install` to `pkg-install` and add `reduceapt` function only in `apt purge` script.

cycool29:
> @Botspot I think there should be a pkg-purge script, just like pkg-install, but just to purge packages in installation script to avoid conflicts. ( like removing steamlink when installing Steam because it conflicts with steam-launcher )

Botspot:
> > @Botspot I think there should be a pkg-purge script, just like pkg-install, but just to purge packages in installation script to avoid conflicts. ( like removing steamlink when installing Steam because it conflicts with steam-launcher )
> 
> If this is possible, we can just add `reduceapt` in `pkg-purge` and replace all `apt purge` to `pkg-purge`

Or... simply suppress all output like QEMU does.
```
#Remove any qemu packages first to prevent conflicting packages
sudo apt purge -yqq qemu* &>/dev/null
```

cycool29:
> > > @Botspot I think there should be a pkg-purge script, just like pkg-install, but just to purge packages in installation script to avoid conflicts. ( like removing steamlink when installing Steam because it conflicts with steam-launcher )
> > 
> > 
> > If this is possible, we can just add `reduceapt` in `pkg-purge` and replace all `apt purge` to `pkg-purge`
> 
> Or... simply suppress all output like QEMU does.
> 
> ```
> #Remove any qemu packages first to prevent conflicting packages
> sudo apt purge -yqq qemu* &>/dev/null
> ```

It is an option too, but that will be hard to diagnose if there is any error during uninstallation.

cycool29:
Well, I have updated all scripts using `apt install` to use `pkg-install`. 
Now left `apt purge`, should we remove all output from `apt purge` or just remove unwanted lines?

Jai-JAP:
```
reduceapt() { #remove unwanted lines from apt output
  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
}

pkg-purge() {
    purge_list="$1" #quotation-marked, space-seperated list of package names to install
    app=$(basename "$2")
    echo $purge_list > "${2}/../../data/purged-packages/${app}"
    if [ $(id -u) -ne 0 ]; then 
      sudo -E apt purge $1 2>&1| reduceapt
    else
      apt purge $1 2>&1 | reduceapt
    fi
}

restore-purged() {
    app=$(basename "$2")
    if [[ -f "${2}/../../data/purged-packages/${app}" -a -n $(cat "${2}/../../data/purged-packages/${app}") ]] ; then
      if [ $(id -u) -ne 0 ]; then 
        sudo -E apt install $1 2>&1| reduceapt
      else
        apt install $1 2>&1| reduceapt
      fi
    fi
}
```


Syntax
pkg-purge "package1 package2" "${DIRECTORY}"

cycool29:
> ```
> reduceapt() { #remove unwanted lines from apt output
>   grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
> }
> 
> pkg-purge() {
>     purge_list="$1" #quotation-marked, space-seperated list of package names to install
>     app=$(basename "$2")
>     echo $purge_list > "${2}/../../data/purged-packages/${app}"
>     if [ $(id -u) -ne 0 ]; then 
>       sudo -E apt purge $1 2>&1| reduceapt
>     else
>       apt purge $1 2>&1 | reduceapt
>     fi
> }
> 
> restore-purged() {
>     app=$(basename "$2")
>     if [[ -f "${2}/../../data/purged-packages/${app}" -a -n $(cat "${2}/../../data/purged-packages/${app}") ]] ; then
>       if [ $(id -u) -ne 0 ]; then 
>         sudo -E apt install $1 2>&1| reduceapt
>       else
>         apt install $1 2>&1| reduceapt
>       fi
>     fi
> }
> ```
> 
> Syntax: pkg-purge "package1 package2" "${DIRECTORY}" 
> restore-purged "${DIRECTORY}"

The code works for me!

Btw, I think the `pkg-purge` and `restore-purged` should be written as a script instead of a function as `pkg-install` and `purge-installed` is written in script and it may make confusion to app creators.

Jai-JAP:
The above is only the code we can add it either to a script or a function after modifying as required

cycool29:
`pkg-purge`
```
#!/bin/bash
{ #prevents errors if script was modified while in use
set -o pipefail #if any command within a pipe fails, make the whole pipe fail. Necessary for error-catching with reduceapt
DIRECTORY="$(readlink -f "$(dirname "$0")")"
trap "exit 1" TERM
export TOP_PID=$$

error() {
  echo -e "\e[91m$1\e[39m" 1>&2
  kill -w -s TERM $TOP_PID
  exit 1
}

PURGE_LIST="$1" #quotation-marked, space-seperated list of package names to install
app="$(basename "$2")" #remove any slashes to just get program name

reduceapt() { #remove unwanted lines from apt output
  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
}

echo "Running pkg-purge" 
if [ -z "$PURGE_LIST" ];then
  error "No package names specified to pkg-purge!"
elif [ -z "$app" ];then
  error "No app name specified to pkg-purge!"
fi


source "${DIRECTORY}/api"
apt_lock_wait

 
output="$(sudo -E LANG=C LC_ALL=C LANGUAGE=C apt purge $1 --autoremove -y 2>&1 | reduceapt)"
exitcode=$?

errors="$(echo "$output" | grep '^[(W)|(E)|(Err]:')"
if [ $exitcode != 0 ] || [ ! -z "$errors" ];then
  echo -e "\e[91mFailed to uninstall the packages!\e[39m"
  echo -e "APT reported these errors:\n\e[91m$errors\e[39m"
  exit 1
else
  echo $PURGE_LIST > "${DIRECTORY}/data/purged-packages/${app}"
fi 
}

```

Jai-JAP:
`restore-purged`

```
#!/bin/bash
{ #prevents errors if script was modified while in use
set -o pipefail #if any command within a pipe fails, make the whole pipe fail. Necessary for error-catching with reduceapt
DIRECTORY="$(readlink -f "$(dirname "$0")")"
trap "exit 1" TERM
export TOP_PID=$$

error() {
  echo -e "\e[91m$1\e[39m" 1>&2
  kill -w -s TERM $TOP_PID
  exit 1
}

app="$(basename "$1")" #remove any slashes to just get program name
PURGED_LIST="$(cat "${DIRECTORY}/data/purged-packages/${app}")" #get the list of purged-packages for $app

reduceapt() { #remove unwanted lines from apt output
  grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
}

echo "Running restore-purged" 
if [ ! -f "${DIRECTORY}/data/purged-packages/${app}" ]; then
  error "purged-packages data file not found for ${app}"
elif [ -z "$PURGED_LIST" ]; then
  error "No packages found to restore-purged!"
elif [ -z "$app" ];then
  error "No app name specified to restore-purged!"
fi


source "${DIRECTORY}/api"
apt_lock_wait

 
output="$(sudo -E LANG=C LC_ALL=C LANGUAGE=C apt install $PURGED_LIST -y 2>&1 | reduceapt)"
exitcode=$?

errors="$(echo "$output" | grep '^[(W)|(E)|(Err]:')"
if [ $exitcode != 0 ] || [ ! -z "$errors" ];then
  echo -e "\e[91mFailed to install the packages!\e[39m"
  echo -e "APT reported these errors:\n\e[91m$errors\e[39m"
  exit 1
else
  rm -f "${DIRECTORY}/data/purged-packages/${app}"
fi 
}
```

cycool29:
I have added `pkg-purge` and replaced all app scripts' `apt` with either `pkg-install` or `pkg-purge` **except** `Downgrade Chromium` as it uses `sudo apt upgrade` and it can be replace with neither `pkg-install` nor `pkg-purge`.

But for `restore-purged`, I am still considering because:
- We don't know if user still need the app
- Make confusions to users (e.g. `Box8`6 in pi-apps detected an `Box86` originally installed in system when running the installation script so removed it with `pkg-purge`, but when uninstallation it purged pi-apps `Box86` **but** reinstalled the original `Box86` )

Jai-JAP:
We can add a remember purged packages option to `pkg-purge`

Add `remember_purged=$2`
Change `app=$3`
Replace
```
echo $PURGE_LIST > ${DIRECTORY}/data/purged_packages/${app}
```
with
```
if [[ $remember_purged == 1 ]]; then
  echo $PURGE_LIST > ${DIRECTORY}/data/purged_packages/${app}
fi
```

so that pkg-purge can be used in uninstall scripts without saving `purged_packages`

apps/Angry IP scanner/uninstall
```diff
@@ -11,7 +11,7 @@ reduceapt() { #remove unwanted lines from apt output
   grep -v "apt does not have a stable CLI interface.\|Reading package lists...\|Building dependency tree\|Reading state information...\|Need to get\|After this operation,\|Get:\|Fetched\|Selecting previously unselected package\|Preparing to unpack\|Unpacking \|Setting up \|Processing triggers for "
 }
 
-sudo apt purge -y ipscan 2>&1 | reduceapt
+"${DIRECTORY}/pkg-purge" "ipscan" "$(dirname "$0")" || error "Failed to purge Angry IP Scanner!"
```
Jai-JAP, line 10: This line is not needed as purge-installed will do the work here
This line can be removed

Jai-JAP (changes requested):

apps/Angry IP scanner/uninstall
cycool29: Oops, I forgot to remove it.

Jai-JAP:
Also `restore-purge` must be added in Firefox Rapid Release uninstall script as `firefox-esr` must be restored if installed previously

restore-purged
```diff
@@ -0,0 +1,46 @@
...
+}
+
+echo "Running restore-purged" 
+if [ ! -f "${DIRECTORY}/data/purged-packages/${app}" ]; then
+  error "purged-packages data file not found for ${app}"
+elif [ -z "$PURGED_LIST" ]; then
+  error "No packages found to restore-purged!"
+elif [ -z "$app" ];then
+  error "No app name specified to restore-purged!"
+fi
```
Jai-JAP, line 28: Replace error with echo as script must display an error and go on without stopping in above cases
Also add red color to error messages to echo error in red but not stop

Jai-JAP (changes requested):

cycool29:
@Botspot @Itai-Nelken What do you think?

apps/Downgrade Chromium/uninstall
```diff
@@ -7,6 +7,10 @@ function error {
   exit 1
 }
 
+reduceapt() { #remove unwanted lines from apt output
```
Itai-Nelken, line 10: There is no need to declare `reduceapt` here, as `api` functions are available to the scripts declaring `reduceapt` in the `api` script makes more sense.

pkg-purge
```diff
@@ -0,0 +1,45 @@
...
+error() {
+  echo -e "\e[91m$1\e[39m" 1>&2
+  kill -w -s TERM $TOP_PID
+  exit 1
+}
+
+PURGE_LIST="$1" #quotation-marked, space-seperated list of package names to install
+app="$(basename "$2")" #remove any slashes to just get program name
+
+reduceapt() { #remove unwanted lines from apt output
```
Itai-Nelken, line 17: I think it's better to declare `reduceapt` only once in `api`. that makes changing it much easier.

restore-purged
```diff
@@ -0,0 +1,46 @@
...
+error() {
+  echo -e "\e[91m$1\e[39m" 1>&2
+  kill -w -s TERM $TOP_PID
+  exit 1
+}
+
+app="$(basename "$1")" #remove any slashes to just get program name
+PURGED_LIST="$(cat "${DIRECTORY}/data/purged-packages/${app}")" #get the list of purged-packages for $app
+
+reduceapt() { #remove unwanted lines from apt output
```
Itai-Nelken, line 17: same.

purge-installed
```diff
@@ -11,6 +11,10 @@ function error {
   exit 1
 }
 
+reduceapt() { #remove unwanted lines from apt output
```
Itai-Nelken, line 14: same

Itai-Nelken (changes requested):
Declaring `reduceapt` over and over again is a bad idea, every time we wan't to change it we need to hunt down every script that declares it. instead simply declare it in the `api` script. `api` functions are available to install/uninstall scripts. what do you think @Botspot.

restore-purged
Itai-Nelken: red: `\e[1m` normal: `\e[0m`

Botspot:
> Declaring `reduceapt` over and over again is a bad idea, every time we wan't to change it we need to hunt down every script that declares it. instead simply declare it in the `api` script. `api` functions are available to install/uninstall scripts. what do you think @Botspot.

I completely agree. Honestly I have not been monitoring this PR too closely - I've been re-writing large portions of pi-apps, in order to solve this issue in a far more elegant way.

cycool29:
> This PR might not be merged, however. I've been working on a few major improvements and they will address these issues by converting all app-scripts to an abstract scripting model.

As this, closing.

PR #984, opened 2021-09-30
Fix typo in Clam Antivirus install script

cycool29:
Fix `dpkg-deb: error: failed to read archive '/home/pi.clamtk.deb': No such file or directory`


Itai-Nelken:
Thanks!

PR #989, opened 2021-10-02
Make sure to install libssl with git

mobilegmYT:
Some users have libssl1.0-dev installed for some reason which breaks git cloning. Originally found this issue in https://discord.com/channels/770629697909424159/874656141475463229/893896367624294420 but I'm adding this to the install script itself just in case.


theofficialgman:
closing as I think this was a temporary issue and this is not compatible anymore anyway

PR #990, opened 2021-10-02
Add apt lock checker call from api to lightpad

mobilegmYT:


Botspot:
The `pkg-install` script already does `apt_lock_wait`.

PR #1000, opened 2021-10-03
Update prusaslicer to 2.4

mobilegmYT:


mobilegmYT:
BTW, I think RPi OS currently uses GTK3, but if its GTK2 please tell me

Botspot:
> BTW, I think RPi OS currently uses GTK3, but if its GTK2 please tell me

It's GTK3.

PR #1002, opened 2021-10-04
cli improvements

cycool29:
Add some `echo` to scripts and replace most `echo` with new api message functions: `status`, `status_green` and `warning`.

Fix some minor bugs.

- Use `\e[0m` as closing for `echo -e` 
- Add few `echo` to create newlines to make output neater
- Replace most `echo` with new api message functions
- Check app name before update in `manage`
- Show error messages if update failed in `manage` 
- Add `reduceapt` in `purge-installed`
- Show `No apps/files is updatable` if there isn't any app/files to update in `updater`

### DIFF
--- a/manage
+++ b/manage
@@ -218,10 +218,10 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
     echo
     exit 0
   else
-    error "\nFailed to ${mode} ${app} with $scriptname script"\!"\e[39m
+    echo -e  "\n\e[91mFailed to ${mode} ${app} with $scriptname script"\!"\e[39m
 \e[40m\e[93m\e[5m▲\e[25m\e[39m \e[49m\e[93mNeed help? Copy the \e[1mENTIRE\e[0m\e[49m\e[93m terminal output or take a screenshot.
 Please ask on Github: \e[94m\e[4mhttps://github.com/Botspot/pi-apps/issues/new/choose\e[24m\e[93m
-Or on Discord: \e[94m\e[4mhttps://discord.gg/RXSTvaUvuu" | tee -a "$logfile"
+Or on Discord: \e[94m\e[4mhttps://discord.gg/RXSTvaUvuu\e[0m" | tee -a "$logfile"
     
     format_log_file "$logfile" #remove escape sequences from logfile
     mv "$logfile" "$(echo "$logfile" | sed 's+-incomplete-+-fail-+g')"
@@ -303,7 +303,7 @@ elif [ "$1" == 'update' ];then
   fi
   
   if [ "$failed" == 'true' ]; then
-    error "Failed to update ${app}."
+    echo -e "\e[91mFailed to update ${app}.\e[0m"
   else
     status_green "${app} was updated successfully."
   fi
--- a/manage
+++ b/manage
@@ -73,6 +73,8 @@ else
   supported=no
 fi
 
+
+
 if [ "$1" == 'multi-uninstall' ] || [ "$1" == 'multi-install' ];then
   
   if [ "$1" == 'multi-uninstall' ];then
@@ -80,6 +82,8 @@ if [ "$1" == 'multi-uninstall' ] || [ "$1" == 'multi-install' ];then
   elif [ "$1" == 'multi-install' ];then
     action=install
   fi
+  
+  echo
   
   app_list="$2" #newline-separated list of apps to install/uninstall
   
@@ -127,6 +131,8 @@ $app"
   
 elif [ "$1" == 'install-if-not-installed' ];then
   
+  echo
+  
   #if not installed
   if [ "$(app_status "$2")" != installed ];then
     #install it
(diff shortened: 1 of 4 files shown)


updater
```diff
@@ -430,50 +434,84 @@ elif [ "$runmode" == cli ];then #return list of updatable apps, and ask the user
...
+    echo
+  else
+    status "These files can be updated:"
+    echo
+    echo "$updatable_files" | sed 's/^/  - /g'
+    echo
+  fi
   
-  read -p "Update now? [Y/n] " answer
+  question=$(status 'Update now? [Y/n] ')
```
Botspot, line 464: Not sure why you're placing the question into an array.

manage
```diff
@@ -196,37 +203,47 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
...
-    echo -e "\n\e[42m\e[30m${mode^}ed ${app} successfully.\e[39m\e[49m" | tee -a "$logfile"
+    status_green "\n${mode^}ed ${app} successfully." | tee -a "$logfile"
     
     format_log_file "$logfile" #remove escape sequences from logfile
     mv "$logfile" "$(echo "$logfile" | sed 's+-incomplete-+-success-+g')" #
+    echo
     exit 0
   else
-    echo -e "\n\e[91mFailed to ${mode} ${app} with $scriptname script"\!"\e[39m
+    error "\nFailed to ${mode} ${app} with $scriptname script"\!"\e[39m
```
Botspot, line 221: This breaks the script because the `error ` function runs `exit 1`.

manage
```diff
@@ -275,13 +293,26 @@ elif [ "$1" == 'update' ];then
   cp -rf "${DIRECTORY}/update/pi-apps/apps/${app}" "${DIRECTORY}/apps/${app}"
   
   if [ "$installback" == 'yes' ];then
-    echo "$app was originally installed before updating. Reinstalling the new version now."
```
Botspot, line 278: Why was this line removed?

manage
```diff
@@ -275,13 +293,26 @@ elif [ "$1" == 'update' ];then
...
     "${DIRECTORY}/manage" install "$app"
+    if [ $? == 0 ]; then
+      failed=true
+    else
+      failed=false
+    fi
+  fi
+  
+  if [ "$failed" == 'true' ]; then
+    error "Failed to update ${app}."
```
Botspot, line 306: This breaks the script because the `error ` function runs `exit 1`. In this case, when the first app fails to update, subsequent appd wouldn't be reinstalled at all, because the usage of `error()` would kill the script prematurely.

updater
```diff
@@ -44,9 +44,9 @@ check_update_interval() { #return 0 if update-interval allows update-checking to
   elif [ "$updateinterval" == 'Always' ];then
     return 0
   elif [ -z "$updateinterval" ];then
-    echo "Something isn't right. Does '${DIRECTORY}/data/settings/Check for updates' exist?" 1>&2
+    warning "Something isn't right. Does '${DIRECTORY}/data/settings/Check for updates' exist?" 1>&2
   else
-    echo "Warning: Unrecognized update interval!" 1>&2
+    error "Unrecognized update interval!" 1>&2
```
Botspot, line 49: This breaks the script because the `error ` function runs `exit 1`.

updater
```diff
@@ -217,14 +217,15 @@ update_now_gui() { #input: updatable_files and updatable_apps variables
     mkdir -p "$(dirname "${DIRECTORY}/${file}")"
     
     #copy new version
-    cp -f "${DIRECTORY}/update/pi-apps/${file}" "${DIRECTORY}/${file}" || echo -e "\e[91mFailed to copy ${DIRECTORY}/update/pi-apps/${file}\e[39m!"
+    cp -f "${DIRECTORY}/update/pi-apps/${file}" "${DIRECTORY}/${file}" || error "Failed to copy ${DIRECTORY}/update/pi-apps/${file}!"
```
Botspot, line 220: This breaks the script because the `error ` function runs `exit 1`.

updater
```diff
@@ -258,14 +259,14 @@ update_now_cli() { #input: updatable_files and updatable_apps variables
...
```
Botspot, line 262: This breaks the script because the `error ` function runs `exit 1`.

Botspot:
@cycool29 I've reviewed your PR and made comments as necessary.

updater
cycool29: > Not sure why you're placing the question into an array.

Just want to have `status` colors to the question.

manage
cycool29: There is a line above it mentioned `status "$app's install script has been updated. Reinstalling $app..."`.

updater
Botspot: > > Not sure why you're placing the question into an array.
> 
> Just want to have `status` colors to the question.

Oh I see. This may be fine for now, but down the road I may tweak the `status` function to be more than blue text. It may prefix everything with "`Status: `", or underline the text, or something else undesirable for asking a question. Son in this case, I recommend you replace it with `echo -en` (and then the codes for the blue color) Color codes can be found [here](https://misc.flogisoft.com/bash/tip_colors_and_formatting).

cycool29: Just found that can be done like this: `read -p "$(status 'Update now? [Y/n] ')" answer`.

cycool29:
> @cycool29 I've reviewed your PR and made comments as necessary.

Thanks! 
I have updated the scripts and fixed some bugs.

PR #1003, opened 2021-10-04
Remove `terminal-run` in Zoom script

cycool29:
Since there is `chmod +x "${HOME}/zoom/runzoom.sh"` in the installation script, Zoom can run without `terminal-run` script.


Botspot:
The terminal is there for a reason. Zoom doesn't close cleanly due to emulation issues, so the user is instructed to close the terminal to fully exit zoom.

PR #1005, opened 2021-10-04
Fix a minor syntax error in genapplist-yad.c

Jai-JAP:
Fix a syntax error causing 
```
/home/pi/pi-apps/etc/genapplist-yad.c:101:1: error: expected identifier or ‘(’ before ‘}’ token
 }
 ^
./pi-apps/preload: line 42: /home/pi/pi-apps/etc/genapplist-yad: No such file or directory
The genapplist-yad program failed to compile.
```
while running `preload` standalone
As logs can be seen only if script is run as standalone

### DIFF
--- a/etc/genapplist-yad.c
+++ b/etc/genapplist-yad.c
@@ -96,7 +96,6 @@ void get_app_status_with_none(char *directory, const char *app, char *status, si
 		snprintf(status, size, "%s", st);
 	} else {
 			snprintf(status, size, "none");
-		}
 	}
 }
 


Jai-JAP:
@Botspot @Itai-Nelken please check the above commit
tested working

Botspot:
I thought I had tested this before uploading; turns out I never deleted the existing binary so it was never compiled. (so it never got tested even though I thought it did)

Fortunately, the `preload` script is robust enough to detect compilation errors and works completely around them.

Botspot:
> Whoops! Thanks!

I think it was my fault. (removed the code that checked for an app's `uninstall` script)

PR #1009, opened 2021-10-04
Remove `apt` unwanted lines in `box86` uninstall script

cycool29:
Just remove `apt` unwanted lines after last line `sudo apt update`.


Itai-Nelken:
I think we should move the `reduceapt` function to the `api` script and use it instead.

Botspot:
And I think that running an apt update at all is unnecessary.

PR #1015, opened 2021-10-06
Add `git` function in api to remove target folder first before `git clone`

cycool29:
Completely solve `fatal: destination path already exists and is not an empty directory.` in pi-apps scripts.

### DIFF
--- a/api
+++ b/api
@@ -592,16 +592,28 @@ wget() { #this function intercepts all wget commands being used in app scripts. 
 }
 
 git() {  #remove target folder first before git clone
+  
   args=$*
+  
   if [[ $args == *"clone"* ]]; then 
     for arg in $args; do
-      if [[ $arg == *"://"* ]]; then #check if arg is a link 
-        link=$arg
+      if [[ $arg == *"://"* ]]; then #check if arg is a url
+        url=$arg
       fi
     done
-    rm -rf $(echo $( echo $link | awk -F "/" '{print $NF}' | sed 's/.git//g')) || sudo rm -rf $(echo $( echo $link | awk -F "/" '{print $NF}' | sed 's/.git//g')) #parse the link and get the folder name 
+    
+	[ -z "$url" ] && error "No repository URL specified! "
+	
+	folder=$(echo $( echo $url | awk -F "/" '{print $NF}' | sed 's/.git//g')) #parse the url and get the folder name
+	
+	[ -d "$folder" ] && status -n "Removing $folder directory... " && (rm -rf $(pwd)/$folder || sudo rm -rf $(pwd)/$folder) && status_green "Done"
+	
+  status -n "Cloning $url repository... "
+	local errors="$(command git $* &> /dev/null)" || error "\nFailed to clone $url repository.\nErrors: $errors"
+	status_green "Done"
+  else
+	  command git $*
   fi
-  command git $args
 }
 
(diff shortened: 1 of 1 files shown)


cycool29:
@Botspot

Botspot:
> @Botspot Any ideas?

I'm wondering if such a change would make sense when I am actively rewriting pi-apps with a `git_clone` function that does this and more.
See "the-big-update" branch: https://github.com/Botspot/pi-apps/tree/the-big-update

cycool29:
I would prefer `git clone` instead of `git_clone` as it won't need to change the other scripts and `git_clone` may make future app creators confuse. 

> See "the-big-update" branch: https://github.com/Botspot/pi-apps/tree/the-big-update

I can't find  `git_clone` function in the `api` of the branch...

cycool29:
@Botspot If you are working on the `git_clone` function, sorry to disrupt your plan.
However, I don't think `git_clone` is a good choice.

cycool29:
> Why not have a git function? Because I want app-creators to realize that they're using a function that does more than the plain git command.

There should not have many `pi-apps`-only function.
Why not just detect the `git clone` by checking the `$1` of `git` and if that is `clone`, remove destination folder and display progress, etc.

Just like `wget`, app creators don't know the `wget` is actually overrided to `aria2c`.

I have make a commit to show the status message and error message, which now **does the same thing with `git_clone`**.

Well, if you are stick on `git_clone`, I have tested the `git_clone` function, there is a bug which `dirname` command can't check the folder name ( tested on `git_clone https://github.com/Botspot/pi-apps` and it returns `https://github.com/Botspot/`, I just made some improvements here:
```
git_clone() { #silently clone a git repository but display the output if an error occurs
  # if no args provided, exit
  if [ -z $* ]; then
	error "No arguments provided!"
  fi
  
  # $1 is repo
  local IFS=' '
  for arg in "$@"; do
    if [[ "$arg" == *'://'* ]];then 
      url="$arg"
    fi
  done
  
  [ -z "$url" ] && error "git_clone(): no repository URL specified."
  
  local folder="$(echo $( echo $url | awk -F "/" '{print $NF}' | sed 's/.git//g'))" 
  
  status -n "Downloading $folder repository... "
  
  rm -rf "$folder" || sudo rm -rf "$folder" # use sudo if failed 
  local errors="$(git clone "$@" 2>&1)" || error "\nFailed to download $folder repository.\nErrors: $errors"
  
  status_green 'Done'
}
```

cycool29:
> 
> ```shell
> rm -rf ~/my-amazing-repo
> echo "Downloading my-amazing-repo..."
> git clone https://github.com/awesome-user/my-amazing-repo
> ```
> 
> They won't realize that the `git` function is doing both the `echo` and the `rm` for them. I think this would be a point of confusion going forward if the function was named `git`. _However_, if the function is named `git_clone`, app-creators will instinctively know that something is different. They will be less likely to write unnecessary lines of code as in my example above. I think that ` git_clone` function will make app-writing more intuitive, not less.

Hmm... that is really a issue.

Maybe we can just filter the `rm`s and `echo`s during app submission? Also, can we just remove the `rm`s and `echo`s in the current scripts?

However, from the consequences you proposed, I will prefer `git_clone` now. But this should not happen oftenly, as I said, **there should not have many `pi-apps`-only function.**

PR #1017, opened 2021-10-06
update username in scripts

ryanfortner:
I recently changed my username. While the repositories are still accessible from their old links for the time being, if someone was to sign up with the username `chunky-milk`, all the links will become invalid, so this is necessary.


ryanfortner:
Thanks. Sorry for any inconvenience

Issue #1020, opened 2021-10-06
Purge Installed removes packages which were set to manually installed before running pi-apps


theofficialgman:
how to replicate:
```sudo apt install cmake``` which sets cmake to manually installed

run a pi-apps script which does a pkg-install which overwrittes cmake to automatically installed instead of manually installed (VERY BAD)

run the uninstall script for a pi-apps script with purge-installed and cmake gets REMOVED even though it was manually installed before using pi-apps

this is very dangerous and could cause applications to not function, cmake was just the example used here

Botspot:
Great question. I've explained this change on the commit summary page.
https://github.com/Botspot/pi-apps/commit/a2d7ef55f92c0cea384abb6d1fa48acc43d79e51#commitcomment-57623470

PR #1021, opened 2021-10-06
only mark non-manually installed packages as autoremovable

theofficialgman:
fix for issue https://github.com/Botspot/pi-apps/issues/1020

the previous implementation could result in broken packages as a resulted of a purge-installed from pi-apps

### DIFF
--- a/pkg-install
+++ b/pkg-install
@@ -341,9 +341,21 @@ if echo "$PKG_LIST" | grep -q '*';then
 fi
 }
 
-#mark ALL packages as autoremovable
+#get manually installed packages and save to temporary .txt file
+apt-mark showmanual > /tmp/manual_installs.txt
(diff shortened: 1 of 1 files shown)


theofficialgman:
actually hold on don't merge this yet...
the current system that you have is broken anyway and misplaced

theofficialgman:
@Botspot is there even any reason for this to exist?
```
for package in $PKG_LIST ;do
  sudo -E apt-mark auto "$package" || error "pkg-install: failed to mark the $package package as autoremovable."
done
```
it seems like this causes all the problems and gives no benefit. why did you add it?

Botspot:
Great question. I've explained this change on the commit summary page.
https://github.com/Botspot/pi-apps/commit/a2d7ef55f92c0cea384abb6d1fa48acc43d79e51#commitcomment-57623470

PR #1023, opened 2021-10-07
Minecraft Java: use pkg-install for java 11

ryanfortner:


Botspot:
Merging this feels a bit silly when Lunar Client is replacing the MCJava app very soon. (maybe even today)

PR #1025, opened 2021-10-08
Improve `aria2c` output

cycool29:
Made some enhancement to `aria2c` output.

- Display downloading message instead of hanging the screen while downloading file 
- Check for `-q` and `-qo-` flags more accurately
- Use `--quiet` flag to suppress output
- Directly `exit 1` if download failed (this may break some scripts)


cycool29:
@Botspot Any ideas?

PR #1029, opened 2021-10-09
Add `yad-dark` `yad-light` and `yad-default`

cycool29:
Finally I found a way for `yad` dark theme.

Just set the `GTK_THEME` variable before starting the GUI.

- Added `yad-dark` app list style (Use `Adwaita-dark` theme)
- Added `yad-light` app list style (Use RPi default `PiXflat` theme)
- Replace original `yad` app list style to `system-default` (just do what the original `yad` does, it obey widget style set in `lxappearance`)

Currently only support on the main window, planning to make it as function in `api`. 
That will be a huge project to inherit this to ALL `yad` window in pi-apps.

### DIFF
--- a/gui
+++ b/gui
@@ -261,7 +261,7 @@ while true;do
     #now that list has been generated, kill the splash screen
     (sleep 1; kill "$splashpid" 2>/dev/null) &
 
-    if [ "$guimode" == system-default ] || [[ "$guimode" == yad* ]];then 
+    if [[ "$guimode" == yad* ]];then 
       
       if [ -z "$prefix" ];then
         buttons=("--button=!${DIRECTORY}/icons/search.png!Search:5" \


cycool29:
> > It may not actually be that bad, if the `api` script can `export` the variable to all other scripts.
> 
> Yes, it appears that adding this one line to `api` affects all yad windows: `export GTK_THEME=Adwaita-dark` Seems like one could add some simple logic to a few scripts and this would be an easy merge.

These lines seem works!
```
guimode="$(cat "${DIRECTORY}/data/settings/App List Style")"
[ -z "$guimode" ] && guimode=system-default  
export GTK_THEME=$guimode
```

cycool29:
Now all `yad` window with `api` sourced will follow the `GTK_THEME` variable !

PR #1031, opened 2021-10-10
Add `add_english` function to install english locale

cycool29:
Solve the `apt lock` not detecting error output correctly by changing terminal output locale temporarily.

Tested on vdesktop RPi OS (first use of vdesktop lol).

### DIFF
--- a/api
+++ b/api
@@ -46,9 +46,8 @@ ${blue}┃  ${red}⬛⬛⬛ ${blue}┃\e[97m            ▕  ▕
 #end of output functions
 
 add_english() {  # add en_US locale for more accurate error 
-  english_supported=$(cat /usr/share/i18n/SUPPORTED | grep -o 'en_US.UTF-8' )
-  if [ "$english_supported" == "en_US.UTF-8" ]; then 
-    locale=$(locale -a | grep -o 'en_US.utf8')
+  if [ "$(cat /usr/share/i18n/SUPPORTED | grep -o 'en_US.UTF-8' )" == "en_US.UTF-8" ]; then 
+    locale=$(locale -a | grep -oF 'en_US.utf8')
     if [ "$locale" != 'en_US.utf8' ]; then
       echo -n "Adding en_US locale for better logging... "
       sudo sed -i '/en_US.UTF-8/s/^#[ ]//g' /etc/locale.gen


Botspot:
Are you saying that you found a situation in which the C locale did ***NOT*** produce a predictable output?

cycool29:
> Are you saying that you found a situation in which the C locale did _**NOT**_ produce a predictable output?

Ya, I tested on a Spanish vdesktop and `C` locale seems not working.
After lots of googling I can't find what is actually `C` locale means and does.
So I think directly export `LANGUAGE` as `en_US` will be easier and more specific. 
Many error reports received with `apt_lock_wait` passed but still show message `Could not get lock...` and the problem probably because of the locale/language problem.

cycool29:
@Botspot Any ideas?

Botspot:
I've fixed two issues:

1. If for some reason `/usr/share/i18n/SUPPORTED` returns multiple matches, it would most likely return a bash error. This is fixed by placing the variable in quotation marks when it's being assigned. Also I went ahead and removed the unnecessary variable `english_supported` - only used once.
2. Grep had regex (regular expression) enabled for no reason. The period character (`.`) is a regex symbol for "*match any character*". 
    Here's a possible failure scenario:
    ```
    echo 'en_USaUTF-8' | grep -o 'en_US.UTF-8' #match found
    ```
    To fix, add the `-F` flag.
    ```
    echo 'en_USaUTF-8' | grep -oF 'en_US.UTF-8' #match not found
    ```

I've gone ahead and added both fixes, and this PR will be merged.

cycool29:
> I've fixed two issues:
> 
> 1. If for some reason `/usr/share/i18n/SUPPORTED` returns multiple matches, it would most likely return a bash error. This is fixed by placing the variable in quotation marks when it's being assigned. Also I went ahead and removed the unnecessary variable `english_supported` - only used once.
> 2. Grep had regex (regular expression) enabled for no reason. The period character (`.`) is a regex symbol for "_match any character_".
>    Here's a possible failure scenario:
>    ```
>    echo 'en_USaUTF-8' | grep -o 'en_US.UTF-8' #match found
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    To fix, add the `-F` flag.
>    ```
>    echo 'en_USaUTF-8' | grep -oF 'en_US.UTF-8' #match not found
>    ```
> 
> I've gone ahead and added both fixes, and this PR will be merged.

Thanks!

theofficialgman:
@cycool29 I am getting a very strange error from a user as a result of this change

.NET CLI sees the english locale being available and changes to using it automatically for some reason which is annoying.

I can't reproduce any issues with using the C locale and not generating english even when only using a system with the C locale and spanish (es_US) available.

can you please elaborate on your test setup as I can't replicate

PR #1037, opened 2021-10-11
Remove `yad` as it was installed when installation of pi-apps

cycool29:
Avoid accidentally uninstall `yad`.


Itai-Nelken:
As pkg-install only uninstalls packages it installed, this doesn't matter.
yad won't be installed as its already installed, so it won't be uninstalled.

PR #1054, opened 2021-10-15
change wget function to custom_get

theofficialgman:
overwriting a builtin function name is very dangerous since the get_custom function does NOT adhere to wget spec. Script writers should have to actively changes to this function if they want its features and not be forced to use it

explained here:
https://discord.com/channels/770629697909424159/770629697909424162/898429620662177812
https://discord.com/channels/770629697909424159/770629697909424162/898570566141898862

### DIFF
--- a/api
+++ b/api
@@ -1366,7 +1366,7 @@ git_clone() { #silently clone a git repository but display the output if an erro
   status 'Done'
 }
 
-wget() { #Intercept all wget commands. When possible, uses aria2c.
+get_custom() { #Intercept all wget commands. When possible, uses aria2c.
   local file=''
   local url=''
   local use=aria2c


Botspot:
> this still prints the logging but it works out fine because you changed the status/error/status_green to print to stderr. so I think its still good. just wanted to note anyway
> 
> ```
> wget -4O- https://gist.githubusercontent.com/theofficialgman/fe64667d3532837dd84b1a43741a6e55/raw/a
> 116ef0f322c0ff38f9032d2ce497dff226551e5/output.sh | bash
> ```

[Fixed.](https://github.com/Botspot/pi-apps/commit/20378324ce92ca1e7634db77adc747a18ab214b2)

PR #1065, opened 2021-10-20
Use `sudo` when adding flathub remote and remove unwanted lines

cycool29:
To avoid `error: Flatpak system operation ConfigureRemote not allowed for user`.
Also remove these unwanted lines:
```
Note that the directories 

'/var/lib/flatpak/exports/share'
'$HOME/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.
```


apps/Telegram/install
```diff
@@ -10,9 +10,26 @@ function error {
...
 
-sudo -E flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
+# Add flathub remote
+( sudo flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo || flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo || error "Flatpak failed to add flathub remote!"  ) | grep -v "Note that the directories 
+
+'/var/lib/flatpak/exports/share'
+'$HOME/.local/share/flatpak/exports/share'
+
+are not in the search path set by the XDG_DATA_DIRS environment variable, so
+applications installed by Flatpak may not appear on your desktop until the
```
Botspot, line 20: Did you mean for these lines to be comments?

cycool29: They are in the `grep -v`

Botspot: > They are in the `grep -v`

I guess I depend too much on syntax highlighting! (Also the `grep -v` part was invisible unless I scrolled to the left)

PR #1089, opened 2021-10-27
Minor improvements to `manage` and `api`, add latest commit ID section in `get_device_info`

cycool29:
- Remove `-c` flag in aria2_flags for `--allow-overwrite` to work
- Remove dupicate `format_logfile`
- Add latest commit ID section in `get_device_info` to check if user's pi-apps is even with repo's one
- cd to $HOME in api
- Show `Download url to path` if not download as `$(pwd)/$(basename $url)` in `wget`


manage
```diff
@@ -272,7 +271,7 @@ Or on Discord: \e[94m\e[4mhttps://discord.gg/RXSTvaUvuu\e[0m" | tee -a "$logfile
     fi
     
     format_logfile "$logfile" #remove escape sequences from logfile
-    mv "$logfile" "$(echo "$logfile" | sed 's+-incomplete-+-fail-+g')"
+    mv "$logfile" "$(echo "$logfile" | sed 's+-incomplete-+-fail-+g')" && logfile="$(echo "$logfile" | sed 's+-incomplete-+-fail-+g')"
```
Botspot, line 274: What effect does changing the value of the `$logfile` variable do? The `manage` script exits shortly afterwards.

cycool29: The logfile still needed by `log_diagnose`.

cycool29: image

Botspot: Looks like we're dealing with duplicate code. There's no need to run `log_diagnose` twice in `manage install`.

PR #1096, opened 2021-10-28
BUGFIX: doom3, Chiaki and Cool Retro Term install for bullseye and ubuntu 21.04+

theofficialgman:
also transition to install_packages and purge_packages while we are at it (to avoid unnecessary double update)

qt5-default no longer exists in these new debian based releases (so install the dependencies instead, it was just a meta package basically)

should fix: https://discord.com/channels/770629697909424159/874656141475463229/903012266708963411 https://discord.com/channels/770629697909424159/874656141475463229/903051515210661920
and many other instances of this error in the #error-logs

### DIFF
--- a/apps/Chiaki/install
+++ b/apps/Chiaki/install
@@ -8,7 +8,7 @@ function error {
 }
 
 # install dependencies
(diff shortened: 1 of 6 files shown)


theofficialgman:
@Botspot actually this needs further changes... apparently qt5-default is necessary for qmake to pick up on the version of qt (4 or 5) for prior to bullseye and 21.04.

so I guess ever script needs two install paths, one for buster (and prior to 21.04) and one for bullseye (and later than 21.04)

theofficialgman:
I have a fix though... making a PR now

PR #1099, opened 2021-10-29
install qt5-default if its available in apt repos

theofficialgman:
necessary on debian/ubuntu versions prior to bullseye/hirsute. Previously removed by a prior PR.

Basically this just checks if its in the repo, then install it, if its not, then don't. So dependencies stay clean for newer debian/ubuntu versions and old debian/ubuntu versions are still supported.

@Botspot necessary fix. prior PR merge caused errors for buster users https://discord.com/channels/770629697909424159/874656141475463229/903762557876174909

### DIFF
--- a/apps/Chiaki/install
+++ b/apps/Chiaki/install
@@ -10,6 +10,11 @@ function error {
 # install dependencies
 install_packages cmake python3-protobuf protobuf-compiler libavcodec-dev libopus-dev libsdl2-dev libssl-dev qtbase5-dev qtchooser qtmultimedia5-dev libqt5svg5-dev libqt5multimedia5-plugins libqt5opengl5-dev || error "Failed to install dependencies"
 
+# install qt5-default if its availabe in apt (necessary on debian/ubuntu versions prior to bullseye/hirsute)
+package_available qt5-default
+if [[ $? == "0" ]]; then
+  install_packages qt5-default || error "Failed to install dependencies"
+fi
 # install Chiaki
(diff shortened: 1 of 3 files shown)


Botspot:
Looks like a good use for the `package_available` function.
See: https://github.com/Botspot/pi-apps/blob/master/DOCUMENTATION.md#the-api-script

theofficialgman:
> Looks like a good use for the `package_available` function. See: https://github.com/Botspot/pi-apps/blob/master/DOCUMENTATION.md#the-api-script

wasn't aware of that
I will edit it and force push over

theofficialgman:
@Botspot updated to use package_available

Botspot:
Parenthesis prevent the `error` command from exiting the script.

theofficialgman:
> Parenthesis prevent the `error` command from exiting the script.

hmm... but I need the parenthesis in order for error to only run if nstall_packages has an error exit code... not if package_available has an exit code

theofficialgman:
I guess I could do this...
```
package_available qt5-default
if [[ $? == "0" ]]; then
install_packages qt5-default || error "Failed to install dependencies"
fi

**Comment by theofficialgman** [comment]
@Botspot this should be good to merge
I implemented the above

**Comment by Botspot** [comment]
> do you have any better ideas?
> 
> I just think its ugly but it works as intended

For now I think we'll go with this. If more scripts begin installing certain packages if they are available, I could write a combined function like this:
```bash
install_packages_if_available qt5-default || error "Failed to install dependencies"
```

PR #1100, opened 2021-10-30
New Pi-Apps category for plugins

Crilum:
I think it would be a good idea to make a new category for Pi-Apps plugins, such as `Pi-Apps Terminal Plugin (bash)` and `Pi-Apps Terminal Plugin (python)`.
But, there are only two Plugins right now.. So maybe we should wait until there are some more, so the category looks more full..


Crilum:
Sorry for all the commits.. I didn't realize that you have to make a new branch for a PR.. :man_facepalming:

Botspot:
> So maybe we should wait until there are some more, so the category looks more full..

Yes, as you can tell, the "Tools" category is for everything that does not fit into another category. Creating a new category with only two things in it does not make sense. In general, I think we should wait on creating any new category until it can contain 10 apps, on 32-bit at least.

PR #1103, opened 2021-11-01
update model info and add ram sizing

theofficialgman:


Botspot:
What's the point of the "Ram size" entry?

theofficialgman:
> * Will this detect Android?
> * What's the point of the "Ram size" entry?

I think its supposed to be able to detect android (when run on termux or similar). Someone needs to test that for me

ram size is there for assisting in debugging error logs. say something consistently fails to install on only 1GB ram systems
plus its just nice to know

Itai-Nelken:
This works fine on android using termux (the `getprop` if case is the one that works).

PR #1109, opened 2021-11-04
Update IntelliJ Idea to 2021.2.3, try Java 17

cobalt2727:
i'm kind of just assuming java 17 will cooperate, since that's the next LTS build (and some distros already have it in their repos)


cycool29:
`openjdk-17-jdk` is not yet in Raspberry Pi's repository, even the deb from packages.debian.org won't work.

theofficialgman:
yup I just told cobalt this. I don't see any reason that openjdk-17 should be installed at all if its not needed. even in your refactor cobalt

cobalt2727:
I was leaning towards 17 for long-term futureproofing, seeing as it's the next LTS release. I could add back the fallback to 11 in the initial commit

theofficialgman:
also cobalt, pkg-install is now the old way of installing packages
if you are going to update this and cause users to get an update showing on their devices, you should also update to using the  `install_packages` and `purge_packages` functions. you can looks at the multimc5 script if you need to know the syntax for that

theofficialgman:
@cobalt2727 what you did is not safe I don't think...  that will break the dummy deb and leave the package uninstallable
you also need to modify the uninstall script to use `purge_pacakges`

cobalt2727:
Noted, good call. I really shouldn't be doing this so close to midnight...

theofficialgman:
better to do this if you really want 17 if its available
```
install_packages openjdk-11-jdk || error "Failed to install dependencies"
package_available openjdk-17-jdk
if [[ $? == "0" ]]; then
  install_packages openjdk-17-jdk || error "Failed to install dependencies"
fi
```

and then you need to do this in the uninstall script
```
purge_packages || error "Dependencies failed to uninstall"
```

theofficialgman:
now squash and force push

theofficialgman:
Intellij IDEA updated as of this commit https://github.com/Botspot/pi-apps/commit/247de1d72295b9d3d68712ef24240e675f5a3dc4

still using openjdk-11-jdk since that supported on the most distro versions

PR #1122, opened 2021-11-07
zoom: remove cmake dependency

ryanfortner:
box86/64 isn't being compiled anymore due to apt repos being used, so cmake isn't necessary

### DIFF
--- a/apps/Zoom/install-32
+++ b/apps/Zoom/install-32
@@ -20,9 +20,9 @@ fi
 # Get dependencies
 DISTRO=$(lsb_release -is)
 if [[ "${DISTRO}" == "Debian" || "${DISTRO}" == "Raspbian" ]]; then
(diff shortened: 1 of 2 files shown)


Itai-Nelken:
Can someone test this? I know cmake isn't being used, but last time someone tried to remove it zoom didn't launch I think.

Crilum:
I can try it, but I already have `cmake` installed. Would I need to uninstall it?

cycool29:
> I can try it, but I already have `cmake` installed. Would I need to uninstall it?

That should not affect your system, just note down dependencies uninstalled when uninstalling cmake and install them back after testing.

Crilum:
It works for me. I do have cmake installed already, and if it makes a difference, I can try it again with cmake uninstalled..
2021-11-08-210845_2560x1440_scrot

Itai-Nelken:
Great, thanks for testing @Crilum

Crilum:
Your weclome! :+1:

PR #1123, opened 2021-11-08
Urgent - update electron version for Snapdrop

ryanfortner:
snapdrop will not work for any users until merged


Itai-Nelken:
Thanks!

PR #1129, opened 2021-11-09
Improvements on import-app

### LINKED ISSUE #1127: App editor fails to use install_packages and purge_packages
When I click the `Run install-32` or `Run install-64` button, if my install script uses `install_packages` or `purge_packages`, it fails with the error:
```
/home/pi/pi-apps/apps/Electron Fiddle/install-32: line 11: install_packages: command not found
Closing in 10 seconds.

```

If you just use the install/uninstall button, it works, but not with the App Editor..

cycool29:
Prompt no app imported in import-app, fix typo.


cycool29:
Removed `createapp` fix since it is already fixed in https://github.com/Botspot/pi-apps/commit/12ff3d61c214ddaed345f4496234a7c4f389c7ca.

etc/import-app
```diff
@@ -64,20 +64,21 @@ import_zip() { #given a zipfile, extract it to apps directory and return the nam
...
     <b>1068</b>
     <b>https://link/to/app.zip</b>
-    <b>$HOME/my-app.zip</b>" \
+    <b>$HOME/my-app.zip</b>\n" \
     --form --field='' '')" || exit 0
 fi
 
-#if nothing in blank, exit now.
-[ -z "$import" ] && exit 0
+[ -z "$import" ] && yad "${yadflags[@]}" --width=310 --title="App importer" --image="${DIRECTORY}/icons/logo-64.png" --text="\n   No app has been imported." --button=Close:0 --window-icon="${DIRECTORY}/icons/logo.png"
```
Botspot, line 80: Replacing an `exit 0` with a dialog will allow the script to continue.

Itai-Nelken: Making a function that runs the dialog and then `exit 0` or making it an `if` statement are solutions.

cycool29: Fixed.

cycool29:
@Botspot Actually the `import-app` tool is only for importing apps, or importing all the files?

Botspot:
> @Botspot Actually the `import-app` tool is only for importing apps, or importing all the files?

It's only meant for importing individual apps from another person's branch into your local Pi-Apps folder.

cycool29:
What do you think if we make comparison to user's local pi-apps folder ($HOME/pi-apps) instead of downloading from master? 
That will definitely improve the import app speed. @Botspot

Botspot:
> What do you think if we make comparison to user's local pi-apps folder ($HOME/pi-apps) instead of downloading from master? That will definitely improve the import app speed. @Botspot

Comparing a fresh download of pi-apps with the branch will work around the possibility that there are local changes to apps in your main Pi-Apps folder.

Consider this scenario: If I've made local changes to the Cura app, and if the branch is compared to my local ~/pi-apps folder, the Cura app will show up as an "importable" app. If I don't de-select Cura from the list, my local changes would actually be reverted!
Also, if my local pi-apps installation was out-of-date, other apps in the comparison branch may not match my local pi-apps folder, thereby displaying additional "importable" apps. Importing those apps would effectively update my pi-apps system, but without reinstalling anything.

By downloading a fresh copy of the pi-apps repo, we sidestep those two edge-cases.

PR #1139, opened 2021-11-11
Update XSnow to 3.3.2

cycool29:
Also change to use its website's download link instead of Sourceforge one.

~~I found compilation instructions from [its website](https://www.ratrabbit.nl/ratrabbit/xsnow/downloads/index.html), will try to compile it on arm64 later.~~ 
Currently not working, will try to email the owner of XSnow project for an Raspberry Pi arm64 build.


cycool29:
No luck, ends up with `./xsnow: /lib/aarch64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ./xsnow)`.

I think I can email the owner of the project for an arm64 build for Raspberry Pi.
(Actually there is already one from packages.debian.org, but it needs a newer version of libc6, which breaks the system.)

However, the update of armhf is ready to merge. I will create a new PR once I got the arm64 version to work.

Botspot:
Using `exit 1` is the recommended standard for `install_packages` and `purge_packages` functions. Why? Because those functions diagnose some errors already; having two conflicting error-messages can be confusing.

cycool29:
> Using `exit 1` is the recommended standard for `install_packages` and `purge_packages` functions. Why? Because those functions diagnose some errors already; having two conflicting error-messages can be confusing.

Fixed.

PR #1147, opened 2021-11-11
total rework of persepolis script

theofficialgman:
fixed discord error log: https://discord.com/channels/770629697909424159/770629878319022090/908454398835961948

### DIFF
--- a/apps/Persepolis Download Manager/install
+++ b/apps/Persepolis Download Manager/install
@@ -1,13 +1,33 @@
 #!/bin/bash
 
-cd $HOME
-wget -O adwaita-qt.deb http://mirror.datamossa.io/raspbian/pool/main/a/adwaita-qt/adwaita-qt_1.0-2_armhf.deb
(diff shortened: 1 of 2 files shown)


Botspot:
While we're at it, maybe it would be simpler to install `youtube-dl` from `pip3` like YouTubuddy does?
https://github.com/Botspot/youtubuddy/blob/main/gui#L42-L57

theofficialgman:
@Botspot 
I would have (and wanted to do that) but we can't
its a apt dependency of the `persepolis_3.2.0.2_all.deb`
```
 Depends: python3 (>= 3.5), libqt5svg5 (>= 5.5), python3-pyqt5 (>= 5.5), libnotify-bin (>= 0.7), python3-setproctitle (>= 1.1), aria2 (>= 1.18), python3-requests (>= 2.0), python3-psutil (>= 3.4), adwaita-qt (>= 0.5), youtube-dl (>= 2016.02.22), ffmpeg (>= 2.8), sound-theme-freedesktop (>= 0.8), python3-pyqt5.qtsvg (>= 5.5)
```
if I installed from pip, it would just go ahead and install the apt package anyway
the buster-backports repo has the same version as pip so its good

PR #1161, opened 2021-11-14
get the master date from github repo instead of system clock

theofficialgman:
uses system clock as a fallback incase the github repo time is invalid or missing


Botspot:
Very important for the `manage` script: all debug or status information must to to stderr. Failure to do so will mess up anything that parses `manage`'s output.

Spoken by someone who has made this same oversight more than once...

theofficialgman:
ah ok I wasn't aware of that, thank you. I think your commit from last night fixes that already

Itai-Nelken:
Maybe making a custom echo function in the manage script that prints to stderr is a good idea if it's so important.

PR #669, opened 2021-05-18
Add support for alternate package names

Jai-JAP:
Add support for alternate package names if primary package name is not available in repos.
 
 Box86 install-64 currently looks like
 
 ```
 #install box86 dependencies
 CODENAME="$(lsb_release -cs)"
 if [[ "${CODENAME}" == "focal" || "${CODENAME}" == "buster" ]]; then
 "${DIRECTORY}/pkg-install" "libraspberrypi0:armhf libssh-gcrypt-4:armhf libgssapi-krb5-2:armhf libkrb5-3:armhf libssl1.1:armhf libcups2:armhf libsdl1.2debian:armhf libopusfile0:armhf libc6:armhf libx11-6:armhf libgdk-pixbuf2.0-0:armhf libgtk2.0-0:armhf libstdc++6:armhf libsdl2-2.0-0:armhf mesa-va-drivers:armhf libsdl1.2-dev:armhf libsdl-mixer1.2:armhf libpng16-16:armhf libcal3d12v5:armhf libsdl2-net-2.0-0:armhf libopenal1:armhf libsdl2-image-2.0-0:armhf libvorbis-dev:armhf libcurl4:armhf osspd:armhf pulseaudio:armhf libjpeg62:armhf libudev1:armhf libgl1-mesa-dev:armhf libsnappy1v5:armhf libx11-dev:armhf libsmpeg0:armhf libboost-filesystem1.67.0:armhf libboost-program-options1.67.0:armhf libavcodec58:armhf libavformat58:armhf libswscale5:armhf libmyguiengine3debian1v5:armhf libboost-iostreams1.67.0:armhf  libsdl2-mixer-2.0-0:armhf" "$(dirname "$0")" || exit 1
 elif [[ "${CODENAME}" == "groovy" || "${CODENAME} == "hirsute" ]]; then
[...]

PR #1162, opened 2021-11-14
fix critical Zoom 64-bit installation error

cobalt2727:
also modified dependencies to be 1:1 with the 32-bit install script

### DIFF
--- a/apps/Zoom/install-64
+++ b/apps/Zoom/install-64
@@ -10,7 +10,7 @@ if ! command -v box64 >/dev/null;then
 fi
 
 # Get dependencies
-install_packages libxcb-xtest0 libxcb-xfixes0 libturbojpeg0 pulseaudio-utils pulseaudio | install_packages libturbojpeg libxcb-xtest0 libxcb-xfixes0 pulseaudio-utils pulseaudio || exit 1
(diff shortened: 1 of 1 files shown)


cobalt2727:
tester report from github:

```ruby
(Reading database ... 95%
(Reading database ... 100%
(Reading database ... 93281 files and directories currently installed.)

Apt finished.
Package installation complete.
E: Unsupported file /home/pi/pi-apps-c08187a2.deb given on commandline
Apt finished.
Failed to install the packages!
APT reported these errors:
E: Unsupported file /home/pi/pi-apps-c08187a2.deb given on commandline
E: Unsupported file /home/pi/pi-apps-c08187a2.deb given on commandline
```

Botspot:
Looks like an issue with the `|` operator - bash is interpreting it as a pipe.

cobalt2727:
shouldn't we still have it attempting to install `libxcb` and `pulseaudio` dependencies for both tries though?

Botspot:
> shouldn't we still have it attempting to install `libxcb` and `pulseaudio` dependencies for **both tries** though?

That's not what is happening here. The `install_packages` command is only being run once. However, in this case we have a "`|`" operator between two package names: `libturbojpeg0` and `libturbojpeg`. These are two different packages - some systems have the former while others have the latter.
This `|` functionality was added in PR #669.

PR #1167, opened 2021-11-15
min: update to v1.22.1

ryanfortner:


apps/Min/install-32
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-install_packages https://github.com/ryanfortner/minbrowser-arm-builds/releases/download/v1.22/min_1.22.0_armhf.deb || exit 1
+install_packages https://github.com/ryanfortner/minbrowser-arm-builds/releases/download/1.22.1/min_1.22.1_armhf.deb || exit 1
```
Itai-Nelken, line 3: Is the `v` removed intentionally?

ryanfortner: Yes.

apps/Min/install-64
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-install_packages https://github.com/minbrowser/min/releases/download/v1.22.0/min_1.22.0_arm64.deb || exit 1
+install_packages https://github.com/minbrowser/min/releases/download/v1.22.1/min_1.22.1_arm64.deb || exit 1
```
Itai-Nelken, line 3: Can youbremove the `v` here as well to be consistent please?

Itai-Nelken: oh, I didn't see, never mind.

Itai-Nelken:
Thanks!

PR #1178, opened 2021-11-17
update multimc5 version and add java 17

theofficialgman:
(necessary for minecraft 1.18 prerelease 2 and later)
@Itai-Nelken @Botspot 
current version works fine, but most users won't have java 17 already installed.
as is the norm, we can't use raspbian repos (since they don't have java 17) and buster is also missing java 17 in general

so all debian versions get tar.gz from temurin

### DIFF
--- a/apps/Minecraft Java MultiMC5/install
+++ b/apps/Minecraft Java MultiMC5/install
@@ -74,6 +74,7 @@ case "$__os_id" in
                 hash -r
                 mkdir -p ~/MultiMC/install/java || exit 1
                 cd ~/MultiMC/install/java || exit 1
+                # download java 8 (unavailable in some debian versions)
                 rm -rf jdk8u302-b08
                 rm -rf java-8-temurin-aarch64
                 rm -rf OpenJDK8U-jdk_aarch64_linux_hotspot_8u302b08.tar.gz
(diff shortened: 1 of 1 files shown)


apps/Minecraft Java MultiMC5/install
```diff
@@ -129,7 +152,7 @@ case "$__os_id" in
                 ;;
         esac
         # install dependencies
-        install_packages build-essential libopenal1 x11-xserver-utils git clang cmake curl zlib1g-dev openjdk-8-jre openjdk-11-jdk openjdk-16-jre qtbase5-dev || error "Failed to install dependencies"
+        install_packages build-essential libopenal1 x11-xserver-utils git clang cmake curl zlib1g-dev openjdk-8-jre openjdk-17-jre openjdk-11-jdk openjdk-16-jre qtbase5-dev || error "Failed to install dependencies"
```
Botspot, line 155: Are we sure that the openjdk-17-jre package will be available on Buster?

theofficialgman: its not, make sure to read what that is under
thats the ubuntu only case
buster gets a tar.gz

theofficialgman:
@Botspot should be good as is, see my response

PR #1181, opened 2021-11-17
Improve comparing fork and upstream, added `Imported` category and fix bugs

cycool29:
- Use `-b` to specify a branch when git clone instead of check out branch after clone
- Pull upstream repo before checking diff to avoid outdated app in fork listed as a modified app
- Add imported app to `Imported` category
- Remove category if no app in it 
- Added `Imported` category icon
- Check Internet connection
- Fix PR link not detected


cycool29:
@Botspot

api
```diff
@@ -557,6 +557,19 @@ list_subtract() { #Outputs a list of apps from stdin, minus the ones that appear
 }
 
 read_category_files() { #Generates a combined categories-list from several sources: category-overrides, global categories file, and unlisted apps. Format: "app|category"
+  
+  #remove app category if app folder not found
+  IFS=$'\n'
+  for app in $(cat "${DIRECTORY}/etc/categories" "${DIRECTORY}/data/category-overrides"); do 
+  	if ! [ -d "${DIRECTORY}/apps/$(echo "$app" | sed 's/|.*//')" ] &>/dev/null; then 
+  		sed -i "/$app/d"  "${DIRECTORY}/data/category-overrides"
+		sed -i "/$app/d"  "${DIRECTORY}/etc/categories"
```
Botspot, line 566: The global `categories` file should never be modified.

etc/import-app
```diff
@@ -185,26 +205,31 @@ elif [[ "$import" =~ ^[0-9]+$ ]] || [[ "$import" =~ *'://'*'/pull/'[0-9]+$ ]];th
...
 
-echo -e "\n$(basename "$0") just imported these apps:"
-echo "$imported_apps" | sed 's/^/- /g'
 
 #display on-screen which apps were updated
 IFS=$'\n'
 for app in $imported_apps ;do
   LIST="$LIST
-$([ -f "${DIRECTORY}/apps/$app/icon-64.png" ] && echo "${DIRECTORY}/apps/$app/icon-64.png" || echo "${DIRECTORY}/icons/none-64.png")
+$([ -f "${DIRECTORY}/apps/$app/icon-24.png" ] && echo "${DIRECTORY}/apps/$app/icon-24.png" || echo "${DIRECTORY}/icons/none.png")
```
Botspot, line 225: Is there a reason we are displaying imported apps with 24x24 icons rather than 64x64 icons?

api
cycool29: Fixed.

etc/import-app
cycool29: I think 64x64 is too big for the app name text size beside it, so I think 24x24 will look nicer.

Botspot: Okay.

api
```diff
@@ -557,6 +557,18 @@ list_subtract() { #Outputs a list of apps from stdin, minus the ones that appear
 }
 
 read_category_files() { #Generates a combined categories-list from several sources: category-overrides, global categories file, and unlisted apps. Format: "app|category"
+  
+  #remove app category if app folder not found
+  IFS=$'\n'
+  for app in $(cat "${DIRECTORY}/data/category-overrides"); do 
+  	if ! [ -d "${DIRECTORY}/apps/$(echo "$app" | sed 's/|.*//')" ] &>/dev/null; then
```
Botspot, line 564: Can we switch out these tab characters for double-spaces to be consistent with the rest of the scripts?

cycool29: Fixed.

PR #1196, opened 2021-11-20
Issue forms

Itai-Nelken:


cycool29:
@Botspot I think we should add `legal and not piracy` checkbox in the app request template, adding these lines at the last of `app-request.yml` will done that:
```
        - label: I have confirmed that this app is legal and not piracy.
          required: true
```

PR #1198, opened 2021-11-21
Update Bleachbit to 4.4.2, use official deb, use sudo when removing/`sed`ing file, don't remove `BleachBit as Administrator` menu shortcut

cycool29:


apps/BleachBit/install
```diff
@@ -1,9 +1,6 @@
 #!/bin/bash
 
-install_packages https://osdn.mirror.constant.com/storage/g/r/ra/raspbian-addons/debian/pool/bleachbit_4.4.0-0_all.deb || exit 1
+install_packages https://download.bleachbit.org/bleachbit_4.4.2-0_all_debian11.deb || exit 1
```
Botspot, line 3: Do you know if this breaks Buster-compatibility? (That's not a problem, but if so, we may consider adding a "User error:" if that's the case)

cycool29: I tested this on my Raspbian buster VM (in Windows) and I don't see any problem.
I will test again later in vdesktop.

cycool29: Yes it works!

PR #1202, opened 2021-11-22
Let user choose apps/files to update in updater cli

cycool29:
Let user choose files or apps to update, like in GUI.


cycool29:
> I think this is a good idea, but I would still like the no-input version to be preserved. I use it for my [`update`](https://github.com/Crilum/update) script! 😄

Well the no-input version is still in `cli-yes`, which you used in your script.

Botspot:
I did a quick test and it failed to work as expected.

## Setup:
- Modify two apps to make them updatable.
- Modify two files to make them updatable.

I first did Ctrl+A for each list and it functioned as expected.
But the second time I tested it, I purposefully did not select any items from either list.
And yet, it still updated one app and one file.

cycool29:
> But the second time I tested it, I purposefully did not select any items from either list.
And yet, it still updated one app and one file.

Ya... That is really a bug. It will automatically select the one where the pointer is when ENTER, so I made a shortcut to let user use ESC to skip if they don't want to update anything.

cycool29:
I found a **pure** bash script, works what we want [here](https://unix.stackexchange.com/a/673436).

Will try to add it in `api`.

cycool29:
I just tested and it worked with Botspot's setup, but now it doesn't have the Select All function, since it already selects all options by default.

Instead of adding it in `api`, I think it would be better in `updater`, like `list_updates_gui`.

cycool29:
@Botspot Any thoughts?

theofficialgman:
how about instead of this, you use `dialog` for a CLI based gui for selecting apps in the updater cli
this can be used to make a fully cli only pi-apps gui as well? its format for input is very similar to zenity/yad

you can see it used over at the megascript in our userinput_func https://github.com/cobalt2727/L4T-Megascript/blob/628cdff14d6e2cb7d5975bbc4a73b5371d659677/functions.sh#L84

cycool29:
> how about instead of this, you use dialog for a CLI based gui for selecting apps in the updater cli?

We have already a very nice CLI select menu. I don't think we need to change to use TUI tools like `dialog` and `whiptail`.


> this can be used to make a fully cli only pi-apps gui as well. its format for input is very similar to zenity/yad


Yes, maybe we can add that in `api` for CLI pi-apps interface in the future.

But for now, we just need to let users to *select* apps to update. That would look ugly if there is only one app/file to update, and we occupy the whole terminal space (in my memory, both `dialog` and `whiptail` will always fullscreen) to show one option for user.

cycool29:
@Botspot

cycool29:
@Botspot This PR is ready to merge.

updater
```diff
@@ -229,6 +228,100 @@ file:$file"
...
+    local -n defaults=$3 &>/dev/null
+
+    local selected=()
+    for ((i=0; i<${#options[@]}; i++)); do
+      selected+=("true")
+      printf "\n"
+    done
+
+    # determine current screen position for overwriting the options
+    local lastrow=`get_cursor_row`
```
Itai-Nelken, line 252:
```suggestion
    local lastrow=$(get_cursor_row)
```
This is the new and recommended way to do that.

updater
```diff
@@ -229,6 +228,100 @@ file:$file"
...
+            ((idx++))
+        done
+    }
+
+    local active=0
+    while true; do
+        print_options $active
+
+        # user key control
+        case `key_input` in
```
Itai-Nelken, line 306:
```suggestion
        case $(key_input) in
```

Itai-Nelken (approved):
Looks good to me @Botspot.

PR #1207, opened 2021-11-22
Update tip for uploading files in app-request issue form

Itai-Nelken:
This isn't the best way to put it. but what I mean is that you can click on the text box to expand the regular github one and then you can upload files (not tested, but I see no reason for it not to work).

### DIFF
--- a/.github/ISSUE_TEMPLATE/app-request.yml
+++ b/.github/ISSUE_TEMPLATE/app-request.yml
@@ -32,7 +32,7 @@ body:
     attributes:
       label: Upload file
       description: "If you have a deb of the app, or if you've created a zip file for this app, or if you have any other relevant files, upload it here."
-      placeholder: "Tips: You can upload files to transfer.sh and paste the link here"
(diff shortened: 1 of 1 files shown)


Itai-Nelken:
I also added a checkbox for making sure the app isn't piracy and is legal (@cycool29).

cycool29:
I prefer transfer.sh because GitHub only supports `GIF, JPEG, JPG, MOV, MP4, PNG, CSV, DOCX, FODG, FODP, FODS, FODT, GZ, LOG, MD, ODF, ODG, ODP, ODS, ODT, PDF, PPTX, TXT, XLS, XLSX or ZIP`, some common used file types such as `XZ` and `DEB` is not included.



Maybe we should write :
```
Tips: You can upload files by here by clicking to expand the regular text-box.
      If the file type isn't supported, you can use transfer.sh and paste the link here.
```

Itai-Nelken:
What do you prefer @Botspot?

Botspot:
It's not optimal because Github refuses some filetypes, but this is an improvement so I'll merge it.

PR #1209, opened 2021-11-23
Update Drawing to 0.8.4

cycool29:
Fix error report from Discord.

### DIFF
--- a/apps/Drawing/install-32
+++ b/apps/Drawing/install-32
@@ -2,5 +2,5 @@
 
 #PACKAGES FOUND IN: https://launchpad.net/~cartes/+archive/ubuntu/ppa/+packages
 
-install_packages https://launchpad.net/~cartes/+archive/ubuntu/drawing/+files/drawing_0.8.3-0ubuntu1~hirsute_armhf.deb || exit 1
+install_packages https://launchpad.net/~cartes/+archive/ubuntu/drawing/+files/drawing_0.8.4-0ubuntu1~hirsute_armhf.deb || exit 1
 
--- a/apps/Drawing/install-32
+++ b/apps/Drawing/install-32
@@ -2,5 +2,5 @@
 
 #PACKAGES FOUND IN: https://launchpad.net/~cartes/+archive/ubuntu/ppa/+packages
 
-install_packages https://launchpad.net/~cartes/+archive/ubuntu/drawing/+files/drawing_0.8.4-0ubuntu1~hirsute_armhf.deb || exit 1
+install_packages https://launchpad.net/~cartes/+archive/ubuntu/drawing/+files/drawing_0.8.4-0ubuntu1~focal_armhf.deb || exit 1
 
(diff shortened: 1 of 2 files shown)


theofficialgman:
this is really prone to breakage (like every time this ppa updates)... don't you think it would be better to add the ppa?
might I suggest also to use the focal (or bionic) version to avoid any future linker errors

```
source /etc/os-release
echo "deb http://ppa.launchpad.net/cartes/drawing/ubuntu focal main" | sudo tee /etc/apt/sources.list.d/cartes-ubuntu-drawing-$VERSION_CODENAME.list
sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys BA15C6600C3B7BB85D2A7A46A7D69E90DC319136
install_packages drawing
```

cycool29:
> this is really prone to breakage (like every time this ppa updates)... don't you think it would be better to add the ppa?

I don't think it is 'better', this potentially breaks how pi-apps update. 
Pi Apps uses a very simple update logic: If the script from GitHub is modified, then reinstall the app. 
The app is out of pi-apps's control if we use PPA as pi-apps can't detect if the app is outdated. 

> might I suggest also to use the focal (or bionic) version to avoid any future linker errors (since hirsute is newer than buster)

I don't know about Ubuntu, but hirsute works well on rpi os in previous release. I will try out the focal one, if it works well I will change the deb to focal.

cycool29:
Just tested the focal deb and I don't see any significant problem, will change to that now.

theofficialgman:
> > this is really prone to breakage (like every time this ppa updates)... don't you think it would be better to add the ppa?
> 
> I don't think it is 'better', this potentially breaks how pi-apps update. Pi Apps uses a very simple update logic: If the script from GitHub is modified, then reinstall the app. The app is out of pi-apps's control if we use PPA as pi-apps can't detect if the app is outdated.

yeah but there is no reason to have pi-apps handle updating this package at all if you use the ppa
just let normal apt update/upgrade/update buddy/OS updater handle it

theofficialgman:
keeping it the way it is now makes you reliant on an unstable name whenever the package gets updated unnecessarily (as you have just experienced)

theofficialgman:
there are other apps in pi-apps that already do what I'm suggesting, such as installing mono in mission planner: https://github.com/Botspot/pi-apps/blob/9ac7ed8e50fa7728904150f7d1dd4bbf5f10b28b/apps/Mission%20Planner/install

cycool29:
> yeah but there is no reason to have pi-apps handle updating this package at all if you use the ppa
just let normal apt update/upgrade/update buddy/OS updater handle it

That is quite messy. This will get two app store involved: pi-apps to install the app, and apt to update the app. 
It is OK (like how Mission Planner do), but I simply don't like that.

*If it already works, don't fix it! :)* -- **Botspot**

theofficialgman:
> That is quite messy. This will get two app store involved: pi-apps to install the app, and apt to update the app. It is OK (like how Mission Planner do), but I simply don't like that.

remember... apt is the ONLY app store here. its what installs everything and keeps everything updated already. pi-apps would not work without apt (clearly)
manually watching out for a file change and updating pi-apps accordingly to avoid adding another apt repo is not a great idea as it is right now, imho

PR #1213, opened 2021-11-25
Fix bleachbit desktop shortcut path

mobilegmYT:
Installed bleachbit, and apprantly in newest deb they renamed the desktop shortcut. Also the non-root shortcut was removed apprantly.


Botspot:
Thanks!

PR #1234, opened 2021-12-02
import-app: Add app name to `imported_apps` when importing folder and fix bugs

cycool29:
To avoid showing `No app has been imported` when importing folders.

### DIFF
--- a/etc/import-app
+++ b/etc/import-app
@@ -103,7 +103,7 @@ if [[ "$import" == /* ]] ;then
     
   elif [ -d "$import" ];then
     #given a folder, copy it to the apps directory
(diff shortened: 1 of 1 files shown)


Botspot:
The LIST variable is generated on line 225 in all scenarios, so this fix should set the `imported_apps` variable rather than directly setting the LIST variable.

cycool29:
> The LIST variable is generated on line 225 in all scenarios, so this fix should set the `imported_apps` variable rather than directly setting the LIST variable.

Fixed.

Botspot:
Let's add imported apps to the "Imported" category once, at the bottom to avoid duplicate code.

PR #1244, opened 2021-12-06
Add whole `data` folder to .gitignore

cycool29:


Itai-Nelken:
I'm not sure that works, maybe `data/*`

cycool29:
`data` means git will ignore the `data` folder itself and all the files/subfolders under it, so when you delete the `data` folder, git will ignore it.

`data/*` will ignore only the files/subfolders under the `data` folder, **but not `data`**, so when you delete the `data` folder, git will alert you.

At here that doesn't bring much difference, since we just want to ignore the whole `data` folder, so `data` should be better.

cycool29:
> I'm not sure that works, maybe `data/*`?

It should be `data/**` in gitignore syntax. But `data` already did the job.

PR #1253, opened 2021-12-08
Update Blockbench link

cycool29:


Itai-Nelken:
Thanks! Could you check other apps that download from raspbian-addons?

cycool29:
> Thanks! Could you check other apps that download from raspbian-addons please?

Only Blockbench and Freetube are downloaded from raspbian-addons in pi-apps, and both of them were updated to the new link.

Botspot:
> Thanks! Could you check other apps that download from raspbian-addons please?

Fortunately it's really easy to do this. You can go to Pi-Apps -> Search button -> only search scripts, and then enter a search term like `raspbian-addons`.

Itai-Nelken:
> > Thanks! Could you check other apps that download from raspbian-addons please?
> 
> Fortunately it's really easy to do this. You can go to Pi-Apps -> Search button -> only search scripts, and then enter a search term like `raspbian-addons`.

I didn't know that!

PR #1254, opened 2021-12-08
automatic apps list wiki page

theofficialgman:
with pi-apps growing to a large number of apps. I figured it would be helpful for users if the current appslist was available on the wiki

this scripting is still in the works and there probably will be changes before I open it up for merging

dropping these links here for my future reference: 
https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

### DIFF
--- a/.github/workflows/update-wiki.sh
+++ b/.github/workflows/update-wiki.sh
@@ -1,0 +1,48 @@
+conversion() {
+
+    readarray -t categories < <(cat $GITHUB_WORKSPACE/etc/categories | sed 's/$/|/g' | awk -F'|' '{print $2}' | sort - | uniq)
+    readarray -t script_name < <(cat $GITHUB_WORKSPACE/etc/categories | sed 's/$/|/g' | awk -F'|' '{print $1}')
+    readarray -t script_category < <(cat $GITHUB_WORKSPACE/etc/categories | sed 's/$/|/g' | awk -F'|' '{print $2}')
+
+    for category in "${categories[@]}"; do
+        if [[ "$category" == "hidden" ]]; then
+            continue
+        fi
+        echo "# $category"
+        iter=0
+        for script in "${script_name[@]}"; do
+            if [[ "${script_category[$iter]}" == "$category" ]] && [[ -a "$GITHUB_WORKSPACE/apps/$script" ]]; then
+
+                # determine if app is arm64, arm32, or both
+                if [[ -a "$GITHUB_WORKSPACE/apps/$script/install" ]]; then
+                    arch="ARM32/ARM64"
+                elif [[ -a "$GITHUB_WORKSPACE/apps/$script/install-32" ]]; then
+                    if [[ -a "$GITHUB_WORKSPACE/apps/$script/install-64" ]]; then
+                        arch="ARM32/ARM64"
+                    else
+                        arch="ARM32 ONLY"
+                    fi
+                elif [[ -a "$GITHUB_WORKSPACE/apps/$script/install-64" ]]; then
+                    arch="ARM64 ONLY"
+                else
+                    arch="Package app"
+                fi
+                script_website=$(cat "$GITHUB_WORKSPACE/apps/$script/website" 2>/dev/null)
+                script_credits=$(cat "$GITHUB_WORKSPACE/apps/$script/credits" 2>/dev/null)
+                echo ""
+                script_url=$(echo $script | sed -e 's/ /%20/g')
+                echo "### <img src=https://github.com/Botspot/pi-apps/blob/master/apps/$script_url/icon-64.png height=32> ***[$script](https://github.com/Botspot/pi-apps/tree/master/apps/$script_url)***"
+                [[ ! -z "$script_website" ]] && [[ ! -z "$script_credits" ]] && echo "$script_website - $script_credits<br />"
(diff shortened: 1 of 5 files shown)


theofficialgman:
there is a live version of the wiki at my fork. changes committed in this PR will appear there
https://github.com/theofficialgman/pi-apps/wiki/Apps-List

Botspot:
Great idea!

Could the `etc/update-wiki.sh` script be moved elsewhere, to a place which signifies that it's not used in the pi-apps application? Having useless scripts alongside useful ones does not seem optimal.

Also, have you considered moving the app icons to the left so they appear first, in a straight vertical column?

theofficialgman:
@Botspot I've implemented your suggestions and a few others. Take another look

theofficialgman:
going to document this now so I don't forget. In order for this to work in pi-apps, botspot you will need to do a couple of things.

Go to your github account settings: 
image
and click generate token, give it a name, give to repo access, and then make it so it never expires, then copy the resulting token.

go to the settings in the pi-apps repo, under secrets, make a new secret, name it as seen in the pic below, and paste the token into it
image

this allows the github actions workflow to push to the wiki

Botspot:
Would it make sense to indicate whether an app is a package-app or not?

theofficialgman:
I wondered that  myself... we can do so easily as all package-app fall into this else:
```
                else
                    arch="ARM32/ARM64"
                fi
 ```
 right now I just report them as arm32/64

**Comment by Botspot** [comment]
For package-apps, I suggest replacing "ARM32/ARM64" with "Package app". After all, package-apps are automatically compatible with any CPU architecture, not just arm.

**Comment by Botspot** [comment]
I believe I've gotten the token and environment secret up and running. Are we ready to merge?

File: .github/workflows/update-wiki.sh [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+                elif [[ -a "$GITHUB_WORKSPACE/apps/$script/install-64" ]]; then
+                    arch="ARM64 ONLY"
+                else
+                    arch="Package app"
+                fi
+                script_website=$(cat "$GITHUB_WORKSPACE/apps/$script/website" 2>/dev/null)
+                script_credits=$(cat "$GITHUB_WORKSPACE/apps/$script/credits" 2>/dev/null)
+                echo ""
+                script_url=$(echo $script | sed -e 's/ /%20/g')
+                echo "### ![This is an image](https://github.com/Botspot/pi-apps/blob/master/apps/$script_url/icon-24.png) ***[$script](https://github.com/Botspot/pi-apps/tree/master/apps/$script_url)***"
```
Itai-Nelken, line 34: Maybe use the larger icon? The smallrr one looks a bit bad in my opinion.

theofficialgman: image
I feel like the larger one is just too big... let me see if it can be scaled or something

theofficialgman: ###  ***[Color Emoji font]

yup I can scale it down and use the larger one

theofficialgman: better now? before it was the 24x24 icon, now its the 64x64 icon scaled down by half to 32x32

theofficialgman:
> I believe I've gotten the token and environment secret up and running. Are we ready to merge?

I'll change the name back to Botspot in the yaml now and make it open for merging.

Botspot:
How about formatting the app's description as a quote instead of code?

theofficialgman:
> How about formatting the app's description as a quote instead of code?

quotes are a little bit harder to format automatically, and I honestly liked the code format better

Botspot:
> quotes are a little bit harder to format automatically

`sed 's/^/> /g'`

> and I honestly liked the code format better

Fair enough. You can leave it as-is.

theofficialgman:
@Botspot alright ready for merging

theofficialgman:
you might need to manually click the button under github actions to run it the first time

Botspot:
`actions/checkout@v2 and docker://decathlon/wiki-page-creator-action:latest are not allowed to be used in Botspot/pi-apps. Actions in this workflow must be: within a repository owned by Botspot.`
This is after enabling all Actions in the repo's settings.

theofficialgman:
oh you don't have your github email hidden `Botspot@users.noreply.github.com` I assumed it was this because my public facing one follows that

yours is an outlook email

.github/workflows/update-wiki.sh
Itai-Nelken: Much Better!

theofficialgman:
image

if you don't care for yours to be public (as it is now) then edit your email into the .yaml

theofficialgman:
at least I thought that was the issue... it looks like your github email has been changed to private... you just keep making commits with your other email

theofficialgman:
maybe it just takes time to change because I think you did it right if you enabled all actions

@Botspot just try it again

theofficialgman:
@Botspot yeah you must have not hit the save button. because if my actions permissions is set to "allow local actions only" it does not work, but if its set to "allow all actions" it does work

PR #1261, opened 2021-12-10
Add `get_codename` function

cycool29:
Usage in scripts:

```bash
if [ "$(get_codename)" == "bullseye" ]; then
    #install for bullseye
else 
    #install for others 
fi
```

### DIFF
--- a/api
+++ b/api
@@ -1563,6 +1563,10 @@ files_to_functions() { #turn every file in the function-files directory back int
   done
 }
 
+get_codename() { #get debian/ubuntu codename
+	grep Provides <(sudo apt info tzdata 2>&1) | sed 's/Provides: tzdata-//g'
+}
+
 #command interceptors - functions that enhance a command
 git_clone() { #silently clone a git repository but display the output if an error occurs
   # $1 is repo


Botspot:
> `grep Provides <(sudo apt info tzdata 2>&1) | sed 's/Provides: tzdata-//g'`

This is a weird way to do this. Not only does it require sudo, but it also takes a couple of seconds. Does this unorthodox approach have any advantages over simpler, faster methods?

ryanfortner (commented):
echo $(lsb_release -sc) works for me...

cycool29:
> > `grep Provides <(sudo apt info tzdata 2>&1) | sed 's/Provides: tzdata-//g'`
> 
> This is a weird way to do this. Not only does it require sudo, but it also takes a couple of seconds. Does this unorthodox approach have any advantages over simpler, faster methods?

I found `grep Provides <(apt-cache show tzdata 2>&1) | sed 's/Provides: tzdata-//g'` works for me. It doesn't require sudo, and it is a bit faster.

> `lsb_release -sc` works for me...

It is editable by changing the `VERSION_CODENAME` section in `/etc/os-release`. 

Actually the APT one is also editable, in `/var/lib/apt/lists/raspbian.raspberrypi.org_raspbian_dists_bullseye_main_binary-armhf_Packages`. 
Since it is deeper and most of the users won't even know that file, it is more accurate.

theofficialgman:
I already told you cycool29 that getting tzdata does not work on any ubuntu based installs
I know they aren't supported but they should still be identified correctly

theofficialgman:
just because /etc/os-release is editable doesn't mean it shouldn't be used
its a part of a critical package in debian/ubuntu, if the user messes with it, it is their fault

cycool29:
Oh, I just found your reply that the command returns bookworm on Ubuntu focal. That is really a big problem.



> just because /etc/os-release is editable doesn't mean it shouldn't be used its a part of a critical package in debian/ubuntu, if the user messes with it, it is their fault. just like if the user messes with any other config file and breaks their install.
> 
> I agree that `lsb_release` should just be used, it is what add-apt-repository uses as well as other functions, there is no guarantee that the tzdata hack works

Hmm... ya, your reasons are true. Almost everything in Linux is a file and we can't stop users to modify them (even the tzdata hack is also modifiable).
I am changing it to `lsb_release -sc`.

cycool29:
@Botspot Ready to merge.

cycool29:
I purposely use `apt-get` to avoid stable CLI interface warning.

PR #1264, opened 2021-12-12
Add SysMonTask

ryanfortner:
SysMonTask is a Linux system monitor similar to the Windows task manager, written in Python.

https://github.com/KrispyCamel4u/SysMonTask


Itai-Nelken (approved):

Botspot:
```
Package installation complete.
The folder you are executing pip from can no longer be found.
Failed to upgrade psutil
```

Botspot:
```
Package installation complete.
The folder you are executing pip from can no longer be found.
Failed to upgrade psutil
```

ryanfortner:
> ```
> Package installation complete.
> The folder you are executing pip from can no longer be found.
> Failed to upgrade psutil
> ```

Will Sysmontask still launch without running `pip3 install -U psutil`? I know that it won't launch without this step on my buster system but I haven't tried bullseye.

Botspot:
> > ```
> > Package installation complete.
> > The folder you are executing pip from can no longer be found.
> > Failed to upgrade psutil
> > ```
> 
> Will Sysmontask still launch without running `pip3 install -U psutil`?

Yes it did launch. Strange.

Itai-Nelken:
That's because psutil is probably already installed. Maybe from commanderpi which also uses it.

Crilum:
Yeah, It works fine for me, even with the error.. I do have psutil installed, so that's why.. CPU temp doesn't show though..

apps/SysMonTask/install
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+install_packages https://apt.raspbian-addons.org/debian/pool/main/s/sysmontask/sysmontask_1.3.9-ubuntu20.10_all.deb || exit 1
+
+sudo pip3 install -U psutil || error "Failed to upgrade psutil"
```
Crilum, line 5:
```suggestion
pip3 install -U psutil || error "Failed to upgrade psutil"
```
Found the problem.. pip isn't supposed to be run with sudo.

Itai-Nelken: Oh, thats right! Running it as sudo installs it for the root user, not the current user.

Itai-Nelken: @ryanfortner

Botspot:
We're getting many error reports that contain this:
```
The folder you are executing pip from can no longer be found.
Failed to upgrade psutil
```
Any ideas?

Botspot:
I found the problem. It turns out the `install_packages` function was leaving the working directory as `/tmp/pi-apps-local-packages`!

PR #1239, opened 2021-12-03
Add DroidCam

cycool29:
To close #851

PR #1266, opened 2021-12-13
Add pull request template

ryanfortner:


Botspot:
While this idea makes sense if we were receiving spam PRs, we also have to consider the convenience for the regular riders, like itai, cycool, rpinews, myself, etc.
Setting up a structured system for PRs may increase the time & hassle it takes to create one. I could compare this to the intrusion of governmental regulation, or "red tape".

For example, many of our PRs don't involve code, and others don't need testing. Those that do need testing (like new apps), are adequately tested by multiple people before being merged.

Adding a PR template isn't necessarily bad. But we have to ask ourselves if it's really solving a problem, and if the side-effects of the solution are worth it.

ryanfortner:
@cycool29 @Itai-Nelken what do you think?

cycool29:
Botspot's comments are true, it will take inconvenience for something that doesn't need further information, like updating an app, fixing a typo, etc.

I think we can create multiple templates, instead of *totally* not using PR template:
- Update for apps (like https://github.com/Botspot/pi-apps/pull/1268)
- Minor bugs (like https://github.com/Botspot/pi-apps/pull/1221)
- New apps (like https://github.com/Botspot/pi-apps/pull/1239)
- New functions / big changes (like https://github.com/Botspot/pi-apps/pull/1202)
- Small changes (like https://github.com/Botspot/pi-apps/pull/1280)

ryanfortner:
After thinking it over, I don't really see how this could benefit us. Pull requests aren't as frequently opened as issues are, so at this time I don't think it's necessary.

Also, @cycool29 , you can only have one PR template [see the github docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository), so we would have to put everything in one template.

PR #1270, opened 2021-12-15
rename KeePassX to KeePassXC

ryanfortner:
Sorry I didn't notice this when opening the original pr...


Itai-Nelken:
Pleasebadd a runonce task that changes the name on existing installs. Otherwise it will be added as a new app.

Botspot:
> Please add a runonce task that changes the name on existing installs. Otherwise it will be added as a new app. You can add it in the `gui` script.

It will be added as a new app no matter what, so the best we can do is to use a runonce to remove the old app-folder.

Itai-Nelken:
And change rename existing installs

Botspot:
> And rename existing installs

Yes - rename the `data/status` file like this:
```
mv -f "${DIRECTORY}/data/status/KeePassX" "${DIRECTORY}/data/status/KeePassXC" 2>/dev/null
```

ryanfortner:
I suppose leaving the name without the C is still an option. I don't think users will mind

ryanfortner:
That would make things less complicated anyway

theofficialgman:
KeePassX and KeePassXC are two different projects... I imagine it would be quite confusing if the user wanted KeePassX and instead got KeePassXC. it should probably be renamed even if its not ideal

theofficialgman:
changes implemented in https://github.com/Botspot/pi-apps/commit/86fe356309e7a02cb24ece9f262ef518124858fa

PR #1273, opened 2021-12-17
temps: remove duplicate dependencies

ryanfortner:


apps/Temps/credits
```diff
@@ -1,3 +1,3 @@
 Thanks to Crilum on GitHub for making the scripts.
 Thanks to jackd248 on GitHub for the original temps.
-Thanks to promusic on GitHub for forking and continuing development on this awesome app!
+Thanks to promusic on GitHub for forking and continuing development on this awesome app!
```
touhidurrr, line 3: Just curious why you removed the last newline. Isn't keeping newline in at the end of a file is considered a good practice?

PR #1274, opened 2021-12-17
Add Unciv app

touhidurrr:
Add Unciv 3.18.10-patch1 app


apps/Unciv/description
```diff
@@ -0,0 +1,3 @@
+An open-source, mod-friendly Android+Desktop remake of Civ V
+
+To run: Menu -> Games -> Sublime Text
```
Botspot, line 3: Could you fix the "To run" entry?

apps/Unciv/install-32
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+### make directory ~/Unciv if not available
+if [[ ! -d ~/Unciv ]]; then
+  mkdir ~/Unciv || error 'Failed to make folder $HOME/Unciv!'
+fi
+
+### Add LICENSE from Github if doesn't exist
+if [[ ! -f ~/Unciv/LICENSE ]]; then
+  wget -O ~/Unciv/LICENSE ~ https://raw.githubusercontent.com/yairm210/Unciv/master/LICENSE || error 'Failed to make folder $HOME/Unciv!'
```
Botspot, line 10: The error message on this line does not make sense. Also, why is there a `~` character in this wget command?

apps/Unciv/install-32
```diff
@@ -0,0 +1,57 @@
...
+  # Extract jre folder
+  tar --overwrite -xvf ~/zulu8.58.0.13-ca-jdk8.0.312-linux_aarch32hf.tar.gz --wildcards */jre || error 'Failed to ectract jre for Unciv!'
+
+  # move jre to ~/Unciv
+  mv ~/zulu8.58.0.13-ca-jdk8.0.312-linux_aarch32hf/jre ~/Unciv/jre || error "Failed to move jre for Unciv to $HOME/Unciv!"
+fi
+
+## remove artifacts
+if [[ -f ~/zulu8.58.0.13-ca-jdk8.0.312-linux_aarch32hf.tar.gz ]]; then
+  rm ~/zulu8.58.0.13-ca-jdk8.0.312-linux_aarch32hf.tar.gz || echo 'Failed to remove jre artifatcs! Skipping ...'
```
Botspot, line 35: You could simplify this by removing the `if` statement and using `rm -f`.

apps/Unciv/install-32
touhidurrr: Unintended. Fixed in https://github.com/touhidurrr/pi-apps/commit/ff9b43b14eb696080bd03d096f4f9c2d43f42236.

apps/Unciv/install-32
touhidurrr: Ok, yes, it works. Both for folder and file. Fixed in https://github.com/touhidurrr/pi-apps/commit/00638217a58f8323b8cd04dea85caed6463a73ed.

apps/Unciv/uninstall
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+rm -f ~/.local/share/applications/unciv.desktop
+rm -rf ~/Unciv
+purge_packages || exit 1
```
touhidurrr, line 5: @Botspot fixed the mistakes that you pointed out. However, I am wondering what this line does. I copied it from another app entry that I was using as a reference. But don't know what it does.

touhidurrr:
@Botspot fixed the mistakes that you pointed out. However, I am wondering what this line does. I copied it from another app entry that I was using as a reference. But don't know what it does.
https://github.com/touhidurrr/pi-apps/blob/00638217a58f8323b8cd04dea85caed6463a73ed/apps/Unciv/uninstall#L5

apps/Unciv/description
touhidurrr: Fixed in https://github.com/Botspot/pi-apps/pull/1274/commits/277c85b4483299b52082f17881fa6c5667e03748.

touhidurrr:
Also since the Unciv app is updated very frequently, I want to add some kind of auto update mechanism to it. Can I do it?

cycool29:
> However, I am wondering what this line does. I copied it from another app entry that I was using as a reference. But don't know what it does.
https://github.com/touhidurrr/pi-apps/blob/00638217a58f8323b8cd04dea85caed6463a73ed/apps/Unciv/uninstall#L5


That is to remove APT packages installed when installing the app.
Since the app don't install packages when install, so that line is unneeded.

touhidurrr:
> That is to remove APT packages installed when installing the app.
Since the app don't install packages when install, so that line is unneeded.

Very vague. How does it figure out if any apt package was installed when installing the app?

cycool29:
> Very vague. How does it figure out if any apt package was installed when installing the app? 

https://github.com/Botspot/pi-apps/blob/master/DOCUMENTATION.md#purpose-1

> Should I remove it or not?

Yes.

PR #1285, opened 2021-12-20
Temurin repo addition

theofficialgman:
transitions MMC and Lunar client to use temurin apt repo and safe installation process.
~~Currently untested~~
edit: tested now.... just not sure I like what Lunar client needs to work

### DIFF
--- a/apps/Minecraft Java MultiMC5/install
+++ b/apps/Minecraft Java MultiMC5/install
@@ -44,86 +44,71 @@ case "$__os_id" in
     Raspbian|Debian)
         case "$__os_codename" in
             bullseye|buster|stretch|jessie)
-                install_packages wget apt-transport-https gnupg || error "Failed to install dependencies"
+                install_packages lsb-release wget apt-transport-https gnupg || error "Failed to install dependencies"
                 hash -r
-                cd /tmp
-                rm -rf public
-                rm -rf adoptopenjdk-keyring.gpg
-                wget https://adoptopenjdk.jfrog.io/adoptopenjdk/api/gpg/key/public
-                gpg --no-default-keyring --keyring ./adoptopenjdk-keyring.gpg --import public
-                gpg --no-default-keyring --keyring ./adoptopenjdk-keyring.gpg --export --output adoptopenjdk-archive-keyring.gpg 
-                rm -rf public
-                rm -rf adoptopenjdk-keyring.gpg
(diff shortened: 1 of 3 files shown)


theofficialgman:
MMC works
Lunar client does not. It appears that the java directory is hardcoded inside the appimage to /opt/jdk/jdk-17.0.1+12/bin/java @gl91306 is this true? I was under the impression that since pi-apps set the JAVA_HOME and PATH variables that we could specify our own java directory

theofficialgman:
@Botspot Do NOT merge until this function in lunar client has been patched:
```
    function getJavaExecutable(oe) {
        const se = oe.executablePathInArchive.slice();
        return "/opt/jdk/jdk-17.0.1+12/bin/java"
        console.log(oe);
        console.log(oe.download.url);
```

theofficialgman:
@Botspot @gl91306 in the meantime before Lunar client gets a new build with a better java location or way to actually read the java_home variable, I have pushed a hotfix patch to symlink the old location of the java bin to the new location of the java bin.

If you don't want to do this botspot, I will revert the changes to lunar client in this PR and it can be updated to use the temurin repo at a later date once the lunar client appimage is updated

theofficialgman:
force pushed the changes just to make the commits cleaner and make easier to revert if necessary

theofficialgman:
force pushed so incase anyone checks out my branch, they will have the required api functions from the current master
no changes

Botspot:
I've opened an issue on the lunar repo. https://github.com/gl91306/lunar/issues/23

touhidurrr (changes requested):
This will edit keyrings every time an update is available. Why not you do something like this so that doesn't happen.
```bash
if [ ! -f /path/to/keyring ]; then
  Try to add keyring
  Try to add repository
fi
```

touhidurrr:
Also how to do code reveiw?

theofficialgman:
@touhidurrr please view latest commit: https://github.com/Botspot/pi-apps/pull/1285/commits/ef8679eb5a3eb360991b0b2ff4d653f1c202b2ac
this should alleviate your concerns

theofficialgman:
added "only remove repo if safe" to Lunar and Unciv

theofficialgman:
@Botspot I've removed the lunar client from this PR, so this PR can be merged without issues

I cherry picked the changes I had here for it and the new lunar client PR can be tracked here if using the temurin apt repo ever comes to it: https://github.com/Botspot/pi-apps/pull/1305

PR #1286, opened 2021-12-21
Use instant binary instead of adding adoptium repo

cycool29:
Just want to make the installation simple, and easy for uninstallation.

### DIFF
--- a/apps/Unciv/install
+++ b/apps/Unciv/install
@@ -7,44 +7,21 @@ install_packages lsb-release wget gpg || error "Failed to install dependencies"
 mkdir -p $HOME/Unciv || error "Failed to create folder $HOME/Unciv"
 
 ### Add LICENSE from Github if doesn't exist
-if [[ ! -f ~/Unciv/LICENSE ]]; then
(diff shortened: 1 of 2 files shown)


Botspot:
@theofficialgman

theofficialgman:
no @cycool29 this is exactly what we are trying to avoid
the temurin repo is up now. all scripts will be using it if they need functional java
we don't want each and every script to download its own java version, taking up extra valuable space on a users install

making the number of lines of code less does not make it simple

theofficialgman:
for example: now pi-apps has (or will soon have) 4 applications all that need working java versions
Unciv, Minecraft Java, Minecraft Java MultiMC5, and Minecraft Java Server.

if each of these scripts installed their own java versions in their own folders instead of using the apt repo, the size of the JDK installed would be 4X (which is significant when the JRE is about 80MB, and JDK is over 300MB)

cycool29:
What I think is the GPG key is not deleted when uninstalling. 
Since `remove_repofile_if_unused` is here, that problem is solved.

PR #1289, opened 2021-12-21
Unciv: version update to 3.18.11-patch1

touhidurrr:
* update version to 3.18.11
* reduce unnecessary script execution
* try to continue wget download to save time and data

### DIFF
--- a/apps/Unciv/install
+++ b/apps/Unciv/install
@@ -17,25 +17,23 @@ fi
 # so, we will use Temurin OpenJDK 8 installation cause thats the minimal Java version Unciv needs
 # Temurin very smoothly on Raspberry Pi 3B+ without any problems
 
-if [[ ! -f /usr/share/keyrings/adoptium-archive-keyring.gpg ]]; then
-  status "Adding Adoptium repository:"
+status "Adding Adoptium repository:"
 
-  echo "- public key -> keyring"
-  rm -f /tmp/adoptium-public-key /tmp/adoptium-archive-keyring.gpg
-  wget -O /tmp/adoptium-public-key https://adoptium.jfrog.io/artifactory/api/security/keypair/default-gpg-key/public
-  gpg --no-default-keyring --keyring /tmp/adoptium-keyring.gpg --import /tmp/adoptium-public-key
-  rm -f /tmp/adoptium-public-key
+echo "- public key -> keyring"
+rm -f /tmp/adoptium-public-key /tmp/adoptium-archive-keyring.gpg
+wget -O /tmp/adoptium-public-key https://adoptium.jfrog.io/artifactory/api/security/keypair/default-gpg-key/public
+gpg --no-default-keyring --keyring /tmp/adoptium-keyring.gpg --import /tmp/adoptium-public-key
+rm -f /tmp/adoptium-public-key
 
-  echo " - keyring -> GPG key"
-  gpg --no-default-keyring --keyring /tmp/adoptium-keyring.gpg --export --output /tmp/adoptium-archive-keyring.gpg
-  rm -f /tmp/adoptium-keyring.gpg
+echo " - keyring -> GPG key"
+gpg --no-default-keyring --keyring /tmp/adoptium-keyring.gpg --export --output /tmp/adoptium-archive-keyring.gpg
+rm -f /tmp/adoptium-keyring.gpg
 
-  echo " - Moving GPG key to /usr/share/keyrings"
-  sudo mv -f /tmp/adoptium-archive-keyring.gpg /usr/share/keyrings
+echo " - Moving GPG key to /usr/share/keyrings"
+sudo mv -f /tmp/adoptium-archive-keyring.gpg /usr/share/keyrings
 
-  echo " - Creating /etc/apt/sources.list.d/adoptium.list"
-  echo "deb [signed-by=/usr/share/keyrings/adoptium-archive-keyring.gpg] https://adoptium.jfrog.io/artifactory/deb $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/adoptium.list >/dev/null
-fi
+echo " - Creating /etc/apt/sources.list.d/adoptium.list"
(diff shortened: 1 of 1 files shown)


apps/Unciv/install
```diff
@@ -17,34 +17,42 @@ fi
...
+if [ $? != 0 ]; then
   sudo rm -f /etc/apt/sources.list.d/adoptium.list /usr/share/keyrings/adoptium-archive-keyring.gpg
   error "Failed to install temurin packages. Adoptium repository has been removed."
 fi
 
-### Download Unciv.jar
-wget -O "$HOME/Unciv/Unciv.jar" https://github.com/yairm210/Unciv/releases/download/3.18.10-patch1/Unciv.jar || exit 1
+status "Downloading Unciv.jar"
+echo '- writing into /tmp/Unciv.jar'
+wget -c -O /tmp/Unciv.jar https://github.com/yairm210/Unciv/releases/download/3.18.11/Unciv.jar \
```
Botspot, line 50: While adding the `-c` flag may make sense if we were really running the `wget` utility, that's not what's happening here.
Pi-Apps has a _function_ named `wget`. This function automatically switches to use `aria2c` - it's faster, more reliable, and continues downloads by default.
In some cases, `aria2c` cannot completely replace wget, due to how each download utility works. That's why this function will fallback to use `wget` if any unknown flags are passed. In this case, you've added the `-c` flag, and that will cause the `wget` function to fallback to use the `wget` command.

touhidurrr: You really have an unique naming sense. If you were making function that `get` anyways. then name it as something like `download` so that people can read its description from the documentaion !!! why wrap `aria2c` as `wget` !!
Anyways, I will read `aria2c` docs and edit that.

touhidurrr: Ok, I read the docs, it seems that the `-c` flag is not necessary here. Although I would personally prefer `wget` here. I am removing it in https://github.com/Botspot/pi-apps/pull/1289/commits/233a21a95ede8e6feadc7ac5ab793d73208b9223.

Anyways, @Botspot, according to the docs `aria2c` uses `--allow-overwrite` flag, which is problematic.  The behavious of `-c` flag in `aria2c` is not like `wget`. If somehow the script throws error at any point in the script, `aria2c` will download all files again for the flag `--allow-overwrite`. Which seems like a bug to me.

touhidurrr:
@Botspot, read https://github.com/Botspot/pi-apps/pull/1289#discussion_r773335873, maybe its a bug.

apps/Unciv/install
```diff
@@ -17,34 +17,42 @@ fi
 # so, we will use Temurin OpenJDK 8 installation cause thats the minimal Java version Unciv needs
 # Temurin very smoothly on Raspberry Pi 3B+ without any problems
 
-status "Adding Adoptium repository:"
+if [[ ! -f /usr/share/keyrings/adoptium-archive-keyring.gpg ]]; then
```
theofficialgman, line 20: who said the keyring might not change in the future? I don't really see any benefit to not running the "generate a new keyring" part of the script on every update, it will just make sure that the latest keyring is always installed

touhidurrr: bruh, you guys are just increasing the update time. treating updates and first time installation is a problem itself. again, when did you last say a public keyring change? never saw it happen for a official repo myself. stop making up unrealistic arguments like kids.

touhidurrr: doing more thing is just making path for more errors to happen. a download failure is not rare. so no need for downloading things again and again that we do not need.

apps/Unciv/install
cycool29: > Anyways, @Botspot, according to the docs aria2c uses --allow-overwrite flag, which is problematic. The behavious of -c flag in aria2c is not like wget. If somehow the script throws error at any point in the script, aria2c will download all files again for the flag --allow-overwrite. Which seems like a bug to me.


I made a PR to remove the `-c` flag before but I don't know why the `-c` flag is added again...

apps/Unciv/install
theofficialgman: the repo is brand new, it has already gotten a new signing key once (about 1 week ago). You can view the history of this repo addition at the issue relating to it:https://github.com/adoptium/installer/issues/330

I don't expect another key change to occur, but that doesn't mean it won't, which is why I suggested leaving it in on every update/install.
and fyi, the process of downloading, formatting, and adding the key takes well less than a second, so its not increasing the update time in any significant manner

touhidurrr: reverted the changes in https://github.com/Botspot/pi-apps/pull/1289/commits/1d988707a1a1c0203545b492def38d450f549bb2

apps/Unciv/install
```diff
@@ -35,16 +35,22 @@ sudo mv -f /tmp/adoptium-archive-keyring.gpg /usr/share/keyrings
...
-### Download Unciv.jar
-wget -O "$HOME/Unciv/Unciv.jar" https://github.com/yairm210/Unciv/releases/download/3.18.10-patch1/Unciv.jar || exit 1
+status "Downloading Unciv.jar"
+echo '- writing into /tmp/Unciv.jar'
+wget -O /tmp/Unciv.jar https://github.com/yairm210/Unciv/releases/download/3.18.11-patch1/Unciv.jar \
+    || error 'Failed to download Unciv.jar!'
+echo '- moving /tmp/Unciv.jar to ~/Unciv/Unciv.jar'
+mv /tmp/Unciv.jar ~/Unciv/Unciv.jar || exit 1
+echo '- removing /tmp/Unciv.jar'
+rm -f /tmp/Unciv.jar
```
Botspot, line 53: Why are we downloading to /tmp/Unciv.jar, then moving that file to ~/Unciv/Unciv.jar? Is there some safety or performance benefit to doing this?

touhidurrr: yes. it ensures that the already downloaded `Unciv.jar` files doesn't get corrupt when the update fails. So, even if the download when updating fails, the user can play the previous version without any problem.

touhidurrr: it can even enable users update the app while playing probably.

PR #1301, opened 2021-12-25
Add change settings menu icon `runonce`

cycool29:


gui
```diff
@@ -172,6 +172,8 @@ runonce <<"EOF"
   refresh_all_pkgapp_status
 EOF
 
+runonce <<"EOF"
+  [ -a "$HOME/.local/share/applications/pi-apps-settings.desktop" ] && sed -i "s/logo.png/settings.png/g" "$HOME/.local/share/applications/pi-apps-settings.desktop"
```
Itai-Nelken, line 176: I think an `EOF` is needed at the end as well.

cycool29: Fixed. Thanks for noticing.

gui
```diff
@@ -172,6 +172,10 @@ runonce <<"EOF"
...
```
Botspot, line 176: For the record, if the final command in a runonce exits with a non-zero exit code (indicating failure), the `runonce` function will note its failure and continue trying indefinitely.
A `true` command needs to be added before the EOF to prevent this from happening.

cycool29: This works for me:

```bash
[ -a "$HOME/.local/share/applications/pi-apps-settings.desktop" ] && sed -i "s/logo.png/settings.png/g" "$HOME/.local/share/applications/pi-apps-settings.desktop" || true
```

cycool29:
@Botspot This PR is ready to merge.

gui
```diff
@@ -157,6 +157,12 @@ runonce <<"EOF"
   
   refresh_all_pkgapp_status
 EOF
+
+# change settings icon
+runonce <<"EOF"
+  [ -a "$HOME/.local/share/applications/pi-apps-settings.desktop" ] && sed -i "s/logo.png/settings.png/g" "$HOME/.local/share/applications/pi-apps-settings.desktop" || true
```
Botspot, line 163: Why are you using the `-a` flag instead of the `-f` flag here?

cycool29: From `help test`:

```
      -a FILE        True if file exists.
      -f FILE        True if file exists and is a regular file.
```

While `"$HOME/.local/share/applications/pi-apps-settings.desktop"` will mostly be a regular file, I don't think there is any differences between `-a` and `-f`.

PR #1302, opened 2021-12-25
better implementation of getting `screen_width` and `screen_height` in `updater` script

touhidurrr:
this is a better and more universal implementation

### DIFF
--- a/updater
+++ b/updater
@@ -278,8 +278,8 @@ update_now_cli() { #input: updatable_files and updatable_apps variables
   status_green "\nPi-Apps updates complete."
 }
 
-screen_width="$(xrandr | grep "HDMI-1" | awk '{print $4}' | tr 'x+' ' ' | awk '{print $1}')"
-screen_height="$(xrandr | grep "HDMI-1" | awk '{print $4}' | tr 'x+' ' ' | awk '{print $2}')"
+screen_width="$(xrandr | grep -oP '(?<=primary )\d+')"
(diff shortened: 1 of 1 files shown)


theofficialgman:
can confirm, this should work nicely

Botspot:
I will change the implementation in the `gui` script to match this.

touhidurrr:
@Botspot @theofficialgman wanted you guys to look at this post.
https://unix.stackexchange.com/a/683902/383916
Someone is saying that it is not **POSIX** Complient.

theofficialgman:
I don't think POSIX compliance has ever been a concern for Pi-Apps. The reason Bash is used over Sh is so we get those non-POSIX compliant command structures (example: arrays are not POSIX compliant and are not included in Sh as a result).

touhidurrr:
> I don't think POSIX compliance has ever been a concern for Pi-Apps. The reason Bash is used over Sh is so we get those non-POSIX compliant command structures (example: arrays are not POSIX compliant and are not included in Sh as a result).

Makes sense. But I still wanted to show you guys the thread for documentation reasons and future troubleshooting.

Botspot:
> > I don't think POSIX compliance has ever been a concern for Pi-Apps. The reason Bash is used over Sh is so we get those non-POSIX compliant command structures (example: arrays are not POSIX compliant and are not included in Sh as a result).
> 
> Makes sense. But I still wanted to show you guys the thread for documentation reasons and future troubleshooting.

Thank you. I was curious about how that regex was working.

PR #1313, opened 2021-12-28
Exclude template in `list_apps`

cycool29:


Botspot:
Pi-Apps used to explicitly hide `template` for some actions, but I expanded the behaviour to hide all apps that are in the `hidden` category. By default, only the `template` app is `hidden`, but on TiwsterOS, a few more apps are hidden.

I'd rather not give special treatment to the `template` app again. Have you found a scenario where it's visible but should not be visible?

cycool29:
> I'd rather not give special treatment to the template app again. Have you found a scenario where it's visible but should not be visible?

In `list-*` of Pi-Apps Terminal Plugin. 
https://github.com/cycool29/PiApps-terminal_bash-edition/blob/5d34bda7dfe6dbde78d917fcc5e86103cc0d6898/pi-apps-terminal-bash-edition.sh#L147-L232

I know I can manually `grep` out `template`, but since I think the `template` app is also unneeded in those `list_apps` functions in `api`, so I open a PR here first.

PR #1321, opened 2021-12-29
Update README.md

derryktutt:
Made small improvements
(yes, amogos is real)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,11 +42,13 @@ wget -qO- https://raw.githubusercontent.com/Botspot/pi-apps/master/install | bas
 <img src="icons/screenshots/main%20window.png?raw=true" align="right" height="270px"/>
 Supported systems:
 
-- Raspberry Pi OS 32-bit: fully supported.
-- Twister OS: fully supported.
-- Raspberry Pi OS 64-bit: fully supported.
-- Bullseye, Kali, Ubuntu: you may encounter errors for some apps.
-- Android, ChromeOS, non-ARM, non-Debian: Not supported. Your mileage may vary.
+- <a href="https://www.raspberrypi.com/software/operating-systems/">
+    Raspberry Pi OS 32 and 64-bit</a>: fully supported.
+- <a href="https://twisteros.com/download.html">Twister OS</a>: fully supported, preinstalled.
+- <a href="https://www.amogos.studio/">
+    AmogOS</a>: fully supported, preinstalled.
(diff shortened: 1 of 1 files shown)


Botspot:
AmogOS is also available for x86 PCs, but Pi-Apps is not. If we list AmogOS as "supported", then some users may get the false impression that Pi-Apps is supported on x86.
Can we get an expert opinion from @mobilegmYT ? :P

mobilegmYT:
> AmogOS is also available for x86 PCs, but Pi-Apps is not. If we list AmogOS as "supported", then some users may get the false impression that Pi-Apps is supported on x86. Can we get an expert opinion from @mobilegmYT ? :P

I mean technically Pi Apps x86 is preinstalled...

derryktutt:
The AmogOS github page says "Pi-Apps x86 for the x64 build." I will try it in a vm later today.

derryktutt:
Yes, Pi-apps is there.

Crilum:
> Yes, Pi-apps is there.

Are you on an arm build or a x86/x64 build?

derryktutt:
Oh, I was using Arm.

Botspot:
> I mean technically Pi Apps x86 is preinstalled...

But **_we_** don't support it. Not here, at least.

Crilum:
> Oh, I was using Arm.


Did it have the real Pi-Apps, or the x86 version?

derryktutt:
The raspberry pi version has the real one. AmogOS for the raspberry pi is
just Rasbian but it is among us themed.

Crilum:
So we support AmogOS arm?

derryktutt:
If you install AmogOS on the raspberry pi, from AmogOS.studio, Pi-apps is
pre-installed.

derryktutt:
I used AmogOS on distrowatch, and took a screen recording. It has Pi-apps x86. I do not know if it works, because Distrowatch has no internet
[amogOS.mp4.zip](https://github.com/Botspot/pi-apps/files/7793919/amogOS.mp4.zip)
.

Crilum:
Yeah, it was running x86_64 AmogOS, so..

Crilum:
@mobilegmYT correct me if I'm wrong, Raspberry Pi AmogOS has Pi-Apps, and AmogOS x86 has Pi-Apps x86, right?

derryktutt:
> @mobilegmYT correct me if I'm wrong, Raspberry Pi AmogOS has Pi-Apps, and AmogOS x86 has Pi-Apps x86, right?

That is correct.

Crilum:
@Botspot I think we're good now..

PR #1322, opened 2021-12-29
MMC add Automatic Mod installer script from the L4T-Megascript

theofficialgman:
edit: Do not merge

currently only supported on Debian Buster+ and Ubuntu Bionic+

### DIFF
--- a/apps/Minecraft Java MultiMC5/description
+++ b/apps/Minecraft Java MultiMC5/description
@@ -1,6 +1,7 @@
 CUSTOM BUILD of MultiMC5/Launcher code for ARMhf/ARM64 support. Includes Microsoft, Mojang, and PC GamePass Account support. This launcher allows you to easily manage multiple instances of minecraft and your mods at once.
 
 NOTE: The use of performance enhancing mods is highly encouraged, such as optifine with Optifabric and Fabric/Forge or Sodium/Lithium/Phosphor Fabric Mods.
+A prelaunch script can be installed through this app to automate performance mod installation if the user so chooses.
 
 Minecraft Java MultiMC5 for the Raspberry Pi
 To run: Menu -> Games -> MultiMC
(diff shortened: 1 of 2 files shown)


cobalt2727:
I can't check right now, but if this has ARM packages, try pulling it in for Buster users
https://people.debian.org/~paravoid/python-all/

theofficialgman:
> I can't check right now, but if this has ARM packages, try pulling it in for Buster users https://people.debian.org/~paravoid/python-all/

see the discord convo, I already checked and it does not

however, this one does https://community.home-assistant.io/t/home-assistant-core-python-3-8-backport-for-debian-buster/234859

cobalt2727:
Definitely go for that then

theofficialgman:
ok I'm happy enough with this

If the mod installer script fails at any point during installation (like missing dependencies, etc), a warning is thrown to the user and then gets skipped.
its still only really useful if the user in on 1.16+ or if they supply their own mods for older versions and just want to keep them updated. The script can be updated asynchronously (it self updates from github), so more functionality could come in the future if I add it.

theofficialgman:
changing this back to a draft PR as main sodium dev (jellysquid) has decided to make OpenGL 4.5 a hard requirement (and a lot of its featureset) for future releases https://github.com/CaffeineMC/sodium-fabric/commit/69fb6d3616ecee7df042703ba35b21f49d0c55b2

sodium is part of the automatic mod installer script, all of the other mods are still compatible but this was an important one for pis that will probably not be viable in the future

cobalt2727:
What version does Iris require?

theofficialgman:
> What version does Iris require?

Iris is not relevant to this change. Iris and sodium are separate mods. You can use sodium with iris to get improved performance, iris alone does not bring any.

However, I have been informed by jellysquid that Raspberry Pis may still be viable even on the 1.18/next branch (where the changes have occured) since the mesa driver does appear to have all of the features required of it by sodium next. I will re-mark this PR for ready for review is someone can verify that this github actions build still work on raspberry pi 1.18.1 https://github.com/CaffeineMC/sodium-fabric/actions/runs/1647243537

theofficialgman:
Still keeping this as a draft as nothing has changed on the sodium end (it still looks like opengl 4.6+ will be a requirement for upcoming versions)

This PR may be unnecessary in the near future with the coming of modrinth modpack support to MultiMC (we could make performance modpacks for the Pi, or just have a suggested modpack)

theofficialgman:
with modrinth modpack now a thing in multimc, curseforge removal (unusable due to API TOS being incompatible with any launcher), and sodium instability (changes to new maintainers), there isn't a reason to keep this open

PR #1323, opened 2021-12-30
Api update (install_packages and failed installs)

theofficialgman:
- Removed the unnecessary dummy deb package purge if install_packages is run multiple times in a script
- moved dummy deb package purge to occur if the entire app install script fails (that way dummy debs for a "corrupted" install do not remain on the users OS)

these changes now allow for script writers to implement `install_packages` in their script and allow for it to fail non-fatally and keep the previous Depends: packages

for example: its nice to have adoptopenjdk-16 but not necessary for MultiMC to work, so install_packages for this specific package is made non-fatal, but without the changes here, ALL other previously installed dependencies are removed from the dummy deb since it is purged

for scripts where failing `install_packages` is a fatal error, there is no change as the dummy deb gets purged by the manage script now.

### DIFF
--- a/api
+++ b/api
@@ -427,18 +427,7 @@ Package: $package_name" > ~/$package_name/DEBIAN/control
   
   #install dummy deb
   {
-  #If this app's dummy deb is already insalled, uninstall it
-  if package_installed "$package_name" ;then
-    status "Reinstalling the $package_name package..."
-    apt_lock_wait
-    local output="$(sudo -E apt purge -y $package_name 2>&1 | less_apt | tee /dev/stderr)"
-    if [ $? != 0 ];then
-      echo "$output"
-      error "Failed to purge dummy deb ($package_name)"
-    fi
-  else
-    status "Installing the $package_name package..."
-  fi
+  status "Installing the $package_name package..."
   
   apt_lock_wait
   local output="$(sudo -E apt install -fy --no-install-recommends --allow-downgrades "${apt_flags[@]}" ~/$package_name.deb 2>&1 | less_apt | tee /dev/stderr)"
--- a/manage
+++ b/manage
@@ -277,14 +277,22 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
   else #if app failed to install/uninstall
 
     #remove dummy deb if app failed to install (to avoid dummy debs being left on a users install for broken apps)
-    if [[ ${mode} == "install" ]] && [[ "$(app_type "$app")" == standard ]]; then
-      local package_name="$(app_to_pkgname "$app")"
-      local output="$(sudo -E apt purge -y $package_name 2>&1 | less_apt | tee /dev/stderr)"
-      if [ $? != 0 ];then
-        echo "$output"
-        warning "Failed to purge dummy deb ($package_name)"
+    package_name="$(app_to_pkgname "$app")"
+    if [[ ${mode} == "install" ]] && [[ "$(app_type "$app")" == standard ]] && package_installed "$package_name"; then
+      if [[ "$script_input" == "update" ]]; then
+        #if install failed as part of an update, don't autoremove the packages, only purge the dummy deb
+        output="$(sudo -E apt purge -y $package_name 2>&1 | less_apt | tee /dev/stderr)"
+        if [ $? != 0 ];then
+          echo "$output"
+          warning "Failed to purge dummy deb ($package_name)"
+        fi
+        unset output
+      else
+        #install and not update failed, run purge_packages
+        purge_packages || warning "Failed to purge dummy deb ($package_name)"
       fi
     fi
+    unset package_name
 
     echo -e "\n\e[91mFailed to ${mode} ${app}!\e[39m
 \e[40m\e[93m\e[5m◢◣\e[25m\e[39m\e[49m\e[93mNeed help? Copy the \e[1mENTIRE\e[0m\e[49m\e[93m terminal output or take a screenshot.
--- a/manage
+++ b/manage
@@ -277,7 +277,7 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
   else #if app failed to install/uninstall
 
     #remove dummy deb if app failed to install (to avoid dummy debs being left on a users install for broken apps)
-    if [[ ${mode} == "install" ]]; then
+    if [[ ${mode} == "install" ]] && [[ "$(app_type "$app")" == standard ]]; then
       local package_name="$(app_to_pkgname "$app")"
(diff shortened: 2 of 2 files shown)


manage
```diff
@@ -275,6 +275,17 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
     mv "$logfile" "$(echo "$logfile" | sed 's+-incomplete-+-success-+g')" #rename logfile to indicate it was successful
     
   else #if app failed to install/uninstall
+
+    #remove dummy deb if app failed to install (to avoid dummy debs being left on a users install for broken apps)
+    if [[ ${mode} == "install" ]]; then
```
Botspot, line 280: This also needs to check if the app is a `standard` app, not a `package` app.

api
```diff
@@ -427,18 +427,7 @@ Package: $package_name" > ~/$package_name/DEBIAN/control
   
   #install dummy deb
   {
-  #If this app's dummy deb is already insalled, uninstall it
-  if package_installed "$package_name" ;then
-    status "Reinstalling the $package_name package..."
-    apt_lock_wait
-    local output="$(sudo -E apt purge -y $package_name 2>&1 | less_apt | tee /dev/stderr)"
```
Botspot, line 434: When developing `install_packages`, I *really* thought that the dummy deb had to be reinstalled for anything to happen. Otherwise, apt would say something like "XXXX is already installed and the latest version"

theofficialgman: I figured, but that's not the case. Installing over itself works just fine as long as the dummy deb contents have changed

if you try to install it again and the dummy has the same contents as the first time, yes you just get this 
```Note, selecting 
'dummy' instead of '/home/garrett/dummy.deb'

dummy is already the newest version (1.0).
```

but as long as you change the dependencies then it will always install over itself. thats what a second install_packages would do in a script so its perfect
```
Note, selecting 'dummy' instead of '/home/garrett/dummy.deb'
The following packages will be upgraded:
  dummy
1 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
Need to get 0 B/736 B of archives.
After this operation, 0 B of additional disk space will be used.
Get:1 /home/garrett/dummy.deb dummy all 1.0 [736 B]
(Reading database ... 398236 files and directories currently installed.)
Preparing to unpack /home/garrett/dummy.deb ...
Unpacking dummy (1.0) over (1.0) ...
Setting up dummy (1.0) ...
```

manage
theofficialgman: alright should be good now

theofficialgman:
> these changes now allow for script writers to implement `install_packages` in their script and allow for it to fail non-fatally and keep the previous Depends: packages
> 
> for example: its nice to have adoptopenjdk-16 but not necessary for MultiMC to work, so install_packages for this specific package is made non-fatal, but without the changes here, ALL other previously installed dependencies are removed from the dummy deb since it is purged


to add to this, I've noticed at least one other script is are already expecting this non-fatal and don't purge the dummy deb behavior (the new behavior). Specifically, the cura script where if stl-thumb fails to install the script continues and will only throw a warning.

manage
```diff
@@ -275,6 +275,17 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
     mv "$logfile" "$(echo "$logfile" | sed 's+-incomplete-+-success-+g')" #rename logfile to indicate it was successful
     
   else #if app failed to install/uninstall
+
+    #remove dummy deb if app failed to install (to avoid dummy debs being left on a users install for broken apps)
+    if [[ ${mode} == "install" ]] && [[ "$(app_type "$app")" == standard ]]; then
+      local package_name="$(app_to_pkgname "$app")"
+      local output="$(sudo -E apt purge -y $package_name 2>&1 | less_apt | tee /dev/stderr)"
```
Botspot, line 282: Why run `apt purge` rather than `purge_packages`? Also this may run unnecessarily if the dummy deb is not installed, so we need to add a check for that.

theofficialgman: an unnecessary run isn't a big deal, but yeah that can be checked

I ran apt purge because that is what you ran in your script. It mimics the previous functionality by moving it to the manage script only when the app fails while not having the drawbacks of purging the dummy deb unnecessarily during a non-fatal install_packages run

```
  #If this app's dummy deb is already insalled, uninstall it
  if package_installed "$package_name" ;then
    status "Reinstalling the $package_name package..."
    apt_lock_wait
    local output="$(sudo -E apt purge -y $package_name 2>&1 | less_apt | tee /dev/stderr)"
    if [ $? != 0 ];then
      echo "$output"
      error "Failed to purge dummy deb ($package_name)"
    fi
  else
    status "Installing the $package_name package..."
  fi
  ```
  
  the main difference between running what I have written now and doing purge_packages is purge_packages will do an apt autoremove. 
I think an autoremove is fine for a failed first time install but I don't think its a good idea for a failed update.

**Comment by theofficialgman** [comment]
also @Botspot  I just noticed this:
```
    #set the app's status to 'corrupted' if the diagnostics determine the error is NOT internet or system or package, AND if the app is a script-type app
    if ! [[ "$(log_diagnose "$logfile" | head -n1)" =~ ^(system|internet|package)$ ]] && [ "$(app_type "$app")" == app ];then
      echo "corrupted" > "${DIRECTORY}/data/status/${app}"
    fi
```
is this a bug? I thought the only app_type was "standard" and "package". no such thing as "app" and that will prevent all logs for being able to be sent
```
app_type() { #there are 'standard' apps, and there are 'package' apps - an alias to install an apt package from the existing repositories.
  #$1 is input app-name
  local app="$1"
  [ -z "$app" ] && error "app_type: no app specified!"
  
  if [ -f "${DIRECTORY}/apps/${app}/packages" ];then
    echo package
  elif [ -f "${DIRECTORY}/apps/${app}/uninstall" ] || [ -f "${DIRECTORY}/apps/${app}/install" ] || [ -f "${DIRECTORY}/apps/${app}/install-32" ] || [ -f "${DIRECTORY}/apps/${app}/install-64" ];then
    echo standard
  else
    return 1
  fi
  #if neither conditional above evaluated to true, no output will be returned and the function exits with code 1
}
```

theofficialgman:
looks like you introduced the bug 2 weeks ago in this commit https://github.com/Botspot/pi-apps/commit/eea069e028d6ca1a9db95d93eea35a0907a7b94e

Botspot:
Wow you're right.

theofficialgman:
@Botspot I think this addresses your concern https://github.com/Botspot/pi-apps/pull/1323/commits/13f57b8d22f13eda838c17bcdc4b8071c11691b7

(also I assume you will just go ahead and fix that unrelated bug on your own at some point)

PR #1330, opened 2021-12-30
Github actions - update_apps.yaml workflow

theofficialgman:
This workflow is in beta. Currently it runs only when manually told to do so (in the future that will be automatic say once a day).

It also commits directly to the branch which it was run from. It has been suggested that this be changed to instead make a PR every time an app update is found, this could be done with a change of `EndBug/add-and-commit@v7` to one of the many PR actions from the marketplace

### DIFF
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -65,15 +65,11 @@ jobs:
             echo UPDATED_APPS=$updated_apps >> $GITHUB_ENV
           fi
 
-      - name: Push changes to repo
-        uses: EndBug/add-and-commit@v7 # You can change this to use a specific version.
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
         with:
-
-          default_author: github_actions
-
-          # The message for the commit.
-          # Default: 'Commit from GitHub Actions (name of the workflow)'
-          message: |
+          author: github_actions
+          commit-message: |
             Update App Versions, run by GitHub Actions - Apps updated:
 
             ${{ env.UPDATED_APPS }}
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -64,9 +64,9 @@ jobs:
           cd
           if test -f /tmp/updated_apps; then
             sort -u /tmp/updated_apps > /tmp/updated_apps_sorted
-            updated_apps=$(cat /tmp/updated_apps_sorted)
-            rm -f /tmp/updated_apps /tmp/updated_apps_sorted
-            echo UPDATED_APPS=$updated_apps >> $GITHUB_ENV
+            echo "UPDATED_APPS<<EOF" >> $GITHUB_ENV
+            cat /tmp/updated_apps_sorted >> $GITHUB_ENV
+            echo "EOF" >> $GITHUB_ENV
           fi
 
       - name: Create Pull Request
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -1,7 +1,10 @@
 name: Update_Apps
 
 # Controls when the workflow will run
-on:  
+on:
+  # run at 21:00 UTC every day
+  schedule:
+     - cron: '0 21 * * *'
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
@@ -74,6 +77,7 @@ jobs:
             Update App Versions, run by GitHub Actions - Apps updated:
 
             ${{ env.UPDATED_APPS }}
+          branch: "auto-app-updates"
           title: "[AUTO] Update App Versions"
           body: |
             Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action:
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -60,15 +60,16 @@ jobs:
           done
           cd
           if test -f /tmp/updated_apps; then
-            updated_apps=$(cat /tmp/updated_apps)
-            rm -f /tmp/updated_apps
+            sort /tmp/updated_apps | uniq -u > /tmp/updated_apps_sorted
+            updated_apps=$(cat /tmp/updated_apps_sorted)
+            rm -f /tmp/updated_apps /tmp/updated_apps_sorted
             echo UPDATED_APPS=$updated_apps >> $GITHUB_ENV
           fi
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
-          author: github-actions <https://github.com/apps/github-actions>
+          author: github-actions <41898282+github-actions[bot]@users.noreply.github.com>
           commit-message: |
             Update App Versions, run by GitHub Actions - Apps updated:
 
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -47,13 +47,14 @@ jobs:
           chmod +x $GITHUB_WORKSPACE/.github/workflows/updates/*.sh
           cd $GITHUB_WORKSPACE
           apps=( .github/workflows/updates/*.sh )
-          for app_directory in ${apps[@]}; do
+          for app_directory in "${apps[@]}"; do
             #make sure we are still in the main workspace (incase an update script left off elsewhere)
             cd $GITHUB_WORKSPACE
-            export app_name=$(echo ${app_directory%.*} | sed 's:.*/::')
+            export app_name="$(echo ${app_directory%.*} | sed 's:.*/::')"
+            echo "$app_name"
             status "Checking $app_name for updates"
             # move to app folder
-            cd "./apps/$app_name"
+            cd "$GITHUB_WORKSPACE/apps/$app_name"
             # run app update script
             "$GITHUB_WORKSPACE/$app_directory"
           done
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -1,0 +1,78 @@
+name: Update_Apps
+
+# Controls when the workflow will run
+on:  
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  update-apps:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      # Runs a set of commands using the runners shell
+      - name: Run app update scripts
+        run: |
+          #source pi-apps functions
+
+          #export all functions and variables
+          set -a
+          #make DIRECTORY equal to GITHUB_WORKSPACE, for subscripts and api functions
+          DIRECTORY=$GITHUB_WORKSPACE
+          source $GITHUB_WORKSPACE/api
+
+          #add special functions
+          get_release() {
+            curl --silent "https://api.github.com/repos/$1/releases/latest" | jq -r '.tag_name' | sed s/v//g
+          }
+
+          function validate_url(){
+            if [[ `wget -S --spider $1  2>&1 | grep 'HTTP/1.1 200 OK'` ]]; then
+              return 0
+            else
+              return 1
+            fi
+          }
+
+          #stop exporting functions
+          set +a
+          
+          #make sure all update scripts are executable
+          chmod +x $GITHUB_WORKSPACE/.github/workflows/updates/*.sh
+          cd $GITHUB_WORKSPACE
+          apps=( .github/workflows/updates/*.sh )
+          for app_directory in ${apps[@]}; do
+            #make sure we are still in the main workspace (incase an update script left off elsewhere)
+            cd $GITHUB_WORKSPACE
+            export app_name=$(echo ${app_directory%.*} | sed 's:.*/::')
+            status "Checking $app_name for updates"
+            # move to app folder
+            cd "./apps/$app_name"
+            # run app update script
+            "$GITHUB_WORKSPACE/$app_directory"
+          done
+          cd
+          if test -f /tmp/updated_apps; then
+            updated_apps=$(cat /tmp/updated_apps)
+            rm -f /tmp/updated_apps
+            echo UPDATED_APPS=$updated_apps >> $GITHUB_ENV
+          fi
+
+      - name: Push changes to repo
+        uses: EndBug/add-and-commit@v7 # You can change this to use a specific version.
+        with:
+
+          default_author: github_actions
+
+          # The message for the commit.
+          # Default: 'Commit from GitHub Actions (name of the workflow)'
+          message: |
+            Update App Versions, run by GitHub Actions - Apps updated:
+
+            ${{ env.UPDATED_APPS }}
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -60,7 +60,7 @@ jobs:
           done
           cd
           if test -f /tmp/updated_apps; then
-            sort /tmp/updated_apps | uniq -u > /tmp/updated_apps_sorted
+            sort -u /tmp/updated_apps > /tmp/updated_apps_sorted
             updated_apps=$(cat /tmp/updated_apps_sorted)
             rm -f /tmp/updated_apps /tmp/updated_apps_sorted
             echo UPDATED_APPS=$updated_apps >> $GITHUB_ENV
@@ -74,7 +74,7 @@ jobs:
             Update App Versions, run by GitHub Actions - Apps updated:
 
             ${{ env.UPDATED_APPS }}
-          title: "[AUTO] Update Apps Versions"
+          title: "[AUTO] Update App Versions"
           body: |
             Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action:
             Apps updated:
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -68,7 +68,14 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
+          author: github-actions <https://github.com/apps/github-actions>
           commit-message: |
             Update App Versions, run by GitHub Actions - Apps updated:
+
+            ${{ env.UPDATED_APPS }}
+          title: "[AUTO] Update Apps Versions"
+          body: |
+            Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action:
+            Apps updated:
 
             ${{ env.UPDATED_APPS }}
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -33,7 +33,7 @@ jobs:
           }
 
           function validate_url(){
-            if [[ `wget -S --spider $1  2>&1 | grep 'HTTP/1.1 200 OK'` ]]; then
+            if command wget -q --spider "$1"; then
               return 0
             else
               return 1
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -68,7 +68,6 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
-          author: github_actions
           commit-message: |
             Update App Versions, run by GitHub Actions - Apps updated:
 
--- a/.github/workflows/update_github_script.sh
+++ b/.github/workflows/update_github_script.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+# to run this script as part of other update scripts, source the directory like below
+# source $GITHUB_WORKSPACE/.github/workflows/update_github_script.sh
+
+
+# for apps with an install-32 file, make sure that you supply the following variables before sourcing this script (examples for webcord shown below)
+# the ${webVer} variable within the url is important
+
+# webVer=$(get_release SpacingBat3/WebCord)
+# armhf_url="https://github.com/SpacingBat3/WebCord/releases/download/v${webVer}/webcord_${webVer}_armhf.deb"
+# pi_apps_ver_32="$(cat 'install-32' | grep 'version=' | sed 's/version=//')"
+
+
+# for apps with an install-64 file, make sure that you supply the following variables before sourcing this script (examples for webcord shown below)
+# if a variable is already set, no need to duplicate it
+
+# webVer=$(get_release SpacingBat3/WebCord)
+# arm64_url="https://github.com/SpacingBat3/WebCord/releases/download/v${webVer}/webcord_${webVer}_arm64.deb"
+# pi_apps_ver_64="$(cat 'install-64' | grep 'version=' | sed 's/version=//')"
+
+
+# for apps with a unified install file, make sure to supply the following variables before sourcing this script
+
+# webVer=$(get_release subhra74/xdm)
+# all_url="https://github.com/subhra74/xdm/releases/download/${webVer}/xdm-setup-${webVer}.tar.xz"
+# pi_apps_ver="$(cat 'install' | grep 'version=' | sed 's/version=//')"
+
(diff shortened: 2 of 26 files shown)


.github/workflows/update_apps.yml
```diff
@@ -0,0 +1,58 @@
...
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      # Runs a set of commands using the runners shell
+      - name: Run a multi-line script
```
Crilum, line 23:
```suggestion
      - name: Run app update scripts
```

.github/workflows/update_apps.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Push changes to repo
+        uses: EndBug/add-and-commit@v7 # You can change this to use a specific version.
+        with:
+
+          default_author: github_actions
+
+          # The message for the commit.
+          # Default: 'Commit from GitHub Actions (name of the workflow)'
+          message: 'Automatic update'
```
Crilum, line 61: Maybe have something a little more descriptive in the message, Like `Automatic update, run by GitHub Actions`.

It'd also be cood if it generatd a list of all apps updated by the workflow!

theofficialgman: yeah I'm still working on this
adding a check to verify that the new version actually exists for the armhf/arm64 (I've seen a couple of github releases drop support for armhf due to issues and don't want to update the version variable if the file doesn't actually exist)

theofficialgman:
ok formatting aside, is this what you had in mind? https://github.com/Botspot/pi-apps/pull/1330/commits/99306efd4c3ab2c9e5e705bf18e8823c4671bcf5

the full logs are viewable in the github actions under "run app update scripts"
https://github.com/theofficialgman/pi-apps/runs/4670552834?check_suite_focus=true

Crilum:
Yeah! That's great!

Crilum:
It looks like GitHub Actions is only updating btop arm64 and webcord armhf though...

theofficialgman:
> It looks like GitHub Actions is only updating btop arm64 and webcord armhf though...

no its working fine
I'm manually force pushing my branch each time so those two are out of date so I can run the actions again and make sure it updates

Crilum:
Oh ok!

Crilum:
So now we just need to write `update` scripts for each Github installed app right? And, obviously the PR needs to be merged...

theofficialgman:
well 90% of the update scripts for discord and btop are copy paste
this should probably be improved so it can handle install-32, install-64, and install and then added to the api or inside the update_apps.yaml as a function
```
# echo the versions
status "The latest online version is:  $webVer"
status "The current version in Pi-Apps install-32 is:  $pi_apps_ver_32"
status "The current version in Pi-Apps install-64 is:  $pi_apps_ver_64"
#Test to see if version is current:
if [[ "$pi_apps_ver_32" = "$webVer" ]]; then
    #If the version is current do:
    status_green 'Pi-Apps install-32 version is current!'
else
    #If version is not current do:
	if validate_url "$armhf_url"; then
		status_green "Updating pi-apps $app_name install-32 to: $webVer"
		sed -i "s/version=.*/version=${webVer}/g" install-32
		echo "$app_name-armhf " >> /tmp/updated_apps
	else
		warning "Updating $app_name install-32 had been skipped, the upstream file $armhf_url does NOT exist."
	fi
fi

if [[ "$pi_apps_ver_64" = "$webVer" ]]; then
    #If the version is current do:
    status_green 'Pi-Apps install-64 version is current!'
else
	if validate_url "$arm64_url"; then
		status_green "Updating pi-apps $app_name install-64 to: $webVer"
		sed -i "s/version=.*/version=${webVer}/g" install-64
		echo "$app_name-arm64 " >> /tmp/updated_apps
	else
		warning "Updating $app_name install-64 had been skipped, the upstream file $arm64_url does NOT exist."
	fi
fi
```

theofficialgman:
obviously scripts are welcome to have their own custom update script that doesn't use this function, but this usecase is common enough across pi-apps that we should avoid duplication

Crilum:
We can make the above into an `api` function, and another one for `install` type apps..

theofficialgman:
I don't really feel like it should be in the api, since it is only run by the github actions.
see the above commits ^

Crilum:
Yeah, ok. We still need a script for no arch apps (`install`), which I think should be put in the `update_github_script.sh`, as a function, and then put the other armhf/arm64 script inside another function.. Which could then be sourced and run in the workflow...

theofficialgman:
I went ahead and gathered a list of the current apps in pi-apps that use github releases and have the `install` script
most download an `all.deb` file so I'll plan on making that case so it only checks one URL
if the script needs a second, it can make its own custom update file that doesn't use this function

Crilum:
For any update scripts I make, do you want me to just post them here and let you add them to your PR?

cycool29:
Why not just make the `update_github_script.sh` in every app update script?

My idea is: A GitHub workflow triggers the main script that stores functions needed and preparation of updates(like `chmod`ing scripts), and a `for` loop triggers each app update script.

cycool29:
Or maybe `update_github_script.sh` can be compressed as a function? I really don't like it as a script.

Crilum:
That's kind of what it already is, and kink of what I suggested:

to add the `update_github_script.sh` as a function in `api`, but @theofficialgman didn't like it that way so..

theofficialgman:
botspot previously told me he did not want scripts/functions that weren't intended to be run by the user (like these github actions scripts) to be in the normal pi-apps directories

so thats why the scripts and the sourced files are in the `.github/workflow/updates` folder

theofficialgman:
> Why not just make the update_github_script.sh in every app update script?

thats how it used to be
but thats a lot of extra code duplication for many update scripts that will have those commands exactly the same

theofficialgman:
as for the chmoding and for loop stuff, thats already happening in the update_apps.yml

cycool29:
> as for the chmoding and for loop stuff, thats already happening in the update_apps.yml


Yes, so the workflow file can be the `main script` as I described, but it will look really messy (for me).

Crilum:
> > as for the chmoding and for loop stuff, thats already happening in the update_apps.yml
> 
> Yes, so the workflow file can be the `main script` as I described, but it will look really messy (for me).

Why?

cycool29:
> to add the update_github_script.sh as a function in api, but @theofficialgman didn't like it that way so..


No, I don't mean to add it to pi-apps `api`. Instead, we should have another `api` to store extra functions for GitHub Actions.

cycool29:
> Why?

All codes are in blue.

image

But I like colorized code:

image


That's fine, you can leave them as is.

Crilum:
Can you get colored code in YAML? It's blue 'cause it's a sub command in YAML...

Crilum:
> > to add the update_github_script.sh as a function in api, but @theofficialgman didn't like it that way so..
> 
> No, I don't mean to add it to pi-apps `api`. Instead, we should have another `api` (or in the workflow file) to store extra functions for GitHub Actions.

Yeah, I think that would be better..

cycool29:
> Can you get colored code in YAML? It's blue 'cause it's a sub command in YAML...

No, that's why I prefer to put all the code in a separated `main.sh`, and the subcommand in the workflow file will be as simple as:

```yaml
 - name: Run app update scripts
        run: |
          bash .github/workflows/updates/main.sh
```

theofficialgman:
I wouldn't worry about the .yaml file too much
it won't need modification by script writers, its entire job is to load/provide functions, navigate to the app directory, and run any update files

I'm trying my best not to mess with tons of subcommands to avoid variables getting cleared, needing to be exported, etc
lets get the function of the other update scripts in place and then I can edit the main yml later

Crilum:
```
#!/bin/bash
webVer=$(get_release subhra74/xdm)
all_url="https://github.com/subhra74/xdm/releases/download/7.2.11/xdm-setup-${w>
pi_apps_ver="$(cat 'install' | grep 'version=' | sed 's/version=//')"

source update_github_script.sh
```
Is this good for an XDM update script?

theofficialgman:
> source update_github_script.sh

the update script runs in the app folder, so you can't just source the update_github_script.sh file
you have to source its actual path `source $GITHUB_WORKSPACE/.github/workflows/update_github_script.sh`

theofficialgman:
also Xtreme Download Manager doesn't have the `version` variable setup in it yet

Crilum:
Yeah, but we can prepare for it..

Crilum:
XDM install code with version var:

```
#!/bin/bash
version=7.2.11
cd $HOME
mkdir -p xdm 
cd xdm

install_packages openjdk-11-jdk ffmpeg || exit 1

wget -O xdm.tar.xz https://github.com/subhra74/xdm/releases/download/7.2.11/xdm-setup-7.2.11.tar.xz || error "Failed to download XDM setup tar file!"
wget https://github.com/ytdl-org/youtube-dl/releases/download/2021.06.06/youtube-dl || error "Failed to download youtube-dl binary!"
tar -xf xdm.tar.xz || error "Failed to extract XDM setup tar file!"
echo "Executing XDM setup script..."
sudo ./install.sh &>/dev/null || error "Failed to execute XDM setup script!"
sudo chmod +x ./youtube-dl
sudo mv -f ./youtube-dl /opt/xdman || error "Failed to copy youtube-dl binary!"

# Use system java
if [ -d $(dirname $(dirname $(sudo update-alternatives --list java | grep 'java-11-openjdk' --color=none))) ]; then
	sudo rm -rf /opt/xdman/jre/*
	sudo cp -r $(dirname $(dirname $(sudo update-alternatives --list java | grep 'java-11-openjdk' --color=none)))/* /opt/xdman/jre/
elif command -v java &>/dev/null; then
	sudo sed -i s=/opt/xdman/jre/bin/java=$(command -v java)=g /opt/xdman/xdman || error "Failed to change to system Java runtime in /opt/xdman/xdman!"
	sudo sed -i s=/opt/xdman/jre/bin/java=$(command -v java)=g /usr/bin/xdman || error "Failed to change to system Java runtime in /usr/bin/xdman!"
else
	error "OpenJDK-11 installation not found!"
fi
# Fix 'Download failed. Failed to append/convert file parts, please check if the drive is full or write protected'
sudo cp -f $(command -v ffmpeg) /opt/xdman || error "Failed to copy ffmpeg binary!"
# Delete unwanted files after installation
rm -rf $HOME/xdm || error "Failed to remove installation leftovers!"
# Reminder for users to install browser extensiom
echo -e "\nYou may want to install the XDM browser extension. Install it from \e[4mXDM Menu -> Tools -> Browser Monitoring\e[0m page."
```

theofficialgman:
@Crilum no you still don't get what its for
you need to edit the use of the number in the wget command as well
```
wget -O xdm.tar.xz https://github.com/subhra74/xdm/releases/download/${version}/xdm-setup-${version}.tar.xz || error "Failed to download XDM setup tar file!"
```

Crilum:
update script for Angry IP Scanner:

```
#!/bin/bash
version=$(get_release angryip/ipscan)
all_url="https://github.com/angryip/ipscan/releases/download/${version}/xdm-setup-${version}.tar.xz"
pi_apps_ver="$(cat 'install' | grep 'version=' | sed 's/version=//')"

```

Is this better, @theofficialgman

Crilum:
> @Crilum no you still don't get what its for you need to remove the hardcoded number in the wget command as well like this
> 
> ```
> wget -O xdm.tar.xz https://github.com/subhra74/xdm/releases/download/${version}/xdm-setup-${version}.tar.xz || error "Failed to download XDM setup tar file!"
> ```

Sorry, I do realize what the var is for, but I was exited and forgot.. :P

theofficialgman:
> update script for Angry IP Scanner:
> 
> ```
> #!/bin/bash
> version=$(get_release angryip/ipscan)
> all_url="https://github.com/angryip/ipscan/releases/download/${version}/xdm-setup-${version}.tar.xz"
> pi_apps_ver="$(cat 'install' | grep 'version=' | sed 's/version=//')"
> 
> source $GITHUB_WORKSPACE/.github/workflows/update_github_script.sh
> ```
> 
> Is this better, @theofficialgman

yeah thats good
the install itself needs editing as well though

theofficialgman:
wait actually no you messed up there

theofficialgman:
https://github.com/Botspot/pi-apps/blob/609c8f7e3de8bc08eb829c8dd08df736639234c4/.github/workflows/update_github_script.sh
refer to the documentation here
don't go changing the variable names
it has to be webVer otherwise it won't work
(I'll probably add a check for that to skip if the variables aren't present)

Crilum:
@theofficialgman look again.. I fixed it before you commented..

Crilum:
How are we going to deal with two GitHub downloads in the same script? See XDM for an example a script like that..

theofficialgman:
> How are we going to deal with two GitHub downloads in the same script? See XDM for an example of a script like that..

there are a couple of options. for one, XDM doesn't necessarily need for youtube-dl to be updated at the same time as xdm itself.

but the update_github_script.sh script could be made more robust to allow for an arbitrary number of supplied URLS.
from the script writers end, it would probably be something like 
```
all_url=( "https://github.com/subhra74/xdm/releases/download/${webVer[0]}/xdm-setup-${webVer[0]}.tar.xz"  "https://github.com/ytdl-org/youtube-dl/releases/download/${webVer[1]}/youtube-dl" )
```

cycool29:
I transferred the `update_github_script.sh` to function:

```sh
update_github_script() {
    local install_32_webVer
	local install_64_webVer
	local install_webVer
	local app
	
    while [[ "${1}" ]]; do
        case "${1}" in
            --app)
                app=${2}
                shift
                ;;
            --install-webVer)
                install_webVer=${2}
                shift
                ;;
            --install-32-webVer)
                install_32_webVer=${2}
                shift
                ;;
            --install-64-webVer)
                install_32_webVer=${2}
                shift
                ;;
            *)
				error "Unknown parameter: ${1}" 
				;;
        esac
		shift
    done
    
    if [ -n "$install_webVer" ] && [ -a "$DIRECTORY/apps/$app/install" ]; then
		install_currentVer="$(grep "version=" "$DIRECTORY/apps/$app/install" | sed s/.*=//g | xargs)"
		if [ "$install_webVer" != "$install_currentVer" ]; then
			echo "Updating $install_currentVer to $install_webVer in $DIRECTORY/apps/$app/install"
			sed -i "s/$(grep "version=" "$DIRECTORY/apps/$app/install")/version=$install_webVer/" "$DIRECTORY/apps/$app/install"
		else
			echo "$DIRECTORY/apps/$app/install version variable match with the web version variable."
		fi
	elif [ -n "$install_32_webVer" ] && [ -a "$DIRECTORY/apps/$app/install-32" ]; then
		install_32_currentVer="$(grep "version=" "$DIRECTORY/apps/$app/install-32" | sed s/.*=//g | xargs)"
		if [ "$install_32_webVer" != "$install_32_currentVer" ]; then
			echo "Updating $install_32_currentVer to $install_32_webVer in $DIRECTORY/apps/$app/install-32"
			sed -i "s/$(grep "version=" "$DIRECTORY/apps/$app/install-32")/version=$install_32_webVer/" "$DIRECTORY/apps/$app/install-32"
		else
			echo "$DIRECTORY/apps/$app/install-32 version variable match with the web version variable."
		fi
	fi
	
    if [ -n "$install_64_webVer" ] && [ -a "$DIRECTORY/apps/$app/install-64" ]; then
		install_64_currentVer="$(grep "version=" "$DIRECTORY/apps/$app/install-64" | sed s/.*=//g | xargs)"
		if [ "$install_64_webVer" != "$install_64_currentVer" ]; then
			echo "Updating $install_64_currentVer to $install_webVer in $DIRECTORY/apps/$app/install-64"
			sed -i "s/$(grep "version=" "$DIRECTORY/apps/$app/install-64")/version=$install_64_webVer/" "$DIRECTORY/apps/$app/install-64"
		else
			echo "$DIRECTORY/apps/$app/install-64 version variable match with the web version variable."
		fi
	fi
}
```

Crilum:
@cycool29 `install_webVer`, `install_32_webVer`, and `install_64_webVer` are supposed to be: `pi_apps_ver`, `pi_apps_ver_32`, and `pi_apps_ver_64`, I think..

cycool29:
> Pretty sure there's only one `webVer`.

Just to avoid if the 32-bit version value is different from the 64-bit version value.

Crilum:
that's a good idea, but we'll have to change the individual update scripts..

Crilum:
Wouldn't you want something more concrete then `sed s/.*=//g`, in case there are more variables then one? Like `sed 's/version=//'`

cycool29:
> Wouldn't you want something more concrete then `sed s/.*=//g`, in case there are more variables then one? Like `sed 's/version=//'`

Look again... I `grep` before I `sed`:
`grep "version=" "$DIRECTORY/apps/$app/install-64" | sed s/.*=//g`

theofficialgman:
> > Pretty sure there's only one `webVer`.
> 
> Just to avoid if the 32-bit version value is different from the 64-bit version value.

we can't have multiple webVer, that would require a complex github api command
the script already exists if there is no version for armhf/arm64 for a specified webVer

theofficialgman:
the if and elif should not exist like that, that is wrong

theofficialgman:
also `--app "Xtreme Download Manager"` this is redundant info if you plan on executing this function from a file already named like the app, which is necessary since there will inevitably be scripts which need custom update files

I guess we could have both... one master update.sh file which just contains a bunch of `update_script_version` function calls, and then other scripts for apps that need custom update scripts that can't be executed by this function

theofficialgman:
also you removed the validate_url part in your function.... this is vital to avoid pushing broken links to the pi-apps repo

theofficialgman:
overall I'm not a fan of the function version, I don't really see any benefit to it and it just makes doing what I said here harder https://github.com/Botspot/pi-apps/pull/1330#issuecomment-1003236359

cycool29:
> we can't actually populate multiple webVer, that would require a complex github api command to find an old release if the latest release does not support one of the archs
the script already exits if there is no version for armhf/arm64 for a specified webVer

That function is not only for the GitHub release. It can be for Sourceforge, GitLab, or even [packages.debian.org](https://packages.debian.org). What this function does is check if the value same as the provided args. If not, change it.

> the if and elif should not exist like that, that is unnecessary, you are already checking for the install file

Do you mean to put the 64-bit `if` under 32-bit `if`?

> also --app "Xtreme Download Manager" this is redundant info if you plan on executing this function from a file already named like the app, which is necessary since there will inevitably be scripts which need custom update files

Then it could be `--app $(basename "$0" | sed s/.sh//g)`.

> also you removed the validate_url part in your function.... this is vital to avoid pushing broken links to the pi-apps repo
(like if armhf or arm64 support gets removed from a new github release, which has happened multiple times)

That will need to extract the URL from the script... I have no idea how to do it.

theofficialgman:
> That function is not only for the GitHub release. It can be for Sourceforge, GitLab, or even packages.debian.org. What this function does is check if the value same as the provided args. If not, change it.

my code also works on more than github releases, its just what I named the file

> That will need to extract the URL from the script... I have no idea how to do it.

yeah its not feasible to extract the url unless you want to modify all scripts to set their URLs in variables at the top of the script (which could be done). Thats why I supplied the user in the update file

cycool29:
> overall I'm not a fan of the function version, I don't really see any benefit to it and it just makes doing what I said here harder #1330 (comment)

Sure, then you can use the script version if that is more comfortable for you. I just don't want to make the process messy. (Script A trigger script B, script B trigger script C...)

theofficialgman:
> but the update_github_script.sh script could be made more robust to allow for an arbitrary number of supplied URLS.

this is the main concern with using a function based approach, I do intend to implement the above. passing in all the variables through `--varname varvale` will be a bit of a nightmare to handle in bash

Crilum:
Nice!

theofficialgman:
there is a bug
I forgot to rename `variable` to `variable2` in the script, 😆

theofficialgman:
alright there we go
that will work assuming the user names their version variables like so:
```
version=
version2=
version3=
```
I could have an override as another input with an array of values incase those default variable names are not desired:
something like
`variable_names=( version_xdm version_ytdl )`

Crilum:
Those are fine. :+1:

theofficialgman:
no changes, just rewrote the commit history so it was compartmentalized and easier to follow

theofficialgman:
ok at this point I've got all the low hanging fruit.
the biggest app that downloads files that could be still use the github actions update mechanism is VSCodium but that will require a custom update script because its filename changes each release

.github/workflows/update_apps.yml
```diff
@@ -0,0 +1,79 @@
...
+          DIRECTORY=$GITHUB_WORKSPACE
+          source $GITHUB_WORKSPACE/api
+
+          #add special functions
+          get_release() {
+            curl --silent "https://api.github.com/repos/$1/releases/latest" | jq -r '.tag_name' | sed s/v//g
+          }
+
+          function validate_url(){
+            if [[ `wget -S --spider $1  2>&1 | grep 'HTTP/1.1 200 OK'` ]]; then
```
Itai-Nelken, line 36:
```suggestion
            if [[ $(wget -S --spider $1  2>&1 | grep 'HTTP/1.1 200 OK'$) ]]; then
```

cycool29, line 36:
```suggestion
            if [ "$(curl -s -o /dev/null -I -w "%{http_code}" $1)" == "200"  ]; then
```

Smarter way.

theofficialgman: using curl changes the http response codes, its no longer 200

I think the function is overkill anyway, this should work
```
if wget -q --spider "$1"; then
```

Crilum:
When/if DDnet and Droidcam get added, those use GitHub to download stuff, so we'll need update scripts, and the version var in each install script..

theofficialgman:
I skipped over a lot of scripts that technically could be added. Some haven't had github updates in many months, and others just have a really small userbase in pi-apps.
I figured the original script writers can just make PRs in the future to support other apps

theofficialgman:
@Itai-Nelken changed so that it makes a PR instead of pushing to the repo
see my test run of it here: https://github.com/theofficialgman/pi-apps/runs/4744773460?check_suite_focus=true

and the PR it made is here: https://github.com/theofficialgman/pi-apps/pull/1

(still need to change the name so its more useful for us)

theofficialgman:
marking this script as ready for review

script writers and pi-apps maintainers are free to write as complicated update checking scripts as they wish, just make sure to not write files inside the $GITHUB_WORKSPACE directory unless those files/changes are intended to be added to the automated PR

the full ubuntu container is available, you should be able to do apt installs, do test builds, etc if desired

.github/workflows/update_apps.yml
```diff
@@ -0,0 +1,86 @@
...
+            Update App Versions, run by GitHub Actions - Apps updated:
+
+            ${{ env.UPDATED_APPS }}
+          branch: "auto-app-updates"
+          title: "[AUTO] Update App Versions"
+          body: |
+            Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action:
+            Apps updated:
+
+            ${{ env.UPDATED_APPS }}
```
Itai-Nelken, line 86: would it be possible to make this a list?
for example:
```markdown
* app1
* app2
```

theofficialgman: yes, I actually store it as a list in a file, then print it as one line, that can be changed

theofficialgman: done with https://github.com/Botspot/pi-apps/pull/1330/commits/a30cf499989c797383d7da2da112c0ad2208a866

Itai-Nelken (approved):
Looks good to me.
What do you think @Botspot?

theofficialgman:
removed the delete branch after merging commit... it doesn't seem to do anything so no purpose in adding it
the delete button exists right after clicking merge, just use that to remove the branch if you don't want it in the github

theofficialgman:
@Botspot is there anything holding you back from merging this?

theofficialgman:
alright cool, now we just wait around till 21:00 UTC time or run it manually to see what it updates

Itai-Nelken:
I have manually run it.

PR #1338, opened 2022-01-02
BUG: imagemagick convert on buster/stretch/bionic

theofficialgman:
addresses issue raised on discord about bug in older versions of imagemagick convert function where png is corrupted and unusable (see below). librsvg2-bin is less than 50KB in size so its not taking up any measurable space

edit: this works on ubuntu bionic but seems to hang on raspbian buster, see below

Screenshot_from_2022-01-01_23-31-54

### DIFF
--- a/api
+++ b/api
@@ -74,7 +74,7 @@ generate_splashscreen() { #Place 10 random app icons on the loading screen for t
   s_IMAGE8_$(base64 "${DIRECTORY}/apps/${app9}/icon-64.png" -w 0)_g ; \
   s_IMAGE9_$(base64 "${DIRECTORY}/apps/${app10}/icon-64.png" -w 0)_g" "${DIRECTORY}/icons/vector/splashscreen-original.svg" > "${DIRECTORY}/icons/vector/splashscreen.svg"
   
-  convert +antialias "${DIRECTORY}/icons/vector/splashscreen.svg" "${DIRECTORY}/icons/splashscreen.png"
+  rsvg-convert "${DIRECTORY}/icons/vector/splashscreen.svg" > "${DIRECTORY}/icons/splashscreen.png"
   
 }
 #end of output functions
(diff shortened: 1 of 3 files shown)


theofficialgman:
tested with raspbian buster imagemagick and this bug exists here in addition to ubuntu bionic

so this should definitely be changed

theofficialgman:
whelp... tested `rsvg-convert` on debian buster in my chroot and than hangs...

note, imagemagick `convert` outputs errors on raspbian when I was tested the conversion
```
convert splashscreen-original.svg splashscreen-original.png
convert-im6.q16: corrupt image `' @ error/constitute.c/ReadInlineImage/997.
convert-im6.q16: non-conforming drawing primitive definition `image' @ error/draw.c/RenderMVGContent/4301.
```
maybe the splashscreen-original.svg you supplied is messed up in some way?

theofficialgman:
I've tried inkscape as well on buster/bionic and that doesn't work great either (a lot of transparencies are missing)

Botspot:
> maybe the splashscreen-original.svg you supplied is messed up in some way? because this test image https://upload.wikimedia.org/wikipedia/commons/b/bd/Test.svg converts just fine in both rsvg-convert and convert

`splashscreen-original.svg` is not intended to be rendered. It's an invalid svg file, only meant to serve as a template for `splashscreen.svg`. Try converting `splashscreen.svg`

theofficialgman:
I left out that info but I converted both the generated splashscreen.svg (which is what I have pictured above) and splash screen-original.svg
Same issue with corrupted png output with convert (on buster/stretch/bionic) and a hang on rsvg convert on buster

Botspot:
Commit https://github.com/Botspot/pi-apps/commit/b7975c0ca6721345a576d5af009c552df4bf7b8e should solve everything and lets yad use the svg if the png is unavailable.

PR #1340, opened 2022-01-02
Updated description

derryktutt:
There were multiple grammar problems, like Vivaldi is a proper noun. It should be capitalized.


Itai-Nelken:
Thanks!

PR #1342, opened 2022-01-03
Use `rsvg-convert` to convert SVG to PNG

cycool29:
Since `yad` supports SVG image, we can simply show SVG splashscreen, without convert it into PNG.

### DIFF
--- a/api
+++ b/api
@@ -74,7 +74,7 @@ generate_splashscreen() { #Place 10 random app icons on the loading screen for t
   s_IMAGE8_$(base64 "${DIRECTORY}/apps/${app9}/icon-64.png" -w 0)_g ; \
   s_IMAGE9_$(base64 "${DIRECTORY}/apps/${app10}/icon-64.png" -w 0)_g" "${DIRECTORY}/icons/vector/splashscreen-original.svg" > "${DIRECTORY}/icons/vector/splashscreen.svg"
   
-  convert +antialias "${DIRECTORY}/icons/vector/splashscreen.svg" "${DIRECTORY}/icons/splashscreen.png"
+  rsvg-convert  "${DIRECTORY}/icons/vector/splashscreen.svg" > "${DIRECTORY}/icons/splashscreen.png"
   
 }
 #end of output functions


Botspot:
SVG images have to be converted to raster format before they can be viewed. This takes time.

On my system, yad started noticeably slower when loading the svg than when loading the png. Since we want the splash screen to show as soon as possible, it makes sense to convert the image ahead of time. But, if we can't find a way to reliably do that, this will definitely be an option, though maybe as a last resort.

cycool29:
> On my system, yad started noticeably slower when loading the svg than when loading the png.

Hmm... that is just very minor changes.

With SVG:

```
real	0m0.985s
user	0m0.725s
sys	0m0.412s
```

With PNG:
```
real	0m0.803s
user	0m0.617s
sys	0m0.522s
```

That is just 0.2 gap.

Botspot:
> That is just 0.2 gap.

How did you time yad? For the command to stop, you have to close the window. Did you simply try closing yad as fast as you could?

cycool29:
I just measure the time when splashscreen shows.

Just add a `exit` in the gui script, where pi-apps kills splashscreen.

Botspot:
I'm not sure you are measuring the time correctly.
We're not interested in how long it takes Pi-Apps to load. We want to know how long it takes for the splash screen to display on-screen. We want to measure the time between when yad is launched and when it displays on-screen.
I'm not aware of a way to do that with the `time` command, so I simply compared the launch-times for png and svg. To the naked eye, svg was noticeably slower.

cycool29:
> How did you time yad? For the command to stop, you have to close the window. Did you simply try closing yad as fast as you could?

I added a exit in gui script where it kill the splashscreen. Now I just realised the method is not accurate...

However, I can't see big speed changes, with my eyes.

Since you say it is noticeably slower, let's just leave it as is.

Botspot:
> Since you say it is noticeably slower, let's just leave it as is.

That's assuming we find a way to reliably generate the png on various systems. If a solution is not found, we will be forced to make yad display the svg.

cycool29:
I changed the title.

Using `convert` to convert SVG to PNG is unreliable, as the SVG delegates may different on everyone's system.

The `convert` command convert the SVG splashscreen into a weird PNG:

image

Botspot:
The rsvg-convert command is installed by the `librsvg2-bin` package so I'll make it a dependency.

theofficialgman:
guys are already said this in my PR @Botspot @cycool29 rsvg-convert was hanging for me on my raspbian buster chroot
https://github.com/Botspot/pi-apps/pull/1338

theofficialgman:
also my PR had all the changes necessary (like adding the dependency)
I just made it a draft since I needed someone to confirm the hang or not on raspbian buster on an actual pi

Botspot:
Oh I'm sorry. I'll spin up a buster vdesktop in a few mins.

Botspot:
Splash screen kinda worked.
splashscreen
The blue glow doesn't look quite right, though at least it works.

Botspot:
Commit https://github.com/Botspot/pi-apps/commit/b7975c0ca6721345a576d5af009c552df4bf7b8e should solve everything and lets yad use the svg if the png is unavailable.

theofficialgman:
> Commit [b7975c0](https://github.com/Botspot/pi-apps/commit/b7975c0ca6721345a576d5af009c552df4bf7b8e) should solve everything and lets yad use the svg if the png is unavailable.

that won't actually fix what I said though, in my chroot, rsvg-convert never exited, it kept running and hogged all cpu until I forge killed it. thats a good fallback if it does fail though

Botspot:
> that won't actually fix what I said though, in my chroot, rsvg-convert never exited, it kept running and hogged all cpu until I force killed it. thats a good fallback if it does fail though

If rsvg-convert truly hangs on people's systems, then by all means we will remove it! However, I just tried running the latest Pi-Apps in a buster vdesktop and rsvg-convert executed fine. (So I'm hoping that hanging issue was just a problem with your chroot)
If anybody comes to us and reports rsvg-convert hanging, then I'll probably be forced to make yad render the svg directly.

PR #1345, opened 2022-01-04
Exclude unneeded file and directory in `package_available`

cycool29:


Botspot:
I presume you came across a system where `/var/lib/apt/lists` was unreadable to the unprivileged user?

cycool29:
I just realized there are only two file/directory needs root privilege:

```
pi@raspberrypi:~ $ grep -rq "^Package: $package"'$' /var/lib/apt/lists
grep: /var/lib/apt/lists/lock: Permission denied
grep: /var/lib/apt/lists/partial: Permission denied
```

And these file/directory are not needed by `package_available`, so I excluded them in the `grep` command.

Botspot:
> I just realized there are only two file/directory needs root privilege:

Okay, so you excluded a directory that would throw an error. But why does it matter, as I've already suppressed any errors from `grep` ? Did it exit with an undesirable exit code or something?

cycool29:
> > I just realized there are only two file/directory needs root privilege:
> 
> Okay, so you excluded a directory that would throw an error. But why does it matter, as I've already suppressed any errors from `grep` ? Did it exit with an undesirable exit code or something?

Yes, since it throws `Permission denied` error, it simply returns error code 2 in every search.

PR #1347, opened 2022-01-04
Speedtest update to 1.1.1

theofficialgman:


theofficialgman:
@Botspot  I've set this as a draft to show that the etc/import apps function needs updating. Since I changed this app from a install-32, install-64 app to an install only app, the importer does not find any updates in this PR

however, the updater does still find this correctly and update using it so thats good

theofficialgman:
squashed commits and marking as ready for review

PR #1350, opened 2022-01-05
Use ASCII version of `Pi-Apps` text in `generate_logo`

cycool29:
Result:

### DIFF
--- a/api
+++ b/api
@@ -32,20 +32,17 @@ status_green() { #announce the success of a major action
 generate_logo() { #display colorized Pi-Apps logo in terminal
   #generate pi-apps logo
   #https://misc.flogisoft.com/bash/tip_colors_and_formatting
-  blue_text='\e[38;5;21m' #blue='\e[38;5;27m' #lighter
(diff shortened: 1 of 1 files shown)


Botspot:
I personally find outline-based text harder to read. Perhaps it would be okay if the text was filled in, though that would require a few more hours spent on https://unicode-search.net.

open-sorcerer64:
> How about this?
> 
> image

That looks amazing imo

Botspot:
How about this?
2022-01-10-214515_1920x1080_scrot

Crilum:
@Botspot I love it!

Crilum:
> > Perhaps it would be okay if the text was filled in, though that would require a few more hours spent on https://unicode-search.net.
> 
> Well that's what I ended up doing.
> 
> How is it? 2022-01-10-214515_1920x1080_scrot

Looks way cooler than this:
image

cycool29:
> Well that's what I ended up doing.
> 
> How is it? 2022-01-10-214515_1920x1080_scrot

Cool!

PR #1352, opened 2022-01-05
Temporary fix of `error while loading shared libraries: libxcb-util.so.0: cannot open shared object file: No such file or directory`

### LINKED ISSUE #1318: OBS Studio does not run
Hi Team,

I have installed OBS studio on a Pi 4 4GB RAM but when i run it nothing happens. the program does not run.

Is there anything else i need to do. 

thanks 

here is the rror i get when trying to run from CLI

pi@raspberrypi:~ $ obs
/usr/bin/obs-exec: error while loading shared libraries: libxcb-util.so.0: cannot open shared object file: No such file or directory

cycool29:
To close #1318.

### DIFF
--- a/apps/OBS Studio/install-32
+++ b/apps/OBS Studio/install-32
@@ -15,3 +15,6 @@ install_packages http://ftp.debian.org/debian/pool/non-free/f/fdk-aac/libfdk-aac
 echo '#!/bin/bash
 MESA_GL_VERSION_OVERRIDE=3.3 /usr/bin/obs-exec "$@"' | sudo tee /usr/bin/obs >/dev/null
 sudo chmod +x /usr/bin/obs
+
+#temporary fix of `error while loading shared libraries: libxcb-util.so.0: cannot open shared object file: No such file or directory`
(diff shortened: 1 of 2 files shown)


apps/OBS Studio/install-32
```diff
@@ -15,3 +15,6 @@ install_packages http://ftp.debian.org/debian/pool/non-free/f/fdk-aac/libfdk-aac
 echo '#!/bin/bash
 MESA_GL_VERSION_OVERRIDE=3.3 /usr/bin/obs-exec "$@"' | sudo tee /usr/bin/obs >/dev/null
 sudo chmod +x /usr/bin/obs
+
+#temporary fix of `error while loading shared libraries: libxcb-util.so.0: cannot open shared object file: No such file or directory`
+! dpkg -S libxcb-util.so.0 2>/dev/null && ! [ -a ""$(dirname "$(dpkg -S libxcb-util.so.1 | grep "libxcb-util.so.1$" | awk '{printf $2}')")/libxcb-util.so.0"" ] && sudo ln -sf "$(dpkg -S libxcb-util.so.1 | grep "libxcb-util.so.1$" | awk '{printf $2}')" "$(dirname "$(dpkg -S libxcb-util.so.1 | grep "libxcb-util.so.1$" | awk '{printf $2}')")/libxcb-util.so.0"
```
Botspot, line 20: Why the -double-double quotes? (`""`)

Botspot:
I've reworked your implementation to run faster, display a status, be more readable, and to avoid leaving the script with a non-zero exit code.

PR #1354, opened 2022-01-06
`package_installed`: match only `$package` at the end of line

cycool29:
For example, `gean` package doesn't exists, but `package_available` still returns 0 as it matched `geany`.

### DIFF
--- a/api
+++ b/api
@@ -112,7 +112,7 @@ package_installed() { #exit 0 if $1 package is installed, otherwise exit 1
   #package_info "$package" | grep -q '^Status: install ok installed$'
   
   #directly search /var/lib/dpkg/status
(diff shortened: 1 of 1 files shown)


Botspot:
Is this tested?

cycool29:
> Is this tested?

Yes. 


Is there any reason you are not using `package_info` to grep `install ok installed`?

Botspot:
> Is there any reason you are not using `package_info` to grep `install ok installed`? I saw you commented out the code.

The execution time was a bit slower.

cycool29:
> The execution time was a bit slower.

OK, then this PR is good to merge.

Botspot:
If regex was disabled, it would probably be even faster.
Like this:
`grep -qxF "Package: $package"`
I'll make that change myself.

PR #1358, opened 2022-01-07
etc/import-app function update (fixes updates which change app type)

theofficialgman:
instead of diffing the contents of the folders when checking for changes in an app, do a diff of the folder itself. fixes issue brought up in https://github.com/Botspot/pi-apps/pull/1347

### DIFF
--- a/etc/import-app
+++ b/etc/import-app
@@ -179,14 +179,31 @@ elif [[ "$import" =~ ^[0-9]+$ ]] || [[ "$import" == *'://'*'/pull/'*[0-9] ]];the
   echo "Done"
   
   #get list of apps that are unique to this pi-apps folder. (not on main repo)
-  #apps="$("$(pwd)/pi-apps/api" list_apps local_only)"
+  applist="$("$(pwd)/pi-apps/api" list_apps local_only)"
   
   #get apps that don't match the ones in main branch - find only new apps
   imported_apps="$(diff -rq "$(pwd)/pi-apps/apps" "$(pwd)/pi-apps-master/apps" | grep "Only in $(pwd)/pi-apps/apps:"  | sed 's+.*/pi-apps-master.++g' | sed 's+.*apps/++g' | sed 's+.*apps: ++g'  | sed 's+/.*++g' | sort | uniq | grep .)"
-  #if no new apps, find modified apps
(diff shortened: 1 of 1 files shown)


theofficialgman:
previously marked as draft while I checked that the updater would correctly handle changing from install-32/64 to install, it does because like this PR, it removes the old app directory before copying over the new one

etc/import-app
```diff
@@ -179,21 +179,41 @@ elif [[ "$import" =~ ^[0-9]+$ ]] || [[ "$import" == *'://'*'/pull/'*[0-9] ]];the
...
   imported_apps="$(diff -rq "$(pwd)/pi-apps/apps" "$(pwd)/pi-apps-master/apps" | grep "Only in $(pwd)/pi-apps/apps:"  | sed 's+.*/pi-apps-master.++g' | sed 's+.*apps/++g' | sed 's+.*apps: ++g'  | sed 's+/.*++g' | sort | uniq | grep .)"
-  #if no new apps, find modified apps
-  [ -z "$imported_apps" ] && imported_apps="$(diff -rq "$(pwd)/pi-apps/apps" "$(pwd)/pi-apps-master/apps" | grep "Files .* and .* differ"  | sed 's+.*/pi-apps-master.++g' | sed 's+.*apps/++g' | sed 's+.*apps: ++g'  | sed 's+/.*++g' | sort | uniq | grep .)"
-  
-  echo -e "\nModified apps in this branch:"
+
+  IFS=$'\n'
+  for app in $applist
+  do
+    echo -en "Scanning apps... $app\033[0K\r" 1>&2
```
Itai-Nelken, line 190: I don't think that `\033[OK` is a valid escape code.

cycool29: That is from the `manage` script: https://github.com/Botspot/pi-apps/blob/779aae10235ca41317ce874a13e407ccc7aeb57b/manage#L441

Botspot: > I don't think that `\033[OK` is a valid escape code.

It is. That special escape code is used to clear any characters that are to the right of the cursor. This is important because some app-names are longer than others. If the new app-name doesn't clear the last one, we may get output like:
```
Scanning apps... Discordft Java MultiMC5
```

PR #1364, opened 2022-01-09
Added Nemo

open-sorcerer64:


apps/Nemo/description
```diff
@@ -0,0 +1,5 @@
+File Manager with loads of features
+
```
cycool29, line 2: From `sudo apt info nemo`.

```suggestion
File manager and graphical shell for Cinnamon, also works well on other desktop environments.

Nemo is the official file manager for the Cinnamon desktop. It allows one
to browse directories, preview files, and launch applications associated
with them. It is also responsible for handling the icons on the Cinnamon
desktop. It works on local and remote filesystems.
Several icon themes and components for viewing different kinds of files
are available in separate packages.

```

Crilum: I think this would be a better description ^.. I looked at the current one ("File Manager with loads of features"), and it doesn't really describe much about the app..

Botspot: > I think this would be a better description ^.. I looked at the current one ("File Manager with loads of features"), and it doesn't really describe much about the app..

Thank you for catching this, @Crilum. I'm not sure why this conversation was resolved without any action being taken, but I've replaced the short description with @cycool29's more detailed one in commit https://github.com/Botspot/pi-apps/commit/deb5da89a362f3f75384a68cafb8165102ec94ac.

PR #1382, opened 2022-01-16
Vivaldi proprietary media fix

open-sorcerer64:
This fixed the HTML proprietary media(H.264 and AAC) issue for me, not sure if this issue exists for everyone else.

**REQUIRES TESTING**


open-sorcerer64:
Should I add this `deb` to my personal repo???

apps/Vivaldi/install
```diff
@@ -2,7 +2,8 @@
 #https://help.vivaldi.com/desktop/install-update/manual-setup-vivaldi-linux-repositories/
 wget -qO- https://repo.vivaldi.com/archive/linux_signing_key.pub | gpg --dearmor | sudo tee /usr/share/keyrings/vivaldi-browser.gpg >/dev/null || exit 1
 echo "deb [signed-by=/usr/share/keyrings/vivaldi-browser.gpg arch=$(dpkg --print-architecture)] https://repo.vivaldi.com/archive/deb/ stable main" | sudo tee /etc/apt/sources.list.d/vivaldi-archive.list >/dev/null
-
+wget -O- https://launchpadlibrarian.net/574476800/chromium-codecs-ffmpeg-extra_97.0.4692.56-0ubuntu0.18.04.1_armhf.deb | tail -c+1153 | tar JOx ./usr/lib/chromium-browser/libffmpeg.so > libffmpeg.so.5.0
```
Botspot, line 5: I've never seen this approach before. Downloading a deb package, cutting it, and extracting a file from it, all simultaneously fascinates me.
How did you find this, and how does it work?

open-sorcerer64:
Vivaldi forums

open-sorcerer64:
Merge???

apps/Vivaldi/install
Itai-Nelken: I'm not sure I like the `tail` part much. it assumes the needed file starts at the 1153 byte and ends at the end of the file, and that might change. I think that a better more future proof way to do it would be to extract the deb and copy the file from there.

open-sorcerer64: So can you do it??

theofficialgman:
@open-sorcerer64 definitely do not merge. just tested downloading that deb and it doesn't exist anymore at that location on the server
probably was a temporary file

open-sorcerer64:
> @open-sorcerer64 definitely not mergeable. just tested downloading that deb and it doesn't exist anymore at that location on the server probably was a temporary file, or got updated at the launchpad and the previous version removed

That's bad, I'll see what I can do

ryanfortner:
@open-sorcerer64 your script is only handling the armhf version of `chromium-codecs-ffmpeg-extra`. You'll need to split the `install` script into separate install-32 and install-64.

theofficialgman:
see https://github.com/Botspot/pi-apps/pull/1877#issuecomment-1137431914

PR #1390, opened 2022-01-17
Stunt Rally: use 32-bit deb

ryanfortner:
This was requested by @Itai-Nelken, and it already includes the desktop entry fixes. I'll open another PR later for a 64-bit deb.


Itai-Nelken:
Thanks!

PR #1396, opened 2022-01-18
Rbcoder117 add powershell

### LINKED ISSUE #1392: Microsoft Powershell
### What is the name of the app?

Microsoft Powershell

### (Optional) Where is the app hosted?

Microsoft.com

### About the app

To Deliver a terminal with easy tools that are usually already known.

### Upload file

[Microsoft Powershell.zip](https://github.com/Botspot/pi-apps/files/7884512/Microsoft.Powershell.zip)

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.
- [x] I have confirmed that this app can run on Raspberry Pi.

rbcoder117:
requested in issue  and also to close #1392 
@Botspot is it ready?


theofficialgman:
I wouldn't merge yet
it would be nice to have auto updating of the version included in this PR so it doesn't have to be done later

create this file `.github/workflows/updates/Microsoft PowerShell.sh`

and put in it:
```
#!/bin/bash

webVer=$(get_release PowerShell/PowerShell)
all_url="https://github.com/PowerShell/PowerShell/releases/download/v${webVer}/powershell-${webVer}-linux-arm32.tar.gz "

```

we might want to wait though still because I need to update the update_github_script.sh script so that it can accommodate supplying and armhf and arm64 url while still using the 'install' script

rbcoder117:
ok

theofficialgman:
the above PR solves the concern, I've also updated my above post with the contents of the file that should be added

PR #1406, opened 2022-01-20
Adds everything for mc server

rbcoder117:


apps/Minecraft Java Server/install
```diff
@@ -88,8 +103,8 @@ description="You can now close your browser.\
 \nIn order to run Minecraft, the game requires a specific version of Java.\
 \nPlease choose below which version of java the server of your choice should run with.\
 \nJava versions can be 8, 16, or 17.\
-\n\nAll version of Minecraft before and including Minecraft 1.16.5 should use Java 8.\
-\nMinecraft 1.17.X should use java 16 and Minecraft 1.18+ should use java 17."
+\n\nAll version of minecraft before and including minecraft 1.16.5 should use Java 8.\
+\nminecraft 1.17.X should use java 16 and minecraft 1.18+ should use java 17."
```
Botspot, line 107: Why are you changing "Minecraft" to lowercase?

Botspot:
Have you tested this? I'm pretty sure these three new options would not show up in the dialog, thereby making them unreachable.

apps/Minecraft Java Server/install
```diff
@@ -66,6 +66,21 @@ case "$server_type" in
     setsid x-www-browser "https://files.minecraftforge.net/net/minecraftforge/forge/" > /dev/null 2>&1 &
     server_jar="forge-*-installer.jar"
     ;;
+    "Paper")
+    # open browser to download the server version of user choice
+    setsid x-www-browser "https://papermc.io/downloads/" > /dev/null 2>&1 &
```
theofficialgman, line 71: remove the `/` after downloads or else you get a 404

rbcoder117: do you mind fixing it I am busy right now

theofficialgman: I don't have permissions on this repo to fix myself

rbcoder117: @theofficialgman I invited you to be a collaborator. I'm still busy

KolpixTV:
Nice, but it would be nice if you could add `[Purpur](https://purpurmc.org/downloads)` 

`server_jar="purpur-*.jar"`

But i think it's wayyy better if you could directly install Paper, Purpur etc. directly over the App like:

1. You Press the Server Software.
2. You choose the Version.
- If you want a custom Version you can add it to the Folder.

So you don't have to install it manually. All Server Versions for Paper, Purpur etc. are [here](https://serverjars.com/) (You can find there also older Versions of Paper, Purpur etc. on the Papermc Website you can only find the new Versions).

rbcoder117:
@KolpixTV ok I will fix

theofficialgman:
> But i think it's wayyy better if you could directly install Paper, Purpur etc. directly over the App like:
> 

> So you don't have to install it manually. All Server Versions for Paper, Purpur etc. are [here](https://serverjars.com/) (You can find there also older Versions of Paper, Purpur etc. on the Papermc Website you can only find the new Versions).

No. The BIG problem with this is not all of the server software is so easy to download, things like paper, spigot, and craftbukkit release new versions daily with bugfixes, and various formatting of the strings used as filenames. Many of the sites make money off of ad revenue and don't let you download with wget commands (including the one you linked). We would also need to maintain a list of ALL versions of the server software supported by EACH AND EVERY server platform (paper, purpur, fabric, forge, etc) which would be a total mess and way over the top.

Instead, how the script is written now, the user takes responsibility and understands where the server software they download comes from, they can choose exactly which version they want, and we don't have to maintain any sort of list of "supported server versions". Its a win for all as it is

theofficialgman:
> But i think it's wayyy better if you could directly install Paper, Purpur etc. directly over the App like:
> 
> 1. You Press the Server Software.
> 2. You choose the Version.
> 
> * If you want a custom Version you can add it to the Folder.
> 
> So you don't have to install it manually. All Server Versions for Paper, Purpur etc. are [here](https://serverjars.com/) (You can find there also older Versions of Paper, Purpur etc. on the Papermc Website you can only find the new Versions).

@KolpixTV maybe I said no to your suggestion too quickly... the website you linked has its own API that we could use to actually get all supported versions of minecraft and updated download links for Bukkit, Paper, Spigot, Purpur, Tuinity, and Vanilla. this could be used to automatically generate a YAD dialog with the list of supported versions which would be auto-downloaded....

example output can be seen by running:
```
curl https://serverjars.com/api/fetchAll/vanilla/0
```

we would be missing fabric and forge though... those would have to fallback to the old system. I'll consider it

theofficialgman:
@KolpixTV alright, here is your dream PR: https://github.com/Botspot/pi-apps/pull/1444
to close this PR

Issue #1387, opened 2022-01-17
Adding DietPi support


pklion:
### Cool ideas?

It looks like there is no official support for DietPi, so it's better if there is a name of DietPi on the supported OS list.

cycool29:
Since DietPi is Debian-based, pi-apps should be compatible with it.

Instead of specifying every compatible Debian-based distro on README, I think we can simply say `Debian-based ARM OS`.

pklion:
I think it's good idea.

PR #1408, opened 2022-01-20
Update README.md

ShadityZ:
Just a little README update.    ;)
- [ShadityOS](https://github.com/ShadityOS) is real (I develop it myself)


Botspot:
If we included every operating system in the compatibility list, the list would be too long.
Just a few days ago, [somebody asked us](https://github.com/Botspot/pi-apps/issues/1387) to add DietPi to the list, and we decided not to, for the same reason.

With that being said, can you can think of any reason to still put this OS on the README, while leaving out DietPi?

ShadityZ:
Nah, I get it. ShadityOS is based off Raspbian anyway i guess.

Botspot:
I've actually decided to remove AmogOS from the list as well, for consistency. ;)

PR #1409, opened 2022-01-20
Added Starship Prompt

open-sorcerer64:
Needs help in uninstall script


apps/Starship Prompt/install
```diff
@@ -0,0 +1,18 @@
...
+mv *.ttf ~/.local/share/fonts/fira-code || error 'Failed to move fonts'
+mv *.otf ~/.local/share/fonts/fira-code || error 'Failed to move fonts'
+
+echo fontname >> ~/.config/lxterminal/lxterminal.conf
+sed -i '/fontname/c\fontname=Fira Code Medium Nerd Font 11' ~/.config/lxterminal/lxterminal.conf
+echo FontName >> ~/.config/xfce4/terminal/terminalrc
+sed -i '/FontName/c\FontName=Fira Code Medium Nerd Font 12' ~/.config/xfce4/terminal/terminalrc
+
+sh -c "$(curl -fsSL https://starship.rs/install.sh)" || error 'Failed to download starship prompt'
+echo  eval "$(starship init bash)" >> ~/.bashrc || error 'Failed to set starship prompt'
```
theofficialgman, line 18: this does not work. did you test it? the eval attempts to run here, you want to place this at the end of the .bashrc, so you need to put it in ' '

```
echo 'eval "$(starship init bash)"' >> >> ~/.bashrc || error 'Failed to set starship prompt'
```
also this is dangerous if run twice, you should be removing all instances of this from .bashrc before adding it once

apps/Starship Prompt/uninstall
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+sh -c 'rm "$(command -v 'starship')"' || error 'Failed to remove starship'
+
+#HOW CAN I REMOVE TEXT FROM ~/.bashrc????
```
theofficialgman, line 5: in general, I would suggest not adding anything to .bashrc, but I see no way around it in this case

you can remove matching text like this
```
sed -i '/\eval "$(starship init bash)"/d' "$HOME/.bashrc"
```

apps/Starship Prompt/install
theofficialgman: see below post, you should do this when adding the line:

```
# remove all occurances
sed -i '/\eval "$(starship init bash)"/d' "$HOME/.bashrc"
# add one occurance to the .bashrc file
echo 'eval "$(starship init bash)"' >> ~/.bashrc || error 'Failed to set starship prompt'
```

Itai-Nelken:
Using `.bashrc` isn't the best idea as not everyone uses bash.
I don't think it's a big problem yet, as people that use pi-apps most likely haven't changed their shell, but it might be a good idea to make an `api` function called `get_shell_rc_file_path` that gets the currently running shell's rc file.

open-sorcerer64:
Yeah, it would be good idea as I too use `zsh` primarily

theofficialgman:
> Using `.bashrc` isn't the best idea as not everyone uses bash. I don't think it's a big problem yet, as people that use pi-apps most likely haven't changed their shell, but it might be a good idea to make an `api` function called `get_shell_rc_file_path` that gets the currently running shell's rc file.

alternatively, just add it to all rc paths that are detected from a list. that way if zsh is installed at all, it will be added there

open-sorcerer64:
Any ideas on how we can grab rc paths

open-sorcerer64:
Uh ho

theofficialgman:
@open-sorcerer64 I assume closing this was an accident?
I'll re-open

open-sorcerer64:
> @open-sorcerer64 I assume closing this was an accident? I'll re-open

Yeah I actually messed up my fork ): thanks

ryanfortner:
This isn't entirely necessary, but maybe we should consider using deb files for starship to make it easier to update for users.

I set up a github action on this repo: https://github.com/ryanfortner/starship-arm, but of course you can also download them from [my apt repo](https://apt.raspbian-addons.org/debian/pool/main/s/starship/)

open-sorcerer64:
Continued in #1616

PR #1430, opened 2022-01-23
update obs-studio to v27.2.0

ryanfortner:


apps/OBS Studio/install-32
```diff
@@ -1,14 +1,14 @@
...
 
-#remove obs-studio package first
-if package_installed obs-studio ;then
-  echo "Removing obs-studio package first..."
+#remove obs package first
+if package_installed obs ;then
+  echo "Removing obs package first..."
   apt_lock_wait
-  sudo apt purge -y --autoremove obs-studio
+  sudo apt purge -y --autoremove obs
```
Botspot, line 7: It's necessary to check for and remove the `obs-studio` package first, because some people already have it installed from the APT repositories. From there, the package is named `obs-studio`:
```
$ apt search --names-only obs
obs-studio/stable 26.1.2+dfsg1-2 armhf
  recorder and streamer for live video content
```

ryanfortner: Should I also have the install scripts purge the `obs` package, because the new debs use obs-studio as the package name?

Botspot: > Should I also have the install scripts purge the `obs` package, because the debs in this PR use obs-studio as the package name?

You can if you want to.

ryanfortner: Done.

PR #1435, opened 2022-01-24
Fix MultiMC install for ubuntu derivatives

isteiger:
On elementary OS for RPi, it outputted that it seems to be too old to run/compile MultiMC, but it is based on Ubuntu 20.04.  I patched the version check to include $VERSION_ID, which is what the version ID is under for the upstream release lsb_release file. I have no idea if it is only elementary OS that does it like this, but this should fix it if it is different, and should not affect it if it is not (hopefully, I haven't had a chance to test it yet).

### DIFF
--- a/apps/Minecraft Java MultiMC5/install
+++ b/apps/Minecraft Java MultiMC5/install
@@ -139,7 +139,7 @@ case "$__os_id" in
                 ;;
             *)
                 requiredver="18.04"
-                if printf '%s\n' "$requiredver" "$DISTRIB_RELEASE" | sort -CV; then
+                if [printf '%s\n' "$requiredver" "$DISTRIB_RELEASE" | sort -CV] || [printf '%s\n' "$requiredver" "$VERSION_ID" | sort -CV]; then
                     status "Skipping OpenJDK PPA, $DISTRIB_CODENAME already has openjdk-16 in the default repositories"
                 else
                     error "$DISTRIB_CODENAME appears to be too old to run/compile MultiMC5"


theofficialgman:
do NOT merge
@isteiger can you please send the output of
```
cat /etc/lsb-release
```
```
cat /etc/upstream-release/lsb-release     
```

unless elementary OS is doing something very strange (like not setting DISTRIB_CODENAME to focal in upstream release like they SHOULD be doing) the current version should work

isteiger:
Screenshot from 2022-01-24 20-07-50

This is interesting. If this is nonstandard, why are they doing this?

theofficialgman:
I assumed that upstream release uses the format and didn't change from DISTRIB_ to VERSION_...
```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAMEformat and didn't change from DISTRIB_ to VERSION_...=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.3 LTS"
```
I know that linux mint (which is what I referenced to see what the variable names were) uses DISTRIB_ for the upstream-release/lsb-release. Not sure if upstream_release its actually standardized or not but I will make a PR to fix this no matter what since its a simple fix. What you did in this PR is still not fully correct

theofficialgman:
actually while I have you, can you tell me if this works on elementaryOS? it requires a properly formatted upstream-release/lsb-release to work on ubuntu derivatives
```
sudo add-apt-repository ppa:openjdk-r/ppa
```

theofficialgman:
also running these commands and sending the output will help me as well determine if their might be a more foolproof way of getting the upstream info then sourcing the file
```
lsb_release -cau
```

theofficialgman:
@isteiger I'm waiting for your response on the above ^

isteiger:
> actually while I have you, can you tell me if this works on elementaryOS?

Screenshot from 2022-01-25 17-07-19

> ```
> lsb_release -ca
> ```
>
> ```
> lsb_release -cau
> ```
Screenshot from 2022-01-25 17-07-55

If there is anything else I can do, lmk.
I will modify this PR with `lsb_release`

theofficialgman:
alright thats enough info, I'll refactor and make a PR to use the lsb_release command directly instead of sourcing the file. It looks like `add-apt-repository` must already be doing that internally since it works. thanks for the notification

isteiger:
No problem, glad I could help!

PR #1438, opened 2022-01-25
improve ubuntu version info check for derivatives in MMC install script

theofficialgman:
/etc/upstream-release/lsb-release file is not standardized and the variable names differ per distro. lsb_release output, however, is standardized, so switch to it instead
(fixes issue @isteiger noticed)

in theory, this should fix your issue @isteiger


isteiger:
Installing right now, and I got passed the part where it said it was an Ubuntu derivative, so it should work from here. Thanks!

theofficialgman:
@isteiger I assume the script finished fine and you were able to run minecraft fine?
(you can just thumbs up this post if thats the case, no need to respond)

PR #1447, opened 2022-01-28
replace generic minecraft logo with lunar client logo

theofficialgman:
I figured if we can't easily change the name to "Lunar Client", we at least can use the logo so its clear even to users who don't read the app description.


Botspot:
Great idea.

PR #1456, opened 2022-02-02
AbiWord

open-sorcerer64:


open-sorcerer64:
Ready for merge???

Botspot:
In the description, can you explain what features distinguish this app from alternative text editors like LibreOffice, WPS Office, and notepad?

cycool29:
AbiWord has no feature better than LibreOffice or WPS, it's only better in loading speed and lightweight (and open source, compared with WPS).

Here is an edited description from [here](https://www.makeuseof.com/tag/abiword-free-lightweight-word-processor/):

> If you need a lightweight word processor compatible with a wide variety of file formats, Abiword might be the right fit for you.
It's not as powerful as Microsoft Word or Libre Office's Writer, but it does have most of the features the average user needs. 
It is open-source and the source code is available at https://gitlab.gnome.org/World/AbiWord.

derryktutt:
> AbiWord has no feature better than LibreOffice or WPS, it's only better in loading speed and lightweight (and open source, compared with WPS).
> 
> Here is an edited description from [here](https://www.makeuseof.com/tag/abiword-free-lightweight-word-processor/):
> 
> > If you need a lightweight word processor compatible with a wide variety of file formats, Abiword might be the right fit for you.
> > It's not as powerful as Microsoft Word or Libre Office's Writer, but it does have most of the features the average user needs.
> > It is open-source and the source code is available at https://gitlab.gnome.org/World/AbiWord.

I have a few corrections:
1. It should be **the** Libre Office Writer
2. It's open-source, you can get the source code at https://gitlab.gnome.org/World/AbiWord.

Crilum:
@derryktutt I don't think it should be `the Libre Office Writer`, I think it should just be `LibreOffice Writer`... That's what it's called on its Menu shortcut:
image

open-sorcerer64:
Ok I'll fix the description

PR #1466, opened 2022-02-06
Record failed apps in PR message

cycool29:
If an app failed to update, we won't know about that unless we check the workflow run. So I record the failed apps into `/tmp/failed_apps`, like how updated apps have been recorded, and print them with the workflow run URL in PR message. 

Here is a live example: https://github.com/cycool29/pi-apps/pull/4


theofficialgman:
looks good, I thought about doing this and was going to make any failures change the workflow status, but that still requires us to check the actions page.

this looks like a good solution to that.

Issue #1436, opened 2022-01-25
Calling `"${DIRECTORY}/manage" update "$i" nofetch` takes too long


theofficialgman:
### Confirmations

- [X] I searched for similar bug reports in https://github.com/Botspot/pi-apps/issues and found none were relevant.
- [X] I tried updating Pi Apps and the issue is still not fixed. [How to update Pi Apps?](https://github.com/Botspot/pi-apps#to-update-pi-apps)

### What happened?

the updater is slow for simple file copy operations

### Description

Each update (app is not being reinstalled, just the new scripts being copied) takes 0.5+ seconds. For what should come down to a simple file copy, this is WAY too long (and only gets worse the bigger pi-apps gets).

The main cause of the issue is how the `manage` script is called for every update app command. Each time, multiple internet based checks are made
https://github.com/Botspot/pi-apps/blob/bf9f8886a03163b8a89d60d65259ca6a8bbe8898/manage#L40
https://github.com/Botspot/pi-apps/blob/bf9f8886a03163b8a89d60d65259ca6a8bbe8898/manage#L47

on my end, these two calls take up a combined 0.45 seconds alone, remember this is per-app since each app update calls the manage script (and this is on a wired internet connection, the penalty will only be worse for people with slower internet or bad wifi).

I'm making this as an issue to better gather feedback for what should be done. Should we move these checks behind `if` statements so they only run when necessary? how about ignore them for non-reinstalling updates? maybe we restructure the manage script so it can accept multiple apps in the update case instead of doing the for loop in the `updater` script
https://github.com/Botspot/pi-apps/blob/bf9f8886a03163b8a89d60d65259ca6a8bbe8898/updater#L236-L239

### What is your operating system?

L4T Ubuntu Bionic

Botspot:
I've been thinking of removing all updating aspects from the `manage` script and placing them in the `updater` script.
Now, this would still end up calling an instance of `manage` for every app that needs to be reinstalled, but it would avoid running `manage` when an app simply needs to be copied, and doesn't need to be reinstalled.

Botspot:
Closing, because I've made these changes. https://github.com/Botspot/pi-apps/commit/53cc33b3c5f4f1fd772ba8de9c782b0430cdaf86

PR #1471, opened 2022-02-07
Improve updater script to run faster

Botspot:
Once merged, this will solve issue https://github.com/Botspot/pi-apps/issues/1436

Making major changes to the `updater` script is risky, because if anything breaks then many users will be stranded with an outdated Pi-Apps installation for 3 months. (before the forced updater in the `manage` script kicks in)

**Testers wanted.**

### DIFF
--- a/updater
+++ b/updater
@@ -1,12 +1,18 @@
 #!/bin/bash
 { #prevents errors if script was modified while in use
 
-DIRECTORY="$(readlink -f "$(dirname "$0")")"
-
 function error {
-  echo -e "\e[91m$1\e[39m"
+  echo -e "\e[91m$1\e[39m" 1>&2
   exit 1
 }
+
+[ -z "$DIRECTORY" ] && DIRECTORY="$(readlink -f "$(dirname "$0")")"
+
+#if being sourced, ensure DIRECTORY variable set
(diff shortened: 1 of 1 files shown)


Botspot:
Tagging a few people:
@theofficialgman @cycool29 @Itai-Nelken @mobilegmYT

theofficialgman:
> Making major changes to the `updater` script is risky, because if anything breaks then many users will be stranded with an outdated Pi-Apps installation for 3 months. (before the forced updater in the `manage` script kicks in)

now might be a good time to change that as well... do you think it would be a beneficial to put something here to force the "3 month update", if this command returns with an error code thats not 0 or 1 (since the error function can make it exit with 1)? 
syntax errors in the script will exit with an error code of 2
https://github.com/Botspot/pi-apps/blob/73cf1d93cf657091fce7929047f69e89f3ba9939/gui#L725

syntax errors in the script will exit with an error code of 2

Sussy-OS:
May I help test this, @Botspot ?

Botspot:
> May I help test this, @Botspot ?

Sure, but I think I'm going to merge this today as I've used it without incident.

Botspot:
These changes were made with commit https://github.com/Botspot/pi-apps/commit/53cc33b3c5f4f1fd772ba8de9c782b0430cdaf86.

PR #1482, opened 2022-02-11
Use DEB package for WhatsApp

cycool29:
I just made WhatsApp into DEBs, so it would be easier to install.


cycool29:
> Did you include the fix?

Do you mean the `WhatsApp works with Google Chrome 60+` fix?

Yes, I have both JS fix and removing Service Worker fix in the DEB

Itai-Nelken:
Thanks!
For future updates, @Botspot made a script to build and package both whatsapp and stackedit: https://github.com/Itai-Nelken/Nativefier-WebApps/releases/tag/v4.0

Issue #1502, opened 2022-02-17
Add ledcontrol


Juliandev02:
### What is the name of the app?

ledcontrol

### (Optional) Where is the app hosted?

GitHub.
See https://github.com/Strawberry-Software-Industries/ledcontrol

### About the app

This is a simple program to control the status LEDs of your Raspberry Pi.
Just install this App, and type `ledcontrol` into the Terminal. To change the Behavior of any led, you need root perms. 
Example: `sudo ledcontrol dr` - This will deactivate the red led
###
![image](https://user-images.githubusercontent.com/81520713/154499526-5086b402-4f4e-4392-882d-928c133f029a.png)
:)
Soon i will build armhf Executeables!
If the app should come in, I will also design an icon for it later. 

### Upload file or Add PR Link

[ledcontrol.zip](https://github.com/Botspot/pi-apps/files/8089205/ledcontrol.zip)

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.
- [X] I have confirmed that this app can run on Raspberry Pi.

Botspot:
This looks interesting and I'm fine with it being added to Pi-Apps.

I vaguely recall that various Raspberry Pi models treated the LEDs differently. I remember being able to control both built-in LEDS on the Pi3, but the method I used would not work for the Pi4.
Also the Pi zero uses its LED differently.
How are these exceptions handled by the program?

Juliandev02:
Currently (as far as I know) the program only works on the Raspberry Pi 4, because I don't know what the LED map looks like on the Pi 3. The same is true for the Pi Zero. I would have to look more closely at how the LED map is on the Pi 3 and the Zero.
I know that led1 is the red power LED and led0 is the green one.

Juliandev02:
I'm sure I'll get it right. I just need to identify which model you have (via Bash Command is possible), and then find the appropriate LED map. 
And yes, the Pi 4 is extremely hard to find. I was extremely lucky that I still got my Pi 4 with 8GB RAM ...

theofficialgman:
closing as prone to breakage (new hardware) and not really useful for desktop users. these are status LEDs and not intended to be utilized by the user (other kernel drivers and hardware interact with them).

PR #1503, opened 2022-02-17
Added ledcontrol

Juliandev02:
More Details in #1502 :)

### DIFF
--- a/apps/ledcontrol/credits
+++ b/apps/ledcontrol/credits
@@ -1,0 +1,1 @@
+Juliandev02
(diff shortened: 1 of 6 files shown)


Botspot:
I would advise waiting until there is:

- Pi3 support
- armhf support

Juliandev02:
kk no problem!

Juliandev02:
I have now looked on the internet.... And I found out that the status LED map of the Pi 3 and Pi 4 is exactly the same. Only the map of the Pi Zero is different. Even the Pi 2 B has the same map. 
And I sit since 4 days to create the configuration...and then realize that I don't really need that at all

Juliandev02:
image
uhm yeah... I think all this was not necessary...
Whatever. I now save the current code completely, and do a rollback.

PR #1524, opened 2022-02-24
Fix scrcpy script

derryktutt:
The scrcpy script did not install all required packages for it to install and did not install correctly . 
https://github.com/Genymobile/scrcpy/blob/master/BUILD.md#simple

### DIFF
--- a/apps/Scrcpy/install
+++ b/apps/Scrcpy/install
@@ -7,13 +7,11 @@ echo " · On some devices, you also need to enable an additional option to contr
 echo " · If you have issues, try to run the app a couple of times through Terminal."
 sudo rm -rf ~/scrcpy /usr/local/share/scrcpy ~/.local/share/applications/scrcpy.desktop 2>/dev/null
 
-install_packages adb ffmpeg libsdl2-2.0-0 || exit 1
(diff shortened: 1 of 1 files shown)


derryktutt:
dont know if this would be possible, but to update without reinstall you can run:
```
cd scrcpy
git pull
./install_release.sh
```

Botspot:
Currently, scrcpy downloads a precompiled release, so build tools are unnecessary. This PR would change its behavior to compile scrcpy. Why?

derryktutt:
scrcpy does not work wit the precompiled version for me.

Botspot:
@derryktutt, we've received an error report for Scrcpy 3 hours ago.
Could you look into it and try to figure out what happened?
[install-fail-Scrcpy.txt](https://github.com/Botspot/pi-apps/files/8142098/install-fail-Scrcpy.txt)

Botspot:
@derryktutt, I think I fixed it.

Turns out this was caused by the internal `wget` function that Pi-Apps uses to speed up downloads.
Within that `wget` function, there were a couple of commands that exited with a non-zero exit code. Under normal circumstances, this was no big deal, but when `~/scrcpy/install_release.sh` uses `set -e`, suddenly now those internals within the `wget` function cause the entire script to abort.

With commit https://github.com/Botspot/pi-apps/commit/2b51911dfabb2e6b5c740b10658a63c9cf8b4e02, I've corrected those lines within the `api` script's `wget` function. Now Scrcpy should install successfully.

Botspot:
@derryktutt, if you use Scrcpy regularly, would you consider helping me to write a better description for this app?

PR #1528, opened 2022-02-26
Include Flameshot in apps to hide for TwisterOS

MakePie:
Flameshot is not included in the apps to hide for TwisterOS even though it is already installed on Twister.

### DIFF
--- a/install
+++ b/install
@@ -118,7 +118,8 @@ Windows 10 Theme
 Chromium Widevine
 Lightpad
 Wine (x86)
-Mac OS Theme"
+Mac OS Theme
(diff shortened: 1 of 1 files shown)


Botspot:
Thank you for the contribution! :)

Itai-Nelken:
It wasn't hidden because the version that comes with TwisterOS was much older while this one is the newest.

Botspot:
> It wasn't hidden because the version that comes with TwisterOS was much older while this one is the newest.

Thank you for reminding me. I've changed it back.

PR #1533, opened 2022-02-26
snapdrop: update Electron version, use debs

ryanfortner:
- uses debs to install Snapdrop for easier install/uninstall
- updated electron version, so the app actually works


ryanfortner:
@Botspot

Issue #1495, opened 2022-02-15
Is it possible to add Fritzing app?


ThaPg:
### What is the name of the app?

Fritzing

### (Optional) Where is the app hosted?

https://github.com/fritzing/fritzing-app

https://github.com/fritzing/fritzing-parts

### About the app

The Fritzing application is an Electronic Design Automation software with a low entry barrier, suited for the needs of makers and hobbyists. It offers a unique real-life "breadboard" view, and a parts library with many commonly used high-level components. Fritzing makes it very easy to communicate about circuits, as well as to turn them into PCB layouts ready for production. It is particularly popular among Arduino and Raspberry Pi users, and is widely used in education and creative tinkering.

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.
- [X] I have confirmed that this app can run on Raspberry Pi.

ThaPg:
I have seen that you can intall the program using `sudo apt-get install fritzing` but it is in version 0.9.3. Now it is 0.9.9. 

Version 0.9.3 is from July 2016
Version 0.9.9 is from September 2021.

ThaPg:
I saw them. Unfortunately it is confused and difficult for me to understand them. Is it possible to create installation script?

Also i found that: https://github.com/fritzing/fritzing-app/wiki/1.3-Linux-notes

Thanks.

theofficialgman:
I'll look into this soon. I really like Fritzing and have used it plenty on Windows 10. I honestly forgot it was FOSS.

I just checked their build check, and they are building on ubuntu xenial, so there shouldn't be any issues building on bionic/buster or newer

ThaPg:
@theofficialgman thanks for the answer. Unfortunately I don't have the knowledge to understand all the commands and coding tips to create a build. 

I'll wait for your .zip archive to try the installation.

I love this program and I'm using in Windows 10, but in version 0.9.3b. 
Also what means FOSS and CI?

theofficialgman:
update on this
builds perfectly on ubuntu bionic without any changes
```
cd $HOME
mkdir Fritzing
cd Fritzing
git clone https://github.com/fritzing/fritzing-parts.git parts
cd /tmp
git clone https://github.com/fritzing/fritzing-app.git
cd fritzing-app
[...]

PR #1539, opened 2022-02-27
add Fritzing app

theofficialgman:
closes: https://github.com/Botspot/pi-apps/issues/1495


theofficialgman:
on my end the build took less than 10 minutes.
I prefer that this remains a source install and not converted to a deb/binary install to maintain compatibility with the largest number of systems

theofficialgman:
I will also note, 0.9.6 (from Feb 21, 2021) is the last version buildable on QT 5.9.5 (of bionic) and 5.11 of buster. This is the same version available on debian bookworm and ubuntu jammy, so its up to date enough.

0.9.7 introduced some changes including `Qt::SplitBehaviorFlags` which is only available on qt 5.14+

PR #1540, opened 2022-02-28
Use `$(dirname "${file}")` as argument of `-d` in aria2c flags

cycool29:
To fix `Permission Denied` error in aria2c.


Botspot:
Interesting.

PR #1550, opened 2022-03-01
PiGro Installer Update

actionschnitzel:
As discussed, I changed the installation file.  https://github.com/actionschnitzel/PiGro-Aid-/issues/47
Also, only "start.sh" needs to be made executable, since pigro will do chmod +x for all .sh files automatically. And I removed neofetch from the install list because it's only optional.

### DIFF
--- a/apps/PiGro/install
+++ b/apps/PiGro/install
@@ -2,26 +2,12 @@
 rm -rf ~/PiGro-Aid- || error "Failed to first remove ~/PiGro-Aid- folder!"
 git_clone https://github.com/actionschnitzel/PiGro-Aid- || error "Failed to clone repository!"
 # Get dependencies
-install_packages neofetch || exit 1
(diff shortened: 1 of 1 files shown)


Botspot:
Looks good, but from your end you need to make sure that PiGro won't mark the `xterm python3-pil python3-pil.imagetk python3-pip mpg123 lolcat` [aclages as **manually installed**. Basically, add an `if` switch to check if each package is missing before installing it.
If PiGro *does* mark each of those packages as **manually installed**, then the `purge_packages` function in our `uninstall` script won't actually do anything.

Botspot:
Also, you said that you "changed the installation file as I suggested", but I don't see it in [recent commits](https://github.com/actionschnitzel/PiGro-Aid-/commits).

actionschnitzel:
I found several typos at the last minute, so I haven't merged the update branch with main yet. I think I'll have this finished in the next few hours

actionschnitzel:
So.. I think it's all done. If pi-apps installs the dependencies, there should be no problems.

PR #1554, opened 2022-03-02
Fix #1551

### LINKED ISSUE #1551: PiKISS GUI cannot be installed
### Confirmations

- [X] I searched for similar bug reports in https://github.com/Botspot/pi-apps/issues and found none were relevant.
- [X] I tried updating Pi Apps and the issue is still not fixed. [How to update Pi Apps?](https://github.com/Botspot/pi-apps#to-update-pi-apps)

### What happened?

PiKISS GUI cannot be installed on Debian Buster (?)

### Description

https://github.com/Jai-JAP/pikiss-gui
```
You'll need the following dependencies:

meson >= 0.57
```
Not [available](https://packages.debian.org/buster-backports/meson) for Debian 10 Buster?

### What is your operating system?

RaspiOS 32bit Debian10

### (Optional) Error log? Terminal output? Debug messages?

_No response_

Jai-JAP:
Fixes #1551

### DIFF
--- a/apps/PiKISS GUI/install
+++ b/apps/PiKISS GUI/install
@@ -5,10 +5,11 @@ git_clone https://github.com/Jai-JAP/pikiss-gui || error 'Failed to clone PiKISS
 install_packages ninja-build gobject-introspection libgee-0.8-dev libgirepository1.0-dev libgtk-3-dev valac libgnome-menu-3-dev || exit 1
 
 if package_installed meson; then
-  sudo apt-get purge -y meson || error 'Failed to uninstall older version of meson'
+  sudo pip3 install --upgrade meson || error 'Failed to upgrade older version of meson from pip'
+else
+  sudo pip3 install meson || error 'Failed to install meson from pip'
 fi
 
-sudo pip3 install meson || error 'Failed to install meson from pip'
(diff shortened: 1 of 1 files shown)


apps/PiKISS GUI/install
```diff
@@ -3,6 +3,11 @@
 git_clone https://github.com/Jai-JAP/pikiss-gui || error 'Failed to clone PiKISS GUI repository!'
 
 install_packages ninja-build gobject-introspection libgee-0.8-dev libgirepository1.0-dev libgtk-3-dev valac libgnome-menu-3-dev || exit 1
+
+if package_installed meson; then
+  sudo apt-get purge -y meson || error 'Failed to uninstall older version of meson'
```
Botspot, line 8: This could be dangerous, as other packages or apps from P-Apps may depend on the `meson` package. Purging meson like this will also purge anything that requires it.

Jai-JAP: Ok then we have to just force update meson from pip

Botspot: > Ok then we have to just force update meson from pip

I think that would be a lot safer.

Issue #1341, opened 2022-01-03
VMWare Horizon Client


DaveDischord:
### Cool ideas?

Would it be possible to add VMWare Horizon Client?  You can find the v2111 tar here: [https://customerconnect.vmware.com/en/downloads/details?downloadGroup=CART22FH2_LIN_2111_TARBALL&productId=1027&rPId=62297#product_downloads](https://customerconnect.vmware.com/en/downloads/details?downloadGroup=CART22FH2_LIN_2111_TARBALL&productId=1027&rPId=62297#product_downloads)

and the Doc here: [https://docs.vmware.com/en/VMware-Horizon-Client-for-Linux/2111/rn/vmware-horizon-client-for-linux-2111-release-notes/index.html#whats-new](https://docs.vmware.com/en/VMware-Horizon-Client-for-Linux/2111/rn/vmware-horizon-client-for-linux-2111-release-notes/index.html#whats-new)

Botspot:
Do you know if it runs on Raspberry Pi?

DaveDischord:
Yes.  I have got it installed and tested to an extent on a pi 4.

cycool29:
After a huge work of extracting tarball and copying binaries, I got it launched:

![](https://i.imgur.com/Gyau5gt.jpg)

Crilum:
@cycool29 would it be possible to put it in a deb?

cycool29:
> @cycool29 would it be possible to put it in a deb?

Yes, but I need someone to test the app. I am not a VMWare user.

I will put the deb here later.

DaveDischord:
I have a few Pis I've wanted to install it on, so I'll happily test it out.

Botspot:
@cycool29?

cycool29:
Sorry, I've forgotten about this.

Here is the deb: https://transfer.sh/L1MSGG/vmware-horizon-client_2111-8.4.0-18957622_armhf.deb

DaveDischord:
OK.  So I had Horizon Client 2106 installed.  Using your package, it
successfully updated to 2111 and the application worked fine.  I also
installed on a fresh rPi OS and it seems to be working as well.  I will try
with different scenarios.

DaveDischord:
OK, so I just attempted an install on Raspberry Pi OS 64-bit and got this error: **Errors were encountered while processing: vmware-horizon-client:armhf**

Could it be an issue with the 64-bit architecture?

DaveDischord:
[...]

PR #1555, opened 2022-03-03
Add VMware Horizon Client

cycool29:
To close https://github.com/Botspot/pi-apps/issues/1341.


Botspot:
Any chance we can ever get an `install-64`?

cycool29:
> Any chance we can ever get an `install-64`?

Maybe with a `chroot`. Not sure if it runs well.

Botspot:
> > Any chance we can ever get an `install-64`?
> 
> Maybe with a `chroot`. Not sure if it runs well.

Don't worry about it then. It's fine to wait until RPiOS 64-bit becomes more mainstream, and then maybe vmware will release an arm64 version then.

PR #1558, opened 2022-03-04
Fix SpeedTest CLI update script and add markdown syntaxes in `update_github_script.sh`

cycool29:
To fix the error reported in https://github.com/Botspot/pi-apps/pull/1557.

Added list and bold markdown syntax to make the update app PR looks nicer.


.github/workflows/updates/SpeedTest-CLI.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-webVer=$(wget -qO- https://www.speedtest.net/apps/cli | grep -o ".*-linux-armhf.tgz" | sed 's+.*armel</a></li><li><a href="++g' | sed 's/.*ookla-speedtest-//g; s/-linux-armhf.tgz//g')
+webVer=$(wget -qO- https://www.speedtest.net/apps/cli -U "Mozilla/5.0 (X11; CrOS aarch64 13597.84.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.106 Safari/537.36" | grep -o ".*-linux-armhf.tgz" | sed 's+.*armel</a></li><li><a href="++g' | sed 's/.*ookla-speedtest-//g; s/-linux-armhf.tgz//g')
```
theofficialgman, line 3: so whats you theory here? do you really think speedtest.net denied access because of lacking the user agent? I've tried it myself and never needed one and this script has ben working for weeks if not months without this issue

cycool29: > so whats you theory here? do you really think speedtest.net denied access because of lacking the user agent? I've tried it myself and never needed one and this script has ben working for weeks if not months without this issue

I tested on my own machine without a user agent and it doesn't work (it says `service unavailable`), but it works with a user agent. 
I think the website is receiving spam requests so they blocked `wget`'s user agent in its `htaccess`.

PR #1580, opened 2022-03-09
Botspots: GUI update [DRAFT]

theofficialgman:
going to make this so comments can easily be made for one click additions


gui
```diff
@@ -191,570 +438,258 @@ runonce <<"EOF"
...
     
-    prevaction="$action" #remember the previous action we came from
+    #constantly read from the pipe and send all output to yad
+    #if an item is selected, this script will be run with the row's contents
+    tail -F $pipe | yad --center --title='Pi-Apps' \
+      --image="${DIRECTORY}/icons/logo-64.png" --image-on-top \
+      --list --no-headers --column=:IMG --column=:IMG --column=Name --column=Sysname:HD --column=tip:HD \
+      --separator='\n' --window-icon="${DIRECTORY}/icons/logo.png" \
+      --print-column=4 --tooltip-column=5 \
+      --dclick-action=true --select-action "env pipe=$pipe geometry2=$geometry2 '$0'" \
```
theofficialgman, line 568:
```suggestion
      --dclick-action "env pipe=$pipe geometry2=$geometry2 '${DIRECTORY}/gui'" --select-action "env pipe=$pipe geometry2=$geometry2 ''${DIRECTORY}/gui''" \
```

this makes it so double clicking an app still opens the window. for example, if the user clicks on an app, the window opens, and if they they unfocus that window it will close. if they want to open it again, they have to click on a DIFFERENT app then back to the app they are interested in. This fixes that so now they can just double click on the app and it will open again (should be intuitive)

gui
```diff
@@ -191,570 +438,258 @@ runonce <<"EOF"
...
+      --button="!${DIRECTORY}/icons/search.png!Search":"env pipe=$pipe geometry2=$geometry2 '$0' '' '' Search/ Search/" \
```
theofficialgman, line 569:
```suggestion
      --button="!${DIRECTORY}/icons/search.png!Search":"env pipe=$pipe geometry2=$geometry2 '${DIRECTORY}/gui' '' '' Search/ Search/" \
```

gui
```diff
@@ -191,570 +438,258 @@ runonce <<"EOF"
...
-install() {
-  app="$1" #one app name per line
+#In YAD mode, two windows are handled in a side-by-side configuration. As a group, they must be centered on the the screen.
+#In Xlunch mode, the window must be centeredm but xlunch only allows absolute offsets.
+{
+  #determine screen_width and screen_height
+  xrandr_out="$(xrandr | grep "HDMI-1" | awk '{print $4}' | tr 'x+' ' ')"
+  screen_width="$(awk '{print $1}' <<<"$xrandr_out")"
+  screen_height="$(awk '{print $2}' <<<"$xrandr_out")"
+  unset xrandr_out
```
theofficialgman, line 454:
```suggestion
  #determine screen_width and screen_height
  screen_dimensions="$(xrandr --nograb --current |  awk -F 'connected |\\+|\\('  '/ connected.*[0-9]+x[0-9]+\+/ && $2 {printf $2 ", "}' | sed -n -e 's/^.*primary //p' | tr 'x+' ' ' | tr ',+' ' ')"
  screen_width="$(awk '{print $1}' <<<"$screen_dimensions")"
  screen_height="$(awk '{print $2}' <<<"$screen_dimensions")"
  unset screen_dimensions
```

gui
```diff
@@ -191,570 +438,258 @@ runonce <<"EOF"
...
+    function error {
+      echo -e "\""\e[91m$1\e[39m"\""
+      echo 'Close this terminal to exit.'
+      echo '' > /tmp/xlunchfailed
+      sleep infinity
+    }
+    trap 'echo "\"""\"" > /tmp/terminalexit' EXIT
+    
+    #uninstall xlunch first
+    sudo rm -rf /etc/xlunch /usr/share/xlunch /usr/bin/xlunch /usr/bin/genentries /usr/bin/updateentries /usr/bin/genentries.desktop.sh /usr/share/icons/hicolor/48x48/apps/xlunch_ghost.png /usr/share/icons/hicolor/48x48/apps/xlunch.png /usr/share/applications/genentries.desktop
```
theofficialgman, line 504: same issue as I pointed out before. most users don't have passwordless sudo (its super insecure and should never be used).
this sudo is used in the gui in a hidden terminal prompt if used from the pi-apps desktop icon, so this script will always fail to run.

theofficialgman:
https://github.com/Botspot/pi-apps/blob/7a19fd1ffebb099280ad8e731836b1c933ae58b4/gui#L382-L386
this is dangerous as this can being run without a terminal showing. calling sudo without a terminal showing will always fail (unless the user uses the bad insecure passwordless sudo). 
if this happens, control is not given back to the terminal after exiting pi-apps and you will get garbage on the output:
```
Sorry, try again.
[sudo] password for garrett: 
sudo: 1 incorrect password attempt
The command 'sudo apt update' failed. Before Pi-Apps will work, you must fix your apt package-management system.
runonce(): '  if [ ! -e "${DIRECTORY}/data/categories" ];then
    "${DIRECTORY}/install"
  fi' failed. Not adding hash to list.
```

consider using `pkexec` for all unsafe uses of sudo within hidden terminal scripts

gui
Botspot: This is actually running inside of `terminal-run`, so if a password is required then the user will be prompted for it.

Sussy-OS:
hi, im back

Sussy-OS:
@Botspot Why does it take a minute longer to load the apps/catagories? And in the terminal, it says `no overrides (override for app directory)`

Botspot:
> @Botspot Why does it take a minute longer to load the apps/catagories? And in the terminal, it says `no overrides (override for app directory)`

@Sussy-OS, are you sure that you have switched branches correctly? The pi-apps gui should look like this:
2022-03-09-122141_1920x1080_scrot

theofficialgman:
just for reference this is an easy way to try out the new gui without affecting anything locally
```
cd /tmp
rm -rf /tmp/pi-apps
git clone https://github.com/Botspot/pi-apps.git
cd pi-apps
git checkout new-gui
./gui
```

Sussy-OS:
> > @Botspot Why does it take a minute longer to load the apps/catagories? And in the terminal, it says `no overrides (override for app directory)`
> 
> @Sussy-OS, are you sure that you have switched branches correctly? The new Pi-Apps GUI should look like this: 2022-03-09-122141_1920x1080_scrot

@Botspot Yes, it does look like that, but when I click on an app/a catagory, it takes a while to either, load the app, or load the apps in the catagory.

preload
```diff
@@ -110,130 +85,161 @@ if [ ! -f "$listfile" ] || [ -z "$(cat "$listfile")" ];then
...
+    #preload apps
+    for app in $APPS
+    do
+      #get installation status of app
+      status="$(cat "${DIRECTORY}/data/status/${app}" 2>/dev/null)"
+      
+      #determine status icon for app
+      if [ -z "$status" ];then
+        #app has never been installed, so no icon
+        add="${DIRECTORY}/icons/none-1.png"$'\n'
```
theofficialgman, line 172: I honestly think this looks kinda jarring (the no icon) next to uninstalled icons. Maybe we could add a bag (or whatever the pi-apps logo is supposed to be) with nothing in it (no ❌  and no ✔️ ) for apps that haven't been installed or uninstalled

theofficialgman:
> Also; when I click on the credits of an app, it opens geany, instead of a window showing the credits.

clicking the credits link opens whatever your default program for viewing text files is with the credits file automatically loaded in it. That works for me and opens my text editor with the credits file loaded

theofficialgman:
> Yes, it does look like that, but when I click on an app/a catagory, it takes a while to either, load the app, or load the apps in the catagory.

I wonder if xrandr is taking a while to return output or if
```
time (  screen_dimensions="$(xrandr --nograb --current |  awk -F 'connected |\\+|\\('  '/ connected.*[0-9]+x[0-9]+\+/ && $2 {printf $2 ", "}' | sed -n -e 's/^.*primary //p' | tr 'x+' ' ' | tr ',+' ' ')"; screen_width="$(awk '{print $1}' <<<"$screen_dimensions")"; screen_height="$(awk '{print $2}' <<<"$screen_dimensions")";   unset screen_dimensions )
  ```
  what is the time output here?
  if this isn't taking long then it must be some other command that we will have to add some logging to to see what is slowing stuff down

**Comment by theofficialgman** [comment]
> Yes, it does look like that, but when I click on an app/a catagory, it takes a while to either, load the app, or load the apps in the catagory.

ok I think I see the cause here, it should go away after the first load of a category. timestampscheck takes a while and gets run each time you click on a category. it takes a bit to update the timestamps (less than a second in my testing but its still longer than if they matched). if you click on `All Apps` then all of them will be updated, and the next time you click on any other category it should be fast.

**Comment by Botspot** [comment]
> > Yes, it does look like that, but when I click on an app/a catagory, it takes a while to either, load the app, or load the apps in the catagory.
> 
> ok I may see the cause here, it should go away after the first load of a category. timestampscheck takes a while and gets run each time you click on a category. it takes a bit to update the timestamps (less than a second in my testing but its still longer than if they matched). if you click on `All Apps` then all of them will be updated, and the next time you click on any other category it should be fast.

Not quite. On first launch, all categories need to be preloaded. The main first page is the fastest to load because it is just categories. But all other app-lists take a bit longer to be generated for the first time. So, if you want to judge the average user's experience, then launch `gui` and wait 7 or so seconds for `preload-daemon` to finish loading all other subcategories. You can tell when it's done when a CPU-usage monitor falls back to normal levels.
Once preloading completes, the loading of categories is nearly instant. Again, under normal circumstances, the user won't notice this sluggishness. (The `pi-apps/install` script` will have already preloaded categories, and most other scripts will run a background process of `preload` during use)

**Comment by Botspot** [comment]
Note: if you do wait a bit before selecting a category and it's still slow, then chances are `preload-daemon` did not run. Most likely it immediately exited because it detected that "Another process of preload-daemon is already running. Exiting now." In the case of testing a secondary Pi-Apps installation, it's possible that your main Pi-Apps installation *already* has `preload-daemon` running for its own session. Obviously, `preload-daemon` running for `~/pi-apps` won't do a bit of good for `/tmp/pi-apps`.

If you think that's the case, then try testing Pi-Apps like this:
```
killall preload-daemon
/tmp/pi-apps/gui
```

**Comment by Sussy-OS** [comment]
> > > Yes, it does look like that, but when I click on an app/a catagory, it takes a while to either, load the app, or load the apps in the catagory.
> > 
> > 
> > ok I may see the cause here, it should go away after the first load of a category. timestampscheck takes a while and gets run each time you click on a category. it takes a bit to update the timestamps (less than a second in my testing but its still longer than if they matched). if you click on `All Apps` then all of them will be updated, and the next time you click on any other category it should be fast.
> 
> Not quite. On first launch, all categories need to be preloaded. The main first page is the fastest to load because it is just categories. But all other app-lists take a bit longer to be generated for the first time. So, if you want to judge the average user's experience, then launch `gui` and wait 7 or so seconds for `preload-daemon` to finish loading all other subcategories. You can tell when it's done when a CPU-usage monitor falls back to normal levels. Once preloading completes, the loading of categories is nearly instant. Again, under normal circumstances, the user won't notice this sluggishness. (The `pi-apps/install` script will have already preloaded categories, and most other scripts will run a background process of `preload` during use)

Ah, thanks @Botspot .

File: preload (hunk 5 above)
Comment (Sussy-OS) [reply preload:172]: Hmm. Okay

**Comment by Sussy-OS** [comment]
But, @Botspot , I have one more question: what theme are you using? It looks like the `Windows 10` Theme.

**Comment by Botspot** [comment]
> But, @Botspot , I have one more question: what theme are you using? It looks like the `Windows 10` Theme.

Yes, it is the Windows 10 theme. No, it doesn't fully work but I put up with it. Bullseye has proven to be quite buggy. :(

**Comment by Sussy-OS** [comment]
> > But, @Botspot , I have one more question: what theme are you using? It looks like the `Windows 10` Theme.
> 
> Yes, it is the Windows 10 theme. No, it doesn't fully work but I put up with it. Bullseye has proven itself to be quite buggy. :(

Ah, okay. I do not like how Bullseye is disabling mouses, unless you use tint2. Is there _any_ current fix for the taskbar sometimes turning to black/the main application is shown, but weirdly?

File: preload (hunk 5 above)
Comment (Botspot) [reply preload:172]: Basically, I considered the perspective of an average new user. According to our analytics, around three apps are installed for every Pi-Apps download. Not very many.

When a new user first runs Pi-Apps, I realized that it would be quite distracting to see a long list of red ❌'s. Also it would be hard to spot which apps have been installed, and subsequently uninstalled.
Removing the default status icon from the list, I believe, improves the first-impressions as it cuts down on visual noise.

@theofficialgman, You say it's "jarring." Is that simply because it's different than what you're used to?
Maybe I could do an empty bag by default, but that would increase visual noise, especially for first launch.

**Comment by Botspot** [comment]
@Sussy-OS, Please open an issue on my Windows-10 repository if you want to continue discussing this.

**Comment by Sussy-OS** [comment]
> @Sussy-OS, Please open an issue on my Windows-10 repository if you want to continue discussing this.

Okay, sorry for getting off-topic.

File: preload (hunk 5 above)
Comment (Botspot) [reply preload:172]: Part of me wants to remove the status icons altogether.
2022-03-09-140458_1920x1080_scrot
Though that may frustrate our power-users who use Pi-Apps frequently and depend on the status icons.

Comment (theofficialgman) [reply preload:172]: can we change the color of the text for each app? maybe just do that to represent install status (green for installed, etc)

Comment (Sussy-OS) [reply preload:172]: > can we change the color of the text for each app? maybe just do that to represent install status (green for installed, etc)

Great idea!

Comment (Botspot) [reply preload:172]: > can we change the color of the text for each app?

I recently made a yad-based accounting software and found that you *can* change a row's color.
How does this look?
2022-03-09-145540_1920x1080_scrot

Comment (theofficialgman) [reply preload:172]: that works for me.. though I wonder about accessibility here.
with the icons on the side, even color blind users (of any type) would be able to tell the difference since the symbols were there. I guess hovering to get the status will always stay and can be used in that case

Comment (Botspot) [reply preload:172]: > that works for me.. though I wonder about accessibility here. with the icons on the side, even color blind users (of any type) would be able to tell the difference since the symbols were there. I guess hovering to get the status will always stay and can be used in that case

The symbols themselves were always quite small. Arguably, using text-color is better for more people.
I'll upload the changes in a few minutes.

**Comment by Botspot** [comment]
The app list now color-codes the app name, instead of using a dedicated icon to show status.
2022-03-09-145540_1920x1080_scrot
To update, `cd` into wherever you downloaded this branch to, and run `git pull`.

**Comment by Sussy-OS** [comment]
@Botspot , **BIG** error!

**Comment by Botspot** [comment]
> @Botspot , **BIG** error!

It's so big that you forgot to include the error. 😆

**Comment by Sussy-OS** [comment]
> > @Botspot , **BIG** error!
> 
> It's so big that you forgot to include the error. 😆

Hold on, it actually *broke* my install of Pi-Apps.

**Comment by Sussy-OS** [comment]
https://user-images.githubusercontent.com/98842253/157544103-f2519bb6-2277-4dfb-8bce-db8352066e0e.mp4

**Comment by theofficialgman** [comment]
LOL @Sussy-OS so you don't know what this repo is called
its not auto-app-updates that for the github actions BOT to use

this is `new-gui`. read the PR

**Comment by Sussy-OS** [comment]
> LOL @Sussy-OS so you don't know what this branch is called its not the `auto-app-updates` branch that for the github actions BOT to use, and it just tracks the master branch, force pushing as necessary, never use it
> 
> this is `new-gui`. read the PR marking as off-topic

@theofficialgman I do too! I just was trying to force-update Pi-Apps. It also ruined my main install of pi-apps.

**Comment by theofficialgman** [comment]
> @theofficialgman I do too! I just was trying to force-update Pi-Apps. It also ruined my main install of pi-apps.

no you don't. if you think the `auto-app-updates` branch should EVER be used, you don't know what you are doing

**Comment by Sussy-OS** [comment]
>

**Comment by Sussy-OS** [comment]
> 

I don't really like the way you said that, but; I just forgot. I forgot that the branch was only meant for github actions, and I was just trying to replicate my error, just the way it happened before.

**Comment by theofficialgman** [comment]
> It also ruined my main install of pi-apps.

regarding this, unless your main pi-apps install is in your downloads folder (which you did a rm -rf on), it should be untouched

pi-apps knows nothing of the other potential folder locations you could have installed pi-apps to.
just run `./gui` in your main pi-apps install, if its not working, its probably because you did something to it, not this PR

**Comment by Sussy-OS** [comment]
> > It also ruined my main install of pi-apps.
> 
> regarding this, unless your main pi-apps install is in your downloads folder (which you did a rm -rf on), it "should" be untouched
> 
> pi-apps knows nothing of the other potential folder locations you could have installed pi-apps to. just run `./gui` in your main pi-apps install folder, if its not working, its probably because you did something to it, not this PR. whats the actual issue with your pi-apps install?

Ah, nevermind: it just modified the menu shortcut.

**Comment by theofficialgman** [comment]
botspot I just want to note I can not run since your colored update.
can't look into it right now but pretty sure that broke it. my version of yad may not support this

File: gui [hunk 6]
```diff
@@ -67,7 +67,7 @@ fi
 
 #if this script runs itself with a 3rd argument, then yad used it.
 if [ ! -z "$3" ];then
-  input="$4" #the 4th argument is the "sysname" column - invisible and used internally for scripts. MAY BE EMPTY, but the 3rd argument is guaranteed to not be empty.
+  input="$3" #the 3rd argument is the "sysname" column - invisible and used internally for scripts.
```
theofficialgman, line 70: no botspot, I think `input` needs to stay as `input="$4"`
you didn't change the number of arguments output

theofficialgman:
also I'm getting a lot of `(yad:22891): Gtk-WARNING **: 18:11:50.186: Don't know color ''` in the dialog because you don't set the color of the main text...

gui
Botspot: > no botspot, I think `input` needs to stay as `input="$4"` you didn't change the number of arguments output

On my system at least, yad avoids outputting any "special" color columns. So arguments passed to `gui` will be one word to the left.

theofficialgman: on my system, with this change, all submenus break.
removing `--print-column=4` is what stopped the columns from showing up

Botspot:
> also I'm getting a lot of `(yad:22891): Gtk-WARNING **: 18:11:50.186: Don't know color ''` in the dialog because you don't set the color of the main text...

Not everybody is using a dark mode. By avoiding setting a color on black/white list items, yad will default to GTK theme defaults.

Botspot:
Also what version of yad are you using? Mine is `0.40.0 (GTK+ 3.24.24)`

theofficialgman:
> > also I'm getting a lot of `(yad:22891): Gtk-WARNING **: 18:11:50.186: Don't know color ''` in the dialog because you don't set the color of the main text...
> 
> Not everybody is using a dark mode. By avoiding setting a color on black/white list items, yad will default to GTK theme defaults.

yeah which is why I wouldn't make that change... but the logging is terrible
its like 200 lines per second

theofficialgman:
> Also what version of yad are you using? Mine is `0.40.0 (GTK+ 3.24.24)`

mine is older `0.38.2 (GTK+ 3.22.30)`

Botspot:
> mine is older `0.38.2 (GTK+ 3.22.30)`

Interesting. Are you on Bullseye? I thought Stretch, Buster, and Bullseye all used 0.40.0.

> but the logging is terrible
its like 200 lines per second

I don't get those errors with yad 0.40.0.

theofficialgman:
> > mine is older `0.38.2 (GTK+ 3.22.30)`
> 
> Interesting. Are you on Bullseye? I thought Stretch, Buster, and Bullseye all used 0.40.0.
> 
> > but the logging is terrible
> > its like 200 lines per second
> 
> I don't get those errors with yad 0.40.0.

read the edit, I'm on ubuntu bionic. its just what the system has the GTK loglevel set to. might must be set to WARNING. either that or yad removed it from warning gtk

Botspot:
> read the edit, I'm on ubuntu bionic. its just what the system has the GTK loglevel set to.

I get those warnings if I choose an invalid color.
```
(yad:3938): Gtk-WARNING **: 17:52:42.355: Don't know color 'd3epdcjiocphkjelcndk'
```

derryktutt:
Updating will break it
(i can't take screenshot)
`Terminal output:
(yad:10988): Gtk-WARNING **: 07:05:32.611: Don't know color 'App folder'

(yad:10988): Gtk-WARNING **: 07:05:32.611: Don't know color 'App folder'

(yad:10988): Gtk-WARNING **: 07:05:32.611: Don't know color 'App folder'

(yad:10988): Gtk-WARNING **: 07:05:32.612: Don't know color 'App folder'
`
sorry didn't get screenshot, try to reproduce please

theofficialgman:
> Updating will break it (i can't take screenshot)

@derryktutt yes this is expected. please do not do that. updates are pulled from the master branch (which this is not) and will break all the changes made here.

theofficialgman:
a few issues I notice that are due to how old pi-apps is structured...
1: the old GUI continues to run after updating, this causes multiple bugs since new formating has changed multiple lists
2. data/preload lists are not regenerated because the timestamps still match even though the generated lists would be very different (icons have been swapped out for color data, etc)

this is quite the oversight on the previous implementation and I don't think there is anything we can do about it.
2 can be easily solved by just adding a runonce to delete the contents of the data/preload folder the first time the new gui launches, but there isn't any way to kill the old gui after update and instead run the new gui script

Botspot:
> 1: the old GUI continues to run after updating, this causes multiple bugs since new formating has changed multiple lists

2. data/preload lists are not regenerated because the timestamps still match even though the generated lists would be very different (icons have been swapped out for color data, etc)

The preload script will be modified during the update, so timestamps won't match. But no. 1 is still a problem.

theofficialgman:
> The preload script will be modified during the update, so timestamps won't match. But no. 1 is still a problem.

well my Lists didn't regenerate after an update... so idk. that one time runonce works fine though

gui
```diff
@@ -1,28 +1,280 @@
 #!/bin/bash
 
-set -a #export all functions and variables for child processes like runonce()
-
-DIRECTORY="$(readlink -f "$(dirname "$0")")"
+if ! command -v yad &>/dev/null; then
+  error "YAD needs to be installed to run Pi-Apps."
```
theofficialgman, line 4: the api functions haven't been sourced yet, if you pass this check, `error` is not available.

its probably best you just replace this with the appropriate echo and exit from the error function
```suggestion
  echo -e "\e[91mYAD needs to be installed to run Pi-Apps.\e[0m" 1>&2
  exit 1
```

gui
```diff
@@ -1,28 +1,280 @@
...
+    
+  fi
+  exit 0
 fi
 
-#in background, check mission-critical scripts and re-download them if they contain a syntax error
+DIRECTORY="$(readlink -f "$(dirname "$0")")"
+
+set -a #make all functions in the api available to subprocesses
+source "${DIRECTORY}/api" || error "failed to source ${DIRECTORY}/api"
```
theofficialgman, line 264: same issues as before, if the api fails to source, then the error function isn't available so it won't execute
```suggestion
source "${DIRECTORY}/api" || (  echo -e "\e[91mfailed to source ${DIRECTORY}/api\e[0m" 1>&2; exit 1 )
```

Botspot: I think that putting `exit 1` in parenthesis won't cause the script to exit, so I'll just add back the `error()` function to the gui script.

theofficialgman: ah right yes it won't, putting the error function at the top is the easiest solution, you can do that

gui
```diff
@@ -131,13 +389,6 @@ runonce <<"EOF"
...
-#for old installs prior to having categories, re-run install script
-runonce <<"EOF"
-  if [ ! -e "${DIRECTORY}/data/categories" ];then
-    "${DIRECTORY}/install"
-  fi
-EOF
-
 #install dependencies
 runonce <<"EOF"
   dependencies='yad curl wget aria2 lsb-release apt-utils imagemagick bc librsvg2-bin shellcheck'
```
cycool29, line 394:
```suggestion
  dependencies='yad curl wget aria2 lsb-release apt-utils imagemagick bc librsvg2-bin shellcheck wmctrl xdotool'
```

Added `wmctrl` and `xdotool` for window focusing. They just cost about 100kb of disk space.

gui
```diff
@@ -1,28 +1,285 @@
...
+      #if status is corrupted or unknown, then show both buttons
+      whichbutton+=("--button=Uninstall!${DIRECTORY}/icons/uninstall.png:2" "--button=Install!${DIRECTORY}/icons/install.png:4")
+    fi
+    
+    #only display app icon if it exists
+    if [ -f "${DIRECTORY}/apps/${app}/icon-64.png" ];then
+      imageline=(--image="${DIRECTORY}/apps/${app}/icon-64.png" --image-on-top)
+    fi
+    
+    (cat "${DIRECTORY}/apps/$app/description" || echo "Description unavailable") | yad --text-info --fontname=12 --wrap --show-uri \
```
cycool29, line 207:
```suggestion
    (while [ "$(xdotool getactivewindow getwindowname)" != "Details of ${app}" ]; do wmctrl -a "Details of ${app}"; done) &
    (cat "${DIRECTORY}/apps/$app/description" || echo "Description unavailable") | yad --text-info --fontname=12 --wrap --show-uri \
```

Automatically focus on app info window.

cycool29 (changes requested):
Use `wmctrl`  and `xdotool` to focus on the app info window.

Please add them as dependencies in the `install` script (or anywhere else needed), it is not in this PR and I can't make a comment on it.

gui
Botspot: > Automatically focus on app info window.

```bash
(while [ "$(xdotool getactivewindow getwindowname)" != "Details of ${app}" ]; do wmctrl -a "Details of ${app}"; done) &
```
That seems to be a bit dangerous without a delay in the loop.

gui
Botspot: > Added `wmctrl` and `xdotool` for window focusing.

Done. I've also added them to the `install` script.

gui
Botspot: @cycool29 I've added your suggestion with a few safety improvements.

theofficialgman:
also one more note.
as DEs move to using wayland as a backend instead of xorg, the location setting of the window breaks. it pops up in random places. this is due to yad using xwayland and I don't think xwayland supports setting the window location

Botspot:
> also one more note. as DEs move to using wayland as a backend instead of xorg, the location setting of the window breaks. new windows pop up in random places. this is due to yad using xwayland and I don't think xwayland supports setting the window location.

IIRC, Ubuntu uses wayland to an extent. Does the current live version of Pi-Apps launch to the center of the screen?

theofficialgman:
> > also one more note. as DEs move to using wayland as a backend instead of xorg, the location setting of the window breaks. new windows pop up in random places. this is due to yad using xwayland and I don't think xwayland supports setting the window location.
> 
> IIRC, Ubuntu uses wayland to an extent. Does the current live version of Pi-Apps launch to the center of the screen?

nope, and it never will... weston implemented location setting into xwayland weston https://cgit.freedesktop.org/wayland/weston/commit/?id=882aff09324ba72d1d9136d6ad9f3c99dab1631b (weston is a reference implementation that DEs incorportate into their own implementations like mutter and kwin), but this was never included in mutter

theofficialgman:
tracking the issue here in mutter now: https://gitlab.gnome.org/GNOME/mutter/-/issues/2193
we will see if gnome devs ignore this or do anything about it

theofficialgman:
@Botspot turns out yad wasn't using xwayland when I thought it was. it was using wayland and is impossible to set position in wayland. Setting the gdk variable to X11 makes yad uses xwayland and fixes it.
all good on wayland based DEs now

theofficialgman:
> @theofficialgman, with your python script to set text_color, the response time for the `api` script is noticeably slower. I think this is mostly due to the delay in running `yad --version`.
> 
> Would you see any potential problems with moving this code to the `preload` script?

it takes anywhere from 0.15 to 0.2 seconds on my end to run `yad --version`. there isn't any issue with putting that inside the preload script, but is it necessary and would that actually be faster? the preload script is called multiple times as well

also `dpkg-query --showformat='${Version}' --show yad` is an option for getting the version, that seems to run in 0.12 seconds consistently. and thats faster than your `package_latest_version` function by a lot

Botspot:
> it takes anywhere from 0.15 to 0.2 seconds on my end to run `yad --version` (use the `time` command). there isn't any issue with putting that inside the preload script, but is it necessary and would that actually be faster? the preload script is called multiple times as well

All I know is that running `pi-apps/api generate_logo` now takes around 1 second, when it used to be nearly instant. Moving it to the preload script allows for a couple of optimizations, so I'll do that.

> also `dpkg-query --showformat='${Version}' --show yad` is an option for getting the version, that seems to run in 0.12 seconds consistently. and thats faster than your `package_latest_version` function by a lot

The `package_latest_version` function searches all *apt* lists for a particular package. The package does not have to be installed for it to work. It's used in the `package_is_new_enough` function to check if a package is recent enough to work for an app.

theofficialgman:
@Botspot found a fast way... might seem a little strange though

`zcat /usr/share/doc/yad/NEWS.gz | head -n 1 | awk '{print $2}' | sed 's/\.//g'`

takes less than 0.005 seconds

theofficialgman:
the changelog is also an option (though it takes more manipulation

```
zcat /usr/share/doc/yad/changelog.Debian.gz | head -n 1 | awk '{print $2}' | sed 's/\.//g' | sed 's/[()]//g' | sed 's/-.*//g'
```

Botspot:
> @Botspot found a fast way... might seem a little strange though. grabbing the version from the top of the news file

I love it! It reminds me of using `dd` to create a menu button.
BTW, it's easier to remove characters from an input stream with: `tr -d '.[()]`

theofficialgman:
so I just realized I also had the version check wrong as well
I has it as `if [ $yad_version -lt 0440 ]; then`
so everything less than yad version 0.44.0 was running the python script. the latest yad is 0.40.0, so literally everything was running the python script

fixed now and am using the fast method

theofficialgman:
@Botspot see commit: https://github.com/Botspot/pi-apps/pull/1580/commits/466107baf74970591a7ec2adea864afbed4128fc
I've implemented an initial form of waiting for one install/uninstall script to finish before allowing another one to start.

See the changes itself for details. I'm looking for feedback, I'd like to make it 100% reliable if possible. I don't want to maintain a list of waiting installs because if the user does a ctrl+c on an install, that could result in it remaining in the waiting list forever, holding up other installs from continuing

theofficialgman:
alright the waitlist and a fallback incase the waitlist disappears are implemented.

Now the install and uninstall commands from the gui will NOT start unless the manage script is not currently running AND the script is next in line in the waitlist (ie: the app name is at the top of the file).

this makes is so the app install/uninstalls will proceed in the order of the users execution. the terminal run window will popup and show a spinning icon and text saying "Waiting for another script to finish before continuing..." until the previous script finishes

@Botspot how does this look now?

Botspot:
@theofficialgman, actually I've been making my own implementation. This will use one terminal for all concurrent app-installs, in addition to a queue-viewer ®️.
2022-04-08-230438_1920x1080_scrot

theofficialgman:
interesting... well feel free to remove or take inspiration from what I added.
I was just going for the quick and simple method

theofficialgman:
@Botspot why did you remove the python script for fixing older yad?

Botspot:
> @Botspot why did you remove the python script for fixing older yad?

I plan to move it to the `preload` script to avoid unnecessary delays when running non-gui api stuff.

theofficialgman:
Alright that's fine, I just expected you to make that commit at the same time. Carry on

Botspot:
off-topic, but I just found a fantastic way to debug bash scripts: rather than `bash -x` which can be quite cryptic, I found you can make it also output the current *line number* of the script.
`PS4='Line ${LINENO}: ' bash -x /path/to/script` [source](https://stackoverflow.com/questions/17804007/how-to-show-line-number-when-executing-bash-script)

theofficialgman:
the conflict is small and just has to do with your fixing a bug in upstream and not commiting it here.

are you still working on queue function/viewer internally? I can help if you push your in progress changes.
(I suggest doing a revert of my changes in a separate commit first if your stuff conflicts)

Botspot:
@theofficialgman, it took me a few extra days as I was tracking some intermittent issues. I've updated the branch with my implementation of a manage daemon.

theofficialgman:
@Botspot I would have prefered if you just git a revert on my commits, since it looks like much of it has been left in but not removed...

I can't look into this in detail right now, maybe by the weekend I can get to it and I'll fix that just so the commit history is more clear.

Botspot:
> @Botspot I would have prefered if you just git a revert on my commits, since it looks like much of it has been left in but the functionality is not used...

I'm not aware how to revert commits with `git cola`, so I ended up cloning the branch to a fresh folder, copying over .git to my main pi-apps folder, and then running `git cola`. This effectively placed my edits upstream to yours, so it ought to have reversed them.
Git cola seemed to show this working, and a quick browser-search of "/tmp/pi-apps-waitlist" doesn't return anything anymore. Could you show me a place where waitlist code still exists after my commits?

theofficialgman:
hmm.. actually it seems to be all gone, my mistake
but you could have just done a `git revert commithash`, commited that, then added your changes in a separate commit

theofficialgman:
@Botspot I reverted the commits on my local history and verified the final resulting files match exactly. This breaks up your change from my changes which I think helps with clarify of whats going on and what actually changed.

Can I push this?

Botspot:
> @Botspot I reverted the commits on my local history and verified the final resulting files match exactly. This breaks up your change from my changes which I think helps with clarify of whats going on and what actually changed.
> 
> Can I push this? its 3 revert commits (one for each commit I made that was removed) and then your commits on top
> 
> I could also just drop my commits, I'm fine with that

Do whatever you think is best. And in the grand scheme of things, it doesn't really matter as merging this PR will squash everything into a single commit.

theofficialgman:
yes, this history may look scary. 
it is not, its just a rebase on master to solve the conflicts so this can be merged through the github web interface when ready

Botspot:
I fixed multiple bugs in the manage script, and I've decided to retire the app status icons. (installed.png, uninstalled.png, corrupted.png, disabled.png)
Going forward, to indicate an app being installed/uninstalled, it will use two icons: the action, and the result.
Changes need to be made to logviewer and createapp.

theofficialgman:
@Botspot see my previous commit. I tested your changes but the python script ran every time preload was called. the variables set in preload-daemon do not survive it running so text_color was never set (I verified with local debugging). the location I've moved it too is safe and guarantees the python script will only be called once when running the gui

theofficialgman:
I forgot to also fix your check for the variable being unset vs set, the check you used interpreted an unset variable and a blank one as the same. this format does not have that issue.
(see https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash)

Botspot:
> I forgot to also fix your check for the variable being unset vs set, the check you used interpreted an unset variable and one set to the empty string as the same. this format does not have that issue.

Interesting - I was just planning for the `if` statement to check the yad version every time, but this way is even faster.

Botspot:
> Changes need to be made to logviewer and createapp.

Done.

theofficialgman:
let me make a suggestion @Botspot . I think your daemon mode is a good idea but poorly executed

lets say I got to uninstall two apps. the first app brings up a terminal run window asking for my password, once I type it in, then the yad window pops up. when I got to uninstall a second app (when the first is still uninstalling), another terminal run window pops up, asking for the password. when I give it the password then it says I can close it out (because it send the uninstall command to the other window).

why even popup with the second window if all you need it for is to execute the daemon? how about do something like the megascript does and popup a zenity/yad prompt for the user to enter their password in, and ignore the whole need the extra redundant terminal windows

better yet would be to have the user type their password in only once, in the first terminal, and don't even popup the second at all

theofficialgman:
something like this should work fine for both systems with a sudo password and without one
```
          sudo -k
          while ! sudo -n true; do
            zenity --password 2>/dev/null | sudo -S echo "" 2>&1 >/dev/null
          done
```

Botspot:
> better yet would be to have the user type their password in only once, in the first terminal, and don't even popup the second at all

The second terminal should not be appearing at all. The fact that you see both means something is going with with the `terminal-manage` function within the `gui` script.

Botspot:
I was confused why `manage` needed a sudo password for merely signaling the daemon process. This appears to be the culprit:
```bash
#An apt repository's Packages file can be corrupted so that an apt update will silently fail. See: https://bugs.launchpad.net/ubuntu/+source/apt/+bug/1809174
#This line will fix the problem by removing any zero-size Packages files.
find /var/lib/apt/lists -type f -name '*Packages' -size 0 2>/dev/null | xargs sudo rm -f
```

theofficialgman:
> ```shell
> #An apt repository's Packages file can be corrupted so that an apt update will silently fail. See: https://bugs.launchpad.net/ubuntu/+source/apt/+bug/1809174
> #This line will fix the problem by removing any zero-size Packages files.
> find /var/lib/apt/lists -type f -name '*Packages' -size 0 2>/dev/null | xargs sudo rm -f
> ```

you should output the filelist to a variable, then only execute sudo rm -f on that variable list if the variable is non-empty

theofficialgman:
something like this might work
```
removal_list="$(find /var/lib/apt/lists -type f -name '*Packages' -size 0 2>/dev/null)"
if [ ! -z "$var" ]; then
    sudo rm -f "$removal_list"
fi

**Comment by theofficialgman** [comment]
edited above to actually work ^

**Comment by Botspot** [comment]
> something like this should work fine for both systems with a sudo password and without one
> 
> ```
>           sudo -k
>           while ! sudo -n true; do
>             zenity --password 2>/dev/null | sudo -S echo "" 2>&1 >/dev/null
>           done
> ```

Would this work for subprocesses which run sudo? (manage daemon runs an additional manage process for every app install/uninstall, and _that_ process runs the actual app script, which may contain sudo)

**Comment by theofficialgman** [comment]
> Would this work for subprocesses which run sudo? (manage daemon runs an additional manage process for every app install/uninstall, and _that_ process runs the actual app script, which may contain sudo)

as long as the main process which launched sudo, any subprocess *should* inherit that (its what is done in the megascript) at least till the sudo timeout runs out...

**Comment by theofficialgman** [comment]
as a quick hack, I added this directly to the top of the GUI script (don't actually do this)

```
while ! sudo -n true; do
  zenity --password 2>/dev/null | sudo -S echo "" 2>&1 >/dev/null
done
```
I got prompted for my password through zenity before anything even ran, then I went to uninstall something and a terminal run windows opened asking for my sudo password. I entered it and started uninstalling something else, no window opened and it was immediately added to the uninstall list.

interestingly, if I don't implement my hack... the "invisible" main pi-apps script is waiting for me to enter my password on the second uninstall. I say invisiable because its asking for my sudo password in the main terminal since I launched the gui with ./gui

**Comment by theofficialgman** [comment]
ok, so far after implementing this change ^, nothing has gone to the void "yet"

File: manage [hunk 11]
```diff
@@ -214,7 +214,15 @@ EOF
 
 #An apt repository's Packages file can be corrupted so that an apt update will silently fail. See: https://bugs.launchpad.net/ubuntu/+source/apt/+bug/1809174
 #This line will fix the problem by removing any zero-size Packages files.
-find /var/lib/apt/lists -type f -name '*Packages' -size 0 2>/dev/null | xargs sudo rm -f
+removal_list="$(find /var/lib/apt/lists -type f -name '*Packages' -size 0 2>/dev/null)"
+if [ ! -z "$removal_list" ]; then
```
Botspot, line 218: In nearly all cases, the removal list is going to be empty. If your intention is to avoid typing in the sudo password more than once, the dialog should be launched whether or not any Packages files are empty. I would advise putting this in the `terminal_manage` function in the gui script.

theofficialgman: no this is not my intention. I'm fine with typing in the sudo password once in the terminal.
in my experience, the reason the other prompts were showing is BECAUSE this prompt was running sudo every time.

now that it doesn't run unless necessary, the other windows don't show up, and you only type in the sudo to the terminal once. I've tested every possible combination and only one terminal shows up now, and you only type in the password once

theofficialgman: basically, the check from here was added: https://github.com/Botspot/pi-apps/pull/1580#issuecomment-1107918726
and then to deal with the common edge case that the manage script is running in a hidden terminal (which it usually is) the yad dialog is used to get the users sudo password

theofficialgman:
so @Botspot I've identified our one and only spot to inject new code before the old gui runs again when updating to this PR (once it gets merged)

the new version of the `updater` script (specifically the `set-status` runmode) gets run before returning to the old GUI.
this allows us to do anything we want before running the old GUI again after update, which we don't want to do because the UI will be broken.

my current approach is to throw in a runonce into that, to kill the gui script. the user will have to re-launch pi-apps (which I think is fine for a one time thing). the only problem with doing this is, if the user has done a fresh clone of the repo, then this runonce also gets run on their install, even when we don't want to kill the gui in this case. HOWEVER, in my testing, the gui it told to be killed, but because of the threaded gui, nature of the new gui, it doesn't actually get killed and the scripts continue. the original gui script gets killed, yes, but the next version takes over immediately. the only thing that doesn't work is the settings button.... which I guess relies on the existence of the original gui script

theofficialgman:
I've pushed a disabled verison of the code above just so you can see it ^

manage
Botspot: I guess that's fine, but I consider it quite strange to have a dedicated gui for an edge-edge-case scenario. I say "edge" twice, because not only does a Packages file need to be corrupted, but also `sudo` needs to require a password, which is not the default on RPiOS.
So basically, this password GUI may very well not ever be shown, to a single user, for the lifetime of Pi-Apps.

Botspot:
> btw, the way I test the "merge" of this PR is by cloning the master branch of pi-apps, then creating the `update` folder within it, cloning pi-apps again in that, entering that and checking out the new-gui

Good idea. This was trickier than expected. First, while the current directory was ~/pi-apps-old-gui/updates/pi-apps, when switching branches git decided to switch branches in ~/pi-apps-old-gui, rather than the one in the updates folder. Strange.

After working around that bug, I couldn't get the updater to run until manually creating `/home/pi/pi-apps-old-gui"`. The script that would normally create that folder was `manage`, so I guess as a side effect, if users install pi-apps but never install any apps, they won't get any update notifications. This is easy to fix. (Note to self to fix this in the morning)

Botspot:
> so if we could somehow make it so this doesn't get run for new installs, and only gets run on the one time update, it would be pretty much perfect.

I think I found a way to do that. First, I thought of the ways in which the old gui would be unable to view the new-gui's app list file, as the columns have changed. Then I realized that after an update and right before the gui re-appears, the old gui's app list format is still contained in the `preload` files.

The first five lines of a new-gui app `LIST-` file are:
```
/home/pi/pi-apps/icons/categories/All Apps.png
All Apps
All Apps/
App folder
#empty line
```
The old format looked like this:
```
/home/pi/pi-apps/icons/categories/All Apps.png
/home/pi/pi-apps/icons/none-1.png
All Apps
All Apps/
App folder
```
Given my findings, this command should detect if the old gui is in use:
```
if [ -z "$(sed -n 5p "${DIRECTORY}/data/preload/LIST-")" ];then
```
This detection won't work for the xlunch gui. There needs to be a separate way to do that, as the xlunch list format has also been changed.

theofficialgman:
> ```
> if [ -z "$(sed -n 5p "${DIRECTORY}/data/preload/LIST-")" ];then
> ```
The 5th line is the color data, so no, it will not always be empty. I suggest checking whether the second line contains *.png or not if you want to go that route

theofficialgman:
> Then I realized that after an update and right before the gui re-appears, the old gui's app list format is still contained in the `preload` files.

no, this is precisely the problem. the new preload-daemon script gets run by `refresh_app_list` BEFORE the set-status gets called and before the old gui loads back up. so the preload files contain the NEW format, which is the problem. if the old format still existed, then the gui would have no issues

theofficialgman:
@Botspot I think my above comment just did everything that could be wanted but simpler.
if GUI_FORMAT_VERSION isn't 2, then we know we are running the old GUI, so kill it, and start the new one.

if GUI_FORMAT_VERSION == 2, then don't do the runonce because we know we are running the new gui format, so continue.

theofficialgman:
xlunch seems to be broken for me... I think this is a issue with xlunch itself and not how pi-apps is using it but I can't be sure as I never use it.
```
pipe is /tmp/tmp.x2dXswdHfa
Timestamps match.
Reading list file...
Entry too long on line 15, maximum length is 255 characters!
Entry too long on line 17, maximum length is 255 characters!
Entry too long on line 16, maximum length is 255 characters!
Entry too long on line 9, maximum length is 255 characters!
Received 'pi-apps-homepage'
Crash Annotation GraphicsCriticalError: |[0][GFX1-]: glxtest: libEGL missing eglGetDisplayDriverName (t=1.16379) [GFX1-]: glxtest: libEGL missing eglGetDisplayDriverName
Crash Annotation GraphicsCriticalError: |[0][GFX1-]: glxtest: libEGL missing eglGetDisplayDriverName (t=1.16379) |[1][GFX1-]: glxtest: libEGL missing eglGetDisplayDriverName (t=1.1639) [GFX1-]: glxtest: libEGL missing eglGetDisplayDriverName
```
Screenshot from 2022-05-03 16-48-06

theofficialgman:
oh I see the problem... switching to xlunch from the settings didn't trigger a preload reload/removal, so it was using the yad preload lists still

theofficialgman:
there is another bug... introduced by your `source` command, fixing it now

theofficialgman:
alright with the above fix, thats the last of the bugs I can find with the new-gui update

@Botspot requesting your review, lets get this merged soon

theofficialgman:
rebase on master to make merging easier

theofficialgman:
I added some updates to the readme to reflect the new gui. someone with a Pi might want to re-take them so they look the same as the others (my theme does not match botspots)

etc/preload-daemon
```diff
@@ -23,7 +23,7 @@ fi
 source "${DIRECTORY}/api" || error "failed to source ${DIRECTORY}/api"
 
 #for mktimestamps function
-source "${DIRECTORY}/preload" source
+source "${DIRECTORY}/preload" "source"
```
Botspot, line 26: What difference does this change make?

Botspot:
> @Botspot I think my above commit just did everything that could be wanted but simpler. I defined a new variable GUI_FORMAT_VERSION. if GUI_FORMAT_VERSION isn't 2, then we know we are running the old GUI but the new preload and other scripts, so kill the old gui, and start the new one (the new one).
> 
> if GUI_FORMAT_VERSION == 2, then don't do the runonce because we know we are already running the new gui format. if we ever change the format in the future, you just bump the version in these two places and the same thing happens.
> 
> let me know if you foresee any danger in doing this.. I think its safe enough and should work, I've tested it out a few times now and so far its worked every time

I like what you've done here.

etc/preload-daemon
theofficialgman: it fixes the bug. the bug without the quotes is source is not passed as a string like you expect. it gets passed as a command or something. which causes the preload script to not exit here where you expect:
https://github.com/Botspot/pi-apps/blob/52a1d52acae1b29c6141645d275e21de438fbb8e/preload#L34

theofficialgman:
@Botspot I just realized you removed the splashscreen. was that intentional? did you want it gone?

Botspot:
> @Botspot I just realized you removed the splashscreen as part of the original commit. was that intentional? did you want it gone?

This was intentional, as changes to the GUI now make Pi-Apps launch nearly instantly. There isn't a 1-2 second delay for a splash screen to appear, and I don't plan on increasing the launch-time to display it. :)

Crilum:
> > @Botspot I just realized you removed the splashscreen as part of the original commit. was that intentional? did you want it gone?
> 
> This was intentional, as changes to the GUI now make Pi-Apps launch nearly instantly. There isn't a 1-2 second delay for a splash screen to appear, and I don't plan on increasing the launch-time to display it. :)

That's good, but c'mon, the splashscreen is so cool! :man_technologist:

Crilum:
Wait, I have an idea! You could make it so when you click on the Pi-Apps icon in the GUI, it blinks though some app icons!

theofficialgman:
> Well, maybe yad ins't able to do that, but that would be cool!

Unfortunately it can't. We only get output from the user clicking a button at the bottom or clicking in the list

Botspot:
> @Botspot I just got a concern. Does the new manage script break apps that use the manage script internally like steam and wine (which each install boc64/86)?

No - at least my method doesn't. The `install`, `uninstall`, `multi-*`, and `install-if-not-installed` modes have not been changed. Only the new `dameon` mode in the manage script employs a type of PID-based lockfile and queue.

theofficialgman:
> No - at least my method doesn't. The `install`, `uninstall`, `multi-*`, and `install-if-not-installed` modes have not been changed. Only the new `dameon` mode in the manage script employs a type of PID-based lockfile and queue.

ok good. yeah forgot dameon mode had to be specifically input as an arg when calling the manage script.

next week I'm going to have a lot more free time and I'd like to finally get this PR merged. How does that sound?

Botspot:
> next week I'm going to have a lot more free time and I'd like to finally get this PR merged. How does that sound?

I could merge it now if you're comfortable with that.

theofficialgman:
> I could merge it now if you're comfortable with that.

No don't do that. In the off chance something is messed up and needs fixing, I want to have time to do it quickly.
Let's wait till next week.

theofficialgman:
`genapplist-yad.c` is not used at all anymore with this PR (and it wouldn't work with  the new format anyway).
Can we go ahead and remove the file (and any old compiled binary)?

Botspot:
> `genapplist-yad.c` is not used at all anymore with this PR (and it wouldn't work with the new format anyway). Can we go ahead and remove the file?

Done.

Botspot:
I think this is ready to be merged.

theofficialgman:
rebase on master, no changes, rebase was clean

theofficialgman:
don't merge yet, I'm updating the tooltips for the new categories, and these categories need icons

theofficialgman:
@Botspot are the icons from some sort of theme you have? can you go ahead and update them to include the new categories?

Botspot:
> @Botspot are the icons from some sort of theme you have? can you go ahead and update them to include the new categories?

Most of them are from [the archdroid icon theme](https://github.com/GreenRaccoon23/archdroid-icon-theme). I'll see what I can do.

Botspot:
All but the Emulators category now has icons.

Botspot:
Terminals and Engineering may need tweaking for better perception.

Sussy-OS:
I can't wait for this to get released. I'm excited!

Crilum:
I feel like `Conky` and `Conky Rings` would be better in `System Management`, rather than `Appearance`

Crilum:
There are 10 (`Caprine`, `Cawbird`, `Discord`, `Microsoft Teams`, `Telegram`, `Thunderbird`, `Wechat`, `WhatsApp`, `Zoom`, and `Zoom PWA`) messaging/communications apps in PI-Apps, so I think there could be a `Communications` category, or something similar.

theofficialgman:
I think these could go under either appearance (since they actually modify the desktop) or system management (since I was putting things like resource monitors under that category).
lets leave it since it is already under appearance

theofficialgman:
@Crilum my original intention with the refactoring was to decrease the `Tools` category since that became a dumping ground for all apps. there were 62 apps in it, now only 39 apps. 

with that said, I'd be fine with adding a subcategory under internet for `Communications` or as its own primary category. lets see what everyone else thinks

Crilum:
The uninstall icon would look better on dark themes if it had a transparent background:

Before:

image


After:

image


Here's the icon:

uninstall

theofficialgman:
> Here's the icon:
> 
> uninstall

updated

Crilum:
>with that said, I'd be fine with adding a subcategory under internet for `Communications` or as its own primary category. lets see what everyone else thinks

Yeah, a subcategory would probably be better.

Crilum:
> Terminals and Engineering may need tweaking for better perception.

I assume you're talking about the icons:
I think Engineering looks fine, although the Terminals icon looks a little weird... I'm not quite sure why, but I think the prompt is too big...

Sussy-OS:
I agree, the prompt is a little too big.

Jai-JAP:
> > Here's the icon:
> > 
> > uninstall
> 
> updated, thanks

I think it will look better if the 3 lines in the bin are still light colored

Botspot:
Terminal and Engineering icons have been updated, and I hand-drew an emulation icon as I couldn't find anything suitable online.

theofficialgman:
https://github.com/Botspot/pi-apps/pull/1580#issuecomment-1127675913
what does everyone else think about this? should it be left as is, should there be a primary communications category, or should there be a communications category under internet?

Sussy-OS:
> [#1580 (comment)](https://github.com/Botspot/pi-apps/pull/1580#issuecomment-1127675913) what does everyone else think about this? should it be left as is, should there be a primary communications category, or should there be a communications category under internet?

Sub-category under internet.

Crilum:
@Botspot I like the Emulation icon!

Sussy-OS:
I like it as well!

Sussy-OS:
> > > > Here's the icon:
> > > > uninstall
> > > 
> > > 
> > > updated, thanks
> > 
> > 
> > I think it will look better if the 3 lines in the bin are still light colored 
> 
> I don't know...
> 
> image vs. image

Deffinitely bottom one.

Sussy-OS:
Good idea though!

theofficialgman:
I'm for the second one as well. being able to see exactly my gtk dark color theme in the background I think it good. all other icons have holes in them to allow for this

Crilum:
@Botspot is there a reason why `Back.png` and `Back-64.png` are different?
image
`Back-64.png` on the left, `Back.png` on the right.

Crilum:
> > I'm for the second one as well. being able to see exactly my gtk dark color theme in the background I think it good. all other icons have transparent holes in them to allow for this
> 
> The original reason why I had a white border around the Uninstall icon was to improve contrast for dark modes. This doesn't happen on all monitors, but on mine there is some sort of signal compression on 1080p. All pixels are sharp, but if two similar colors are near each other horizontally, some blending will occur. This is especially noticeable with single-pixel details, like the stripes on the trash icon. To show you guys what I see, I made a picture of how my monitor displays an earlier comment on this PR. 2022-05-16-103602_1920x1080_scrot

Maybe we should do something like @Jai-JAP was saying, and put a color in the lines. The icon with white lines looks *way* better in your screenshot.

Botspot:
> @Botspot is there a reason why `Back.png` and `Back-64.png` are different? image `Back-64.png` on the left, `Back.png` on the right.

Back-64 is only used in Xlunch gui, and it fits better with the pastel aesthetic of the xlunch-light-3d theme.

theofficialgman:
> 2022-05-16-103602_1920x1080_scrot

I think thats your theme adding a shadow to any non-black color, not your monitor. I agree your picture on the right looks like trash (pun intended)

mine:
Screenshot from 2022-05-16 11-42-24

Crilum:
> > 2022-05-16-103602_1920x1080_scrot
> 
> I think thats your theme adding a shadow to any non-black color, not your monitor. I agree your picture on the right looks like trash (pun intended)
> 
> mine: Screenshot from 2022-05-16 11-42-24

Yeah, the transparent lines on my system look really good.

Crilum:
The back icon doesn't look as good (in my opinion) without the white arrow:
image
image

Sussy-OS:
> > I think thats your theme adding a shadow to any non-black color, not your monitor.
> 
> No, it's definitely my monitor. Viewing the Uninstall icon with a black background is even worse. uninstall My monitor seems to struggle with blue the most.

Wow, that's bad.

theofficialgman:
> @theofficialgman That looks really cool, what theme is that?

its ubuntu budgie on bionic. by default it uses the Pocillo theme for icons and Pocillo-dark-slim for widgets.
I honestly haven't touched it, it just looks like this
Screenshot from 2022-05-16 12-09-45

Botspot:
In any case, I think this is ready to merge.

theofficialgman:
> I thought we fixed the terminal logo for you.

no, you check for 11 or lower still. this is 18.04
```
  #use simpler logo if OS is Buster or lower - to fix issue https://github.com/Botspot/pi-apps/issues/1441
  local version_id="$(grep 'VERSION_ID=' /etc/os-release | tr -cd '0123456789')"
  if [[ "$version_id" != *\.* ]] && [ "$version_id" -ge 11 ];then
```

the simple logo works, you just don't use it

Sussy-OS:
Merge it! :D

Botspot:
> ```
>   if [[ "$version_id" != *\.* ]] && [ "$version_id" -ge 11 ];then
> ```

It should have matched the decimal point and used the fallback logo. `[[ "$version_id" != *\.* ]]`

theofficialgman:
I'll let botspot pull the trigger.

Sussy-OS:
Whooh!

theofficialgman:
> > ```
> >   if [[ "$version_id" != *\.* ]] && [ "$version_id" -ge 11 ];then
> > ```
> 
> It should have matched the decimal point and used the fallback logo. `[[ "$version_id" != *\.* ]]`

if I run it outside of pi-apps, it does... if I do it in pi-apps, it does the first logo
wth

Sussy-OS:
Side-by-side comparison:
image
image

Sussy-OS:
As updating, I encountered this..
image
I do not see any errors, but I was wondering what it is.

Crilum:
Just updated Pi-Apps... There's so many updated files, lol
<details>
$ update -p

Welcome to Crilum's Update Script!

Pi-Apps output:

Updater mode: cli-yes

Checking for online changes... Done
Scanning apps... Done
Scanning files... Done

These apps can be updated:

  - Firefox Rapid Release
  - Snapdrop

These files can be updated:

  - api
  - CHANGELOG.md
  - createapp
  - DOCUMENTATION.md
  - etc/categories
  - etc/logviewer
  - etc/preload-daemon
  - gui
  - icons/categories/Communication-64.png
  - icons/categories/Communication.png
  - icons/categories/Creative Arts-64.png
  - icons/categories/Creative Arts.png
  - icons/categories/Emulation-64.png
  - icons/categories/Emulation.png
  - icons/categories/Engineering-64.png
  - icons/categories/Engineering.png
  - icons/categories/Office-64.png
  - icons/categories/Office.png
  - icons/categories/Packages-64.png
  - icons/categories/Packages.png
  - icons/categories/Programming-64.png
  - icons/categories/Programming.png
  - icons/categories/System Management-64.png
  - icons/categories/System Management.png
  - icons/categories/Terminals-64.png
  - icons/categories/Terminals.png
  - icons/categories/vector/Communication.svg
  - icons/categories/vector/Creative Arts.svg
  - icons/categories/vector/Emulation.svg
  - icons/categories/vector/Engineering.svg
  - icons/categories/vector/Office.svg
  - icons/categories/vector/Packages.svg
  - icons/categories/vector/Programming.svg
  - icons/categories/vector/System Management.svg
  - icons/categories/vector/Terminals.svg
  - icons/failure.png
  - icons/install.png
  - icons/multi-select.png
  - icons/options.png
  - icons/prompt.png
  - icons/screenshots/app list.png
  - icons/screenshots/buttons/install.png
  - icons/screenshots/buttons/settings.png
  - icons/screenshots/buttons/uninstall.png
  - icons/screenshots/details window.png
  - icons/screenshots/main window.png
  - icons/settings-dark.png
  - icons/settings-light.png
  - icons/settings.png
  - icons/success.png
  - icons/uninstall.png
  - icons/update.png
  - icons/vector/failure.svg
  - icons/vector/options.svg
  - icons/vector/settings-dark.svg
  - icons/vector/settings-light.svg
  - icons/vector/settings.svg
  - icons/vector/success.svg
  - icons/wait.png
  - install
  - manage
  - preload
  - README.md
  - settings
  - updater


api file was copied successfully.
CHANGELOG.md file was copied successfully.
createapp file was copied successfully.
DOCUMENTATION.md file was copied successfully.
etc/categories file was copied successfully.
etc/logviewer file was copied successfully.
etc/preload-daemon file was copied successfully.
gui file was copied successfully.
icons/categories/Communication-64.png file was copied successfully.
icons/categories/Communication.png file was copied successfully.
icons/categories/Creative Arts-64.png file was copied successfully.
icons/categories/Creative Arts.png file was copied successfully.
icons/categories/Emulation-64.png file was copied successfully.
icons/categories/Emulation.png file was copied successfully.
icons/categories/Engineering-64.png file was copied successfully.
icons/categories/Engineering.png file was copied successfully.
icons/categories/Office-64.png file was copied successfully.
icons/categories/Office.png file was copied successfully.
icons/categories/Packages-64.png file was copied successfully.
icons/categories/Packages.png file was copied successfully.
icons/categories/Programming-64.png file was copied successfully.
icons/categories/Programming.png file was copied successfully.
icons/categories/System Management-64.png file was copied successfully.
icons/categories/System Management.png file was copied successfully.
icons/categories/Terminals-64.png file was copied successfully.
icons/categories/Terminals.png file was copied successfully.
icons/categories/vector/Communication.svg file was copied successfully.
icons/categories/vector/Creative Arts.svg file was copied successfully.
icons/categories/vector/Emulation.svg file was copied successfully.
icons/categories/vector/Engineering.svg file was copied successfully.
icons/categories/vector/Office.svg file was copied successfully.
icons/categories/vector/Packages.svg file was copied successfully.
icons/categories/vector/Programming.svg file was copied successfully.
icons/categories/vector/System Management.svg file was copied successfully.
icons/categories/vector/Terminals.svg file was copied successfully.
icons/failure.png file was copied successfully.
icons/install.png file was copied successfully.
icons/multi-select.png file was copied successfully.
icons/options.png file was copied successfully.
icons/prompt.png file was copied successfully.
icons/screenshots/app list.png file was copied successfully.
icons/screenshots/buttons/install.png file was copied successfully.
icons/screenshots/buttons/settings.png file was copied successfully.
icons/screenshots/buttons/uninstall.png file was copied successfully.
icons/screenshots/details window.png file was copied successfully.
icons/screenshots/main window.png file was copied successfully.
icons/settings-dark.png file was copied successfully.
icons/settings-light.png file was copied successfully.
icons/settings.png file was copied successfully.
icons/success.png file was copied successfully.
icons/uninstall.png file was copied successfully.
icons/update.png file was copied successfully.
icons/vector/failure.svg file was copied successfully.
icons/vector/options.svg file was copied successfully.
icons/vector/settings-dark.svg file was copied successfully.
icons/vector/settings-light.svg file was copied successfully.
icons/vector/settings.svg file was copied successfully.
icons/vector/success.svg file was copied successfully.
icons/wait.png file was copied successfully.
install file was copied successfully.
manage file was copied successfully.
preload file was copied successfully.
README.md file was copied successfully.
settings file was copied successfully.
updater file was copied successfully.
Updating Firefox Rapid Release...

Firefox Rapid Release was updated successfully.
Updating Snapdrop...

Snapdrop was updated successfully.

Pi-Apps updates complete.
```
</details>

@Sussy-OS I didn't get that...

**Comment by Crilum** [comment]
@Sussy-OS Did you have Pi-Apps open when you updated it?

**Comment by Sussy-OS** [comment]
> Just updated Pi-Apps... There's so many updated files, lol
> 
> @Sussy-OS I didn't get that...

Hmm..

**Comment by theofficialgman** [comment]
>  do not see any errors, but I was wondering what it is.

the old gui is incompatible with the new gui. previously pi-apps launched the old gui after updating, causing visual bugs since the lists formats changed with this PR. this kills it and starts the new one.

the runonce to do that outputs the terminal which is visible in the update output.

> @Sussy-OS I didn't get that...

you only get it if updating via the gui, since thats only when its required

**Comment by Sussy-OS** [comment]
> @Sussy-OS Did you have Pi-Apps open when you updated it?

No, just closed it. I ran `pi-apps update` by the BASH extension

**Comment by Sussy-OS** [comment]
> you only get it if updating via the gui, since thats only when its required

But, I updated by `pi-apps update`

**Comment by theofficialgman** [comment]
> > you only get it if updating via the gui, since thats only when its required
> 
> But, I updated by `pi-apps update`

rigth, the new gui didn't launch then. but the runonce still executed and passed

**Comment by Sussy-OS** [comment]
@Botspot How would you select multiple apps now? I can't seem to hold Ctrl, or Shift with it working.

**Comment by theofficialgman** [comment]
> @Botspot How would you select multiple apps now? I can't seem to hold Ctrl, or Shift with it working.

there is a new multi-select gui accessible from the settings menu
multi-select from the new gui was no longer possible

**Comment by Crilum** [comment]
> > @Botspot How would you select multiple apps now? I can't seem to hold Ctrl, or Shift with it working.
> 
> there is a new multi-select gui accessible from the settings menu multi-select from the new gui was no longer possible

Has the docmentaion/README been updated to show things like this?

**Comment by theofficialgman** [comment]
multi-select was never a documented feature before.

the readme does need updating though, the settings window pictured is old, looks like we missed that

**Comment by Sussy-OS** [comment]
New issue:
image

**Comment by Sussy-OS** [comment]
This is while multi-selecting.

**Comment by Sussy-OS** [comment]
> Looks like somewhere there's something that needs to be quoted... Or something of a similar nature

Yeah..

**Comment by Sussy-OS** [comment]
Looks like it is..
image

**Comment by Crilum** [comment]
```
    🭊▅▅▅▅🬿
 🭈▂▂█▂▂▂▂█▂▂🬽
 ▌  ▗▖▗▖▗▖  ▐   █▀▀🭍 ▄
 ▌  ▗▖▗▖▗▖  ▐   █▄▄🭞 ▄ ▄▄ 🭂▀▀█ █▀▀🭍 █▀▀🭍 🭂🬰🬰🬰
 ▌  ▗▖▗▖▗▖  ▐   █    █    🭓▄▄█ █▄▄🭞 █▄▄🭞 ▄▄▄🭞
 🭓🬭🬭🬰🬰🬰🬰🬰🬰🬭🬭🭞                  █    █

/home/pi/pi-apps/apps/Alacritty does not exist!
/home/pi/pi-apps/apps/Terminal does not exist!

Close this window to exit.
```
Just installing one app doesn't work either.

Sussy-OS:
0_0

Crilum:
That should quote apps...

Sussy-OS:
Starts here, if anybody wants to see what is the matter: https://github.com/Botspot/pi-apps/blob/master/api#L1841

theofficialgman:
I'm checking for the bug right now
this is already correct `"${DIRECTORY}/manage" multi-install "'"$apps"'"` it is the multi-install function in the manage script that is the issue

Sussy-OS:
> I'm checking for the bug right now this is already correct `"${DIRECTORY}/manage" multi-install "'"$apps"'"` it is the multi-install function in the manage script that is the issue
> 
> ```
> garrett@garrett-usb:~/pi-apps$ DIRECTORY=/home/garrett/pi-apps
> garrett@garrett-usb:~/pi-apps$ source api
> garrett@garrett-usb:~$ cd pi-apps
> garrett@garrett-usb:~/pi-apps$ "${DIRECTORY}/manage" multi-install "Angry IP scanner" "btop++"
> /home/garrett/pi-apps/apps/Angry does not exist!
> /home/garrett/pi-apps/apps/IP does not exist!
> /home/garrett/pi-apps/apps/scanner does not exist!
> ```

Yes, but I can not seem to find where that begins (in the manage file).

Sussy-OS:
@theofficialgman How is it going with fixing the bug?

theofficialgman:
@Sussy-OS sorry took a while to figure out exactly how it needed to be fixed (I knew what was wrong but didn't know how to fix it). done now with https://github.com/Botspot/pi-apps/commit/b6d71aaf56fcf2ed200aa1cf215a64120488e240

Sussy-OS:
How do you multi-uninstall?

Sussy-OS:
> alright the waitlist and a fallback incase the waitlist disappears are implemented.
> 
> Now the install and uninstall commands from the gui will NOT start unless the manage script is not currently running AND the script is next in line in the waitlist (ie: the app name is at the top of the file).
> 
> this makes is so the app install/uninstalls will proceed in the order of the users execution. the terminal run window will popup and show a spinning icon and text saying "Waiting for another script to finish before continuing..." until the previous script finishes
> 
> @Botspot how does this look now?
> 
> example screenshot of multiple uninstalls queued up: Screenshot from 2022-04-08 20-36-32

I wish this got added..

theofficialgman:
@Sussy-OS you are misunderstanding how pi-apps has improved. use the normal GUI and select and install an app. then keep browsing the GUI (while the other app hasn't finished) and add another app to install or uninstall, they get added to a queue

Sussy-OS:
> @Sussy-OS you are misunderstanding how pi-apps has improved. use the normal GUI and select and install an app. then keep browsing the GUI (while the other app hasn't finished) and add another app to install or uninstall, they get added to a queue

Ah! That's amazing! I do like your concept of the terminals themselfs waiting, but the GUI queue is amazing! Maybe you could add the terminal queue to the bash extension? But I do not understand why there is a multi-install, and not a multi-uninstall. It would be so much easier to just select all uninstall in one go.

Crilum:
>Maybe you could add the terminal queue to the bash extension?

Or in Pi-Apps Settings.

Sussy-OS:
> > Maybe you could add the terminal queue to the bash extension?
> 
> Or in Pi-Apps Settings.

Yeah, to enable GUI, or Terminal queue.

theofficialgman:
@Sussy-OS added

Sussy-OS:
Yeah, I saw it in manage, but I was wondering why it was not in the settings.

PR #1593, opened 2022-03-13
update balenaEtcher to v1.7.7

ryanfortner:


Botspot:
Thanks! Would you like to add an updater script so this will happen automatically next time?

ryanfortner:
> Thanks! Would you like to add an updater script so this will happen automatically next time?

Sure, but balenaEtcher debs have a commit id after their version number in the filename. I think that will require a little more tweaking :)

Issue #1592, opened 2022-03-12
OBS Studio Unable to run on Bullseye 32/64bit


theofficialgman:
### Confirmations

- [X] I searched for similar bug reports in https://github.com/Botspot/pi-apps/issues and found none were relevant.
- [X] I tried updating Pi Apps and the issue is still not fixed. [How to update Pi Apps?](https://github.com/Botspot/pi-apps#to-update-pi-apps)

### What happened?

OBS deb is built on buster and requires libpython3.7m.so.1.0 when executing `obs`. This is not available in bullseye as it uses python3.9. The deb itself installs (which is why there have been no error messages) but the software does not run.

### Description

<img width="548" alt="Screenshot 2022-03-12 180708" src="https://user-images.githubusercontent.com/28281419/158038109-9ac68c15-d996-460c-90b3-60fc94e78518.png">

@ryanfortner A deb for OBS studio will need to be built directly on bullseye for 32/64bit in addition to the current buster 32/64bit deb.

### What is your operating system?

Bullseye 32bit and 64bit Raspberry PIOS Releases

ryanfortner:
Ill add this to my to-do list, should be able to compile soon.

ryanfortner:
Here is an arm64 bullseye deb I just compiled, I tested it and it works. https://github.com/ryanfortner/ryanfortner/releases/download/1002/obs-studio_27.2.0-101-g358ef281c_arm64.deb

theofficialgman:
@ryanfortner were you able to compile the armhf bullseye deb as well?
I've got confirmation from a user that the `libxcb-util.so.0` library hack is also no longer necessary since you compiled on native bullseye, so I will be removing that as well

ryanfortner:
> @ryanfortner were you able to compile the armhf bullseye deb as well? I've got confirmation from a user that the `libxcb-util.so.0` library hack is also no longer necessary since you compiled on native bullseye, so I will be removing that as well

Yep, just compiled armhf version and updated the install-32 script. Also, I changed the scripts so that the library hack is only run on bullseye installs.

theofficialgman:
no its not necessary at all, I went ahead and added my own commit and pushed
your PR had multiple issues (incorrect links, etc) but you did upload the debs to your github so all is fine now
https://github.com/Botspot/pi-apps/commit/0112d125eea852cfe751034cbbe48050f4ca3cca

ryanfortner:
Thanks!

Crilum:
[...]

PR #1594, opened 2022-03-13
add obs-studio bullseye support

ryanfortner:
for #1592


theofficialgman:
I don't expect this deb to also work on buster. I'm fine with having a unique deb per version and can add that to the install script once they are tested

ryanfortner:
Unfortunately I've lost the buster arm64 deb while adding the bullseye one to my repo, so I'll have to compile that from scratch again. In the meantime, can this pr be merged since the old link is no longer valid?

theofficialgman:
no I won't do that, since that would update all installed users to a broken version if they are on buster
you forget about your mirrors, sourceforge isn't syncing so its perfect and I grabbed the lost releases
here is a zip containing the armhf and arm64 releases
[obs-studio_27.2.0-backup.zip](https://github.com/Botspot/pi-apps/files/8238838/obs-studio_27.2.0-backup.zip)

ryanfortner:
Whoops, I forgot about the mirrors :laughing: thanks for reminding me

I've added back the old debs now

ryanfortner:
@theofficialgman I made some changes to `install-64`, could you please take a look?

theofficialgman:
> @theofficialgman I made some changes to `install-64`, could you please take a look?

can you un-do the hack which was done here locally on your install and see if it still works now that we have bullseye specific debs?
```
#temporary fix of `error while loading shared libraries: libxcb-util.so.0: cannot open shared object file: No such file or directory`
if ! package_installed libxcb-util0 ;then
  libpath="$(dirname "$(dpkg -S libxcb-util.so.1 | grep "libxcb-util.so.1$" | awk '{print $2}')")" #Value on RPiOS Bullseye: /usr/lib/arm-linux-gnueabihf
  if [ ! -e "${libpath}/libxcb-util.so.0" ] ;then
    status "Fixing libxcb-util.so.0 library"
    sudo ln -sf "${libpath}/libxcb-util.so.1" "${libpath}/libxcb-util.so.0"
  fi
fi
```
`libxcb-util.so.0` is what is supposed to not exist in bullseye but the hack "adds it back" just remove it and see if obs still launches

ryanfortner:
@theofficialgman everything good so far?

theofficialgman:
closed with https://github.com/Botspot/pi-apps/commit/0112d125eea852cfe751034cbbe48050f4ca3cca

PR #1601, opened 2022-03-15
Add Splash CLI

Crilum:
### [splash-cli](https://github.com/splash-cli/splash-cli) 
is a CLI tool that gets an image from Unsplash and sets it as your background.
You can also specify what image ID you want, get the image of the day, specify collections for splash to pick from, and more.

Just type `splash`!


apps/Splash CLI/install
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+install_packages npm || exit 1
```
cycool29, line 3: You should install the Node.js app from pi-apps for `npm`.
```suggestion
"${DIRECTORY}/manage" install-if-not-installed Node.js || error "Node.js failed to install somehow!"```

Crilum: Does it contain NPM?

ryanfortner:
there is also [unsplash-wallpapers](https://github.com/soroushchehresa/unsplash-wallpapers), which is available in raspbian addons: https://apt.raspbian-addons.org/debian/pool/main/u/unsplash-wallpapers/ (this is a completely different app but I figured I might as well post it)

Crilum:
> there is also [unsplash-wallpapers](https://github.com/soroushchehresa/unsplash-wallpapers), which is available in raspbian addons: https://apt.raspbian-addons.org/debian/pool/main/u/unsplash-wallpapers/ (this is a completely different app which has a gui, but I figured I might as well post it)

Mm, it looks better than splash-cli, but it looks like it would violate the terms of Unsplash's API:

image


I suppose that would also mean splash-cli is breaking the terms, too...

apps/Splash CLI/install
cycool29: > Does the Node.js app contain NPM?

Yes.

Crilum:
Closing for liability.

Botspot:
> I suppose that would also mean splash-cli is breaking the terms, too...

Maybe they have permission.

Crilum:
Well, if you're good with it, then go ahead and add it if want. If you do want to add it, I can write an update script.

PR #1617, opened 2022-03-21
Fixed Powerline Shell

open-sorcerer64:


cycool29:
He copied [the install script](https://raw.githubusercontent.com/techcoder20/RPI-PowerlineShell-Installer/main/install.sh) and uninstall Starship Prompt before installing.

@open-sorcerer64  BTW, you should not uninstall apps without user permission.

open-sorcerer64:
> What was wrong with the current script? (What did you fix?)

Starship and powerline can not be installed together on bash

Botspot:
> Starship and powerline can not be installed together on bash

That makes sense. For transparency, I think that it would be better to notify the user about the problem, rather than secretly making changes which may confuse the user.
I suggest using an error message to notify the user that "another terminal theme is already installed. Please uninstall that one first" (see the conky app)

theofficialgman:
> That makes sense. For transparency, I think that it would be better to notify the user about the problem, rather than secretly making changes which may confuse the user. I suggest using an error message to notify the user that "another terminal theme is already installed. Please uninstall that one first" (see the conky app)

on this note, it might be a good idea to make this a user error as well `User error: ` so that the upload log button is hidden

open-sorcerer64:
Ok then, I will update the scripts! Also please check Starship while you guys are at it

Botspot:
Any updates to this issue, @open-sorcerer64?

PR #1629, opened 2022-03-22
add system-monitoring-center

ryanfortner:


theofficialgman:
this package can not be installed on debian buster or ubuntu bionic

```
 system-monitoring-center : Depends: bash (>= 5.0) but 4.4.18-2ubuntu1.2 is to be installed
                            Depends: pci.ids but it is not installable
                            Depends: python3 (>= 3.7) but 3.6.7-1~18.04 is to be installed
                            Depends: util-linux (>= 2.33) but 2.31.1-0.4ubuntu3.7 is to be installed
```
bionic output ^

buster is missing only pci.ids

theofficialgman:
tracking with upstream bug: https://github.com/hakandundar34coding/system-monitoring-center/issues/40

theofficialgman:
the debian controls file has been fixed upstream which allows this to be built and installed on debian buster and ubuntu bionic.
there are still a few more bugs being tracked in my issue, notibly: support for USB network cards, GPU vendor/model info causing a crash, and disk mountpoint incorrect reporting.

no new release has been made yet and won't be made till these issues are fixed. the upstream developer has been very fast to resolve my other reported issues.

I will send another message when the upstream deb has been updated so users can test on buster

theofficialgman:
the upstream issues have all been fixed and v1.8.0 installs and runs without issue

theofficialgman:
@Botspot I'm requesting you test this. It should be good to merge and I think its quite a nice system monitor

Botspot:
> @Botspot I'm requesting you test this. It should be good to merge and I think its quite a nice system monitor

Initial impressions were good, but I soon encountered some rough areas that appear to have room for improvement:

It seems like more information about the RAM could be collected.
2022-03-29-184601_1920x1080_scrot
The symbol legend for the graph is not clear. I'm not sure which data value corresponds to which type of line on the graph.
Also some info is missing for the disk.
2022-03-29-184905_1920x1080_scrot
GPU tab is missing more info than I think it could have. On a Pi4, I don't think all of the info is available, but the OpenGL version should be present at the very least. Also this tab was a bit temperamental and flickered. (maybe due to polling the gpu for fps readings)
2022-03-29-190139_1920x1080_scrot
The symbol legend for the graph is not clear. I'm not sure which data value corresponds to which type of line on the graph.
2022-03-29-185117_1920x1080_scrot

Also the UI seemed a bit unstable, and I managed to capture a couple of instances of it not fully rendering. This is probably not the developer's fault, but an unstable implementation of GTK3 + hardware acceleration from the RPi team.
2022-03-29-185329_1920x1080_scrot
2022-03-29-185337_1920x1080_scrot

Overall, it looks good and I've never seen some of these features in a GUI before. Just there's a few places where RPi-specific changes in /sys and /dev need to be accommodated to gather all of the stats.

theofficialgman:
> Overall, it looks good and I've never seen some of these features in a GUI before. Just there's a few places where RPi-specific changes in /sys and /dev need to be accommodated to gather all of the stats.

@Botspot 
I'm going to comment on each of these (because I have no issues on switch now that I had them fixed upstream😆)
for one, make sure you are using v1.8.0 (the latest release)

the ram info missing is a known issue, the current method relies on ACPI (used on most desktop computers)
some of this info could be obtained but most isn't available for SBCs. The "Physical Ram: [Unknown]" is a bug with the pi only, that info should be available

disk read and write speeds isn't a bug, thats how you read the graph. - solid line for read, - - dashed line for writes
your [ not mounted ] issue is strange, I have never seen it on any of my systems, you will have to file a bug yourself and communicate with the dev (they are really good and helpful)

the gpu info is about as good as it gets, just be glad you got the info you have already, there was none when I started 😆 
the "refresh rate" is how fast the gpu is able to draw that window, think of it like a benchmark, I know its strange but the pi gpu is clearly struggling here. future version of this app will likely have load and frequency data plots instead of fps (or on a toggle).
missing.
OpenGL version info is because you probably don't have `mesa-utils` installed, its an optional dependency but should be added to get that

the network speed tab is also probably as good as it gets, the pi wifi driver is not giving this info to userspace, if you know of a place where the info is available, file a bug on the upstream github

the strange rendering issues are almost 100% raspberry pi gpu driver faults, I have had no issues on switch or my various x86_64 linux laptops with amd and intel mesa drivers

theofficialgman:
also for network, you should click on the gear and look at some of the other available devices under "selected device"
maybe the automatic one is wrong and should be hidden

theofficialgman:
just for reference this is what everything is supposed to look like

<details>

Screenshot from 2022-03-29 21-47-11
Screenshot from 2022-03-29 21-47-45
Screenshot from 2022-03-29 21-47-52

Screenshot from 2022-03-29 21-55-46

Screenshot from 2022-03-29 21-48-07
Screenshot from 2022-03-29 21-51-00
Screenshot from 2022-03-29 21-47-55

yes I drew black boxes around my info, it is populated though

theofficialgman:
marking this a a draft as releases are no longer made via github with DEBs but instead uses pip for installation.
.desktop files are now not generated on install, it requires one start of the software via CLI

theofficialgman:
upstream now says they will continue to package deb at least occasionally
so this PR has been updated for the new format and marked as ready for review again

theofficialgman:
@Botspot I suggest you try this again
I think most of your issues (if not all) may have been fixed

Botspot:
> @Botspot I suggest you try this again

@hakandundar34coding
I've tested it again and there was some improvement.
The Network tab had the most improvement, now displaying the ipv4, ipv6, and MAC addresses correctly. Connection Type is still blank, yet it now correctly detects `eth0` to be my connection name and "Bcm2711- Genet-V5" as the device name.

The GPU tab seems unchanged, except that the FPS reading is now around 60.
The RAM Hardware pop-up is as void of information as before.
The only change to the Disk tab is that the dash characters were separated better in the graph legend. It still doesn't know the free space, used space, capacity, and Disk Usage.

hakandundar34coding:
@Botspot 
I know the problems that you write here. I did not want to use this issue for soving problems of a separate application.
I wrote the required outputs in [this](https://github.com/hakandundar34coding/system-monitoring-center/issues/40#issuecomment-1082679424) issue.

About the GPU tab:
This tab is redesigned and glarea widget of GTK3 is removed. Visual problems are fixed. Source code is not uploaded yet.
But a lot of areas for ARM GPUs will be empty.

About Physical RAM:

These files/folders are required for this feature but kernel version affects that (for ARM devices):
```
/sys/devices/system/memory/block_size_bytes
/sys/devices/system/memory/
```

I saw other RB-Pi/ARM devices that do not have block_size_bytes file. But this information can be get on N.Switch devices.

About the Disk tab:
Can you share outputs of these commands:
```
/proc/diskstats
/proc/mounts
/proc/cmdline    (you can remove sensitive information, UUID or PARTUUID information is required.)```

Botspot:
(reply has been sent to https://github.com/hakandundar34coding/system-monitoring-center/issues/40)

theofficialgman:
@Botspot @ryanfortner someone will a RPi needs to test the lastest package.  many changes have been made and feedback is necessary.

hakandundar34coding already asked for this feedback:

> This tab is redesigned and GPU usage, power, frequencies, memory, etc. information is shown. But these information can not be get for all GPU vendors.
> Is there any way to get any of these information on ARM devices (for example RB-Pi devices)?

I can not assist because I don't have a PI and already all my devices work (intel, nvidia desktop close source, nvidia desktop open source, amd, and nvidia tegra/jetsons)

Botspot:
It's nearly flawless. Remaining changes will require parsing `vcgencmd` and will be more difficult, so to avoid delaying this more than necessary, I say we merge it now.

theofficialgman:
sounds good, merged

PR #1630, opened 2022-03-22
update balenaEtcher to v1.7.8

ryanfortner:


theofficialgman:
@ryanfortner 
I get this when installing the deb `chmod: cannot access '/opt/balena-etcher-electron/chrome-sandbox': No such file or directory`

I noticed that balena-etcher-electron does not exist, instead the folder balenaEtcher exists

theofficialgman:
even after fixing the permissions of chrome-sandbox manually, I just get a while window when opening the application
maybe the version of electron 12.2.3 is broken on some systems? maybe etcher should upgrade from `8.9.255.25-electron.0` (like used now) to `9.8.177.11-electron.0` (electron 17.0.0)

can you reproduce?

ryanfortner:
If you ignore this warning, does etcher still launch? It does for me. 

Also, on futurejones' repo, it's noted that you can ignore the warning because it applies to old electron versions. https://github.com/futurejones/balena-etcher-arm/blob/main/etcher-build/README.md

theofficialgman:
> If you ignore this warning, does etcher still launch? It does for me.
> 
> Also, on futurejones' repo, it's noted that you can ignore the warning because it applies to old electron versions. https://github.com/futurejones/balena-etcher-arm/blob/main/etcher-build/README.md

it does not launch, because the owership of chrome-sandbox is not set correctly (its not root owned and not 4755). The failed chmod is trying to do that, but the path is wrong. see upstream bug in previous post ^

```
The SUID sandbox helper binary was found, but is not configured correctly.
Rather than run without sandboxing I'm aborting now. You need to make sure that chrome-sandbox is owned by root and has mode 4755.
```

ryanfortner:
image

theofficialgman:
> image

🤷🏻‍♂️ maybe piOS is now overwritting everything with --no-sandbox? idk, regardless its a known issue, but not something thats really on you to fix

PR #1631, opened 2022-03-22
cura: remove xcbutil linking

ryanfortner:
With the latest build of Cura this line is not required.


ryanfortner:
At least on Debian Buster it isn't... I'll try with Bullseye in a bit.

theofficialgman:
> At least on Debian Buster it isn't... I'll try with Bullseye in a bit.

it never has been required on debian buster, the symlink always just errored and did not do anything on buster since `ibxcb-util.so.1.0.0` does not exist

buster includes `libxcb-util0 ` which gives the lib `libxcb-util.so.0`
bullseye includes `libxcb-util1` which gives the lib `libxcb-util.so.1.0.0`

smartavionics builds on debian buster and now does include this library `libxcb-util.so.0` in the appimage itself, hence why the dynamic linking works without the symlink hack on bullseye

PR #1638, opened 2022-03-24
keepassxc: buster/bullseye specific debs

ryanfortner:


ryanfortner:
@theofficialgman if everything looks good, I can go ahead and merge this.

theofficialgman:
thats fine for now.
though I'd really prefer at some point in the future if you fix your raspbian addons repo to actually use the `dists` properly instead of putting everything under precise

that would let you have all the debs on that one repo

apps/KeePassXC/install-64
```diff
@@ -1,5 +1,11 @@
...
-version=2.6.6-5
+version=2.6.6-1
 
-install_packages https://apt.raspbian-addons.org/debian/pool/main/k/keepassxc/keepassxc_${version}_arm64.deb || exit 1
+if [ "$(get_codename)" == "bullseye" ]; then
+    #install bullseye version
+    install_packages https://apt.raspbian-addons.org/debian/pool/main/k/keepassxc/keepassxc_${version}_arm64.deb || exit 1
+else 
+    #install buster version
+    install_packages https://github.com/ryanfortner/ryanfortner/releases/download/1002/keepassxc_2.6.6-5_arm64.deb || exit 1
```
Botspot, line 10: @ryanfortner, do you want a version variable added to this URL?

theofficialgman: no, there isn't any way to get the latest version number from ryanforner's github so there isn't any reason to specify it with a variable. They made one release a long time ago and then just update/add files to it as necessary. https://github.com/ryanfortner/ryanfortner/releases its clear if you look at the page

Issue #1620, opened 2022-03-21
App Submission: Syncthing


edbernard:
Hi there, I found an app called syncthing a while back and found it to be super useful, I was wondering if you could put it on pi-apps.
[Syncthing.zip](https://github.com/Botspot/pi-apps/files/8317909/Syncthing.zip)

theofficialgman:
going to add some context here for other people:
website: https://syncthing.net/
github: https://github.com/syncthing/syncthing
Description:
> Syncthing is a continuous file synchronization program. It synchronizes files between two or more computers in real time, safely protected from prying eyes. Your data is your data alone and you deserve to choose where it is stored, whether it is shared with some third party, and how it's transmitted over the internet.


the user above submitted a package app but I don't think that is what we should use.

syncthing has updated binary builds for many systems (armhf and arm64 linux included) on their website https://syncthing.net/downloads/ and directly from the github releases https://github.com/syncthing/syncthing/releases/tag/v1.19.1. the debian buster and bullseye official packages of syncthing are quite old, hence why I wouldn't suggest them. Assuming these builds work on both buster and bullseye, this would be a perfect application for the auto update github actions.

theofficialgman:
there is also an official apt repo, but because it doesn't have an index file, its not really useable in our case unless we want to install the apt repo and not handle version management `https://apt.syncthing.net/`

cycool29:
> there is also an official apt repo, but because it doesn't have an index file, its not really useable in our case unless we want to install the apt repo and not handle version management `https://apt.syncthing.net/`

We can manually grep its `Packages` file.
```bash
wget -qO- wget -qO- https://apt.syncthing.net/dists/syncthing/stable/binary-armhf/Packages | sed -z 's/Package: syncthing-discosrv.*//g' | grep 'Version:' | tail -n 1 | sed 's/.*: //g'
```

theofficialgman:
relying on the order of the pacakges file sounds like a bad idea, so is relying on the placement of the syncthing inside the packages file relative to the other packages....

it would be nice if there was some tool that could read the Packages file directly and output the highest version number of a specified package within it

theofficialgman:
```
wget -qO- https://apt.syncthing.net/dists/syncthing/stable/binary-armhf/Packages | awk '/Package: syncthing\n/' RS= | grep "Version:" | awk '{print $2}' | sort -V | tail -n1
```
and here is that script
[...]

PR #1641, opened 2022-03-25
Github Actions: Debian package updater

theofficialgman:
this is a complex helper script designed to make automatically updating packages from online debian repos easy. simply set a couple of variables in your install/install-32/install-64 script and create a matching script in the updates folder and let the github actions take care of the rest

I have added the new "Syncthing" app to showcase this script (not added to the categories list yet). I have tested the github actions and you can see the run of it here https://github.com/theofficialgman/pi-apps/pull/5.
As with the `update_github_script.sh` helper script, you can specify multiple `webPackages` and `packagename` in arrary form and the script will check each for updates.

Documentation for both `update_github_script.sh` and `update_debian_repo_script.sh` will come at some point in the future.

closes: https://github.com/Botspot/pi-apps/issues/1620


theofficialgman:
note, the filepaths variables in Syncthing are specifically wrong and set to 
```
filepath_32="junk"
filepath_64="junk"
```
to show off the github actions script working on the first run

theofficialgman:
Also, you may wonder why use this for debian repos over the current `update_github_script.sh`.
This script is immune to changes in `Filepath:` on the debian repo (this can happen when repos are restructuring or updating versions, I've seen it on archive.raspberrypi.org even.)

We also don't have to maintain redundant paths in two locations. With the `update_github_script.sh` script format, it requires that the majority of the URL be specified in both the update script as well as the install script. This is not the case here, simply give the location of the debian `Packages` file on the repo, as well as the desired `package` name from that repo, and the script handles sorting all the versions available, and creating the full URL. It also verifies the URL is valid as a fallback as with the other script.

.github/workflows/update_debian_repo_script.sh
```diff
@@ -0,0 +1,157 @@
...
+if [ -n "$armhf_webPackages" ] || [ -n "$arm64_webPackages" ] || [ -n "$all_webPackages" ]; then
+#iterate through all variables supplied if an array
+# get longest arrary of passed variables
+num=$(echo -e "${#armhf_webPackages[@]}\n${#arm64_webPackages[@]}\n${#all_webPackages[@]}" | sort -n -r | head -1)
+
+for (( iter=0; iter<$num; iter++ )); do
+#make array appear as one variable inside for loop
+armhf_webPackages=${armhf_webPackages[$iter]}
+arm64_webPackages=${arm64_webPackages[$iter]}
+all_webPackages=${all_webPackages[$iter]}
```
Botspot, line 20: I was not aware this was possible: to reuse the same name for both an array and a variable, without encountering complexities. Where did you learn about this, @theofficialgman?

.github/workflows/update_debian_repo_script.sh
```diff
@@ -0,0 +1,157 @@
...
+arm64_packagename=${arm64_packagename[$iter]}
+all_packagename=${all_packagename[$iter]}
+if [[ $iter == "0" ]]; then
+    version_number=""
+else
+    version_number=$((iter + 1))
+fi
+
+#set pi_apps_filepath variables
+if [ -n "$armhf_webPackages" ] && [ -f install-32 ]; then
```
Botspot, line 31: > `[ -f install-32 ]; then`

What would happen if the working directory was not set to the app folder?

theofficialgman: there would be issues if it wasn't setup as the app folder but now it is setup as the app folder and has been.

the `update_github_script.sh` script makes this same assumption

.github/workflows/update_debian_repo_script.sh
theofficialgman: I'm taking advantage of how bash uses variables and arrays.
the array doesn't get overwritten, only the first value in the array gets overwritten.

when you call the array without specifying the index, its assumes the first always.

so what is happening is I overwrite the first index of the array every time I move through the loop with the desired content of a later index

theofficialgman:
I went ahead and merged this since it won't get used unless its included in an apps updater
I do feel like it is a useful addition and different enough from the github/generic updater script

PR #1646, opened 2022-03-27
Add backtick quotes to versions to look nicer in PR

cycool29:
And also to avoid unwanted markdown marking (like https://github.com/Botspot/pi-apps/pull/1645).


theofficialgman:
I'd rather not do this, since it made the commit message ugly and polluted with backtick quotes
see my test here: https://github.com/theofficialgman/testing/commit/0ca1ee1a957913fecd58df0e35bf4940fbfc49e1

cycool29:
I don't think it is *polluted* with backticks, it looks clear and nice to me even in plain text (commit message).

I just want to make PRs look nicer. Some app versions contain symbols, and I think they look tidier with backtick quotes in PRs.

PR #1652, opened 2022-03-30
fix edex ui

ryanfortner:


Botspot:
`install-64` needs to be fixed as well.

ryanfortner:
@Botspot does everything oook alright?

Issue #1660, opened 2022-04-01
Fix Intellij IDEA and PyCharm for arm64


Botspot:
Note to future self or other contributors: Intellij IDEA and PyCharm scripts need to be changed to avoid hard-coding an armhf JDK path on arm64 operating systems.
Thanks to @jai-jap for noticing this: https://github.com/Botspot/pi-apps/commit/b4b24b4ad8054261d9ece8301478f7c137a56a88#r70139282

Botspot:
I can't work on it now - have other things to do.

PR #1661, opened 2022-04-01
Fix Java home directory for arm64

cycool29:
To close https://github.com/Botspot/pi-apps/issues/1660.


Botspot:
The whole purpose of specifying the JAVA_HOME and PATH variables is to prevent some other Java version from interfering and causing a launch error. If this wasn't an issue then we wouldn't need to set the variables at all.

cycool29:
Fixed. @Botspot

PR #1671, opened 2022-04-08
PiGro install

actionschnitzel:
I added code to the install file that generates ".desktop" files for the desktop and start menu. The original file contains the path "/home/pi/.." . Since the new Pi OS update eliminates the user "Pi", I had to change this. Code is test! / I took it from pi apps installer ;-)

best regards

Timo

### DIFF
--- a/apps/PiGro/install
+++ b/apps/PiGro/install
@@ -29,12 +29,6 @@ Path=${DIRECTORY}/" > ~/Desktop/pigro.desktop
 
 sudo chmod +x ~/Desktop/pigro.desktop
 
-
-DIRECTORY="$(readlink -f "$(dirname "$0")")"
-if [ -z "$DIRECTORY" ] || [ "$DIRECTORY" == "$HOME" ] || [ "$DIRECTORY" == bash ];then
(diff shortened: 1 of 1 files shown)


apps/PiGro/install
```diff
@@ -5,9 +5,46 @@ git_clone https://github.com/actionschnitzel/PiGro-Aid- || error "Failed to clon
...
+
+sudo chmod +x ~/Desktop/pigro.desktop
+
+
+DIRECTORY="$(readlink -f "$(dirname "$0")")"
+if [ -z "$DIRECTORY" ] || [ "$DIRECTORY" == "$HOME" ] || [ "$DIRECTORY" == bash ];then
+  DIRECTORY="$HOME/PiGro-Aid-"
+  
+
+fi
```
Itai-Nelken, line 38: No need to do this again.

actionschnitzel: Done!

actionschnitzel:
Thanks @Botspot . I tested it and it works. But I don't understand that. I did exactly the same thing at first, but always got the message that the file was corrupt. But now it work -.- .... Doesn't matter. Thanks again.

PR #1686, opened 2022-04-14
Fix Box86 arm64

Jai-JAP:
Try to fix #1685 
Do not merge

Issue #1685, opened 2022-04-14
Box86 no binary on arm64


bill88t:
### Confirmations

- [X] I searched for similar bug reports in https://github.com/Botspot/pi-apps/issues and found none were relevant.
- [X] I tried updating Pi Apps and the issue is still not fixed. [How to update Pi Apps?](https://github.com/Botspot/pi-apps#to-update-pi-apps)

### What happened?

The Box86 app does not create a binary on arm64 on installation

### Description

I saw #1625 and gave it a try.
The install succeded, however, it's not in path or anywhere basically, the install log is attached.
![image](https://user-images.githubusercontent.com/21365918/163363687-bf17819c-e39e-4b0c-9f84-dc2ef5a37efa.png)

### What is your operating system?

Raspberry Pi OS bullseye arm64

### (Optional) Error log? Terminal output? Debug messages?

```bash
OS: Debian GNU/Linux 11 (bullseye)
OS architecture: 64-bit
Installed Pi-Apps version: 04/14/2022
Most recent Pi-Apps version: 04/14/2022
Kernel: aarch64 5.15.32-v8+
Device model: Raspberry Pi 400 Rev 1.0
Ram size: 3.61 GB
Raspberry Pi OS image version: 2021-10-30
Language: en_US.UTF-8

BEGINNING OF LOG FILE:
-----------------------

armhf arcitecture already added...
[...]

PR #1725, opened 2022-04-25
Needs to be box86:armhf or it downloads a package from August 2021

Onfy:
I'm not sure why but when apt downloads the plain box86 package on aarch64 it's downloading a rather old build of box86. It doesn't seem to be a box86:arm64 package (that would be quite illogical), maybe it's some sort of meta package? Changing box86 to box86:armhf will make it get the latest build instead.

(Sorry if I did this wrong, this is literally the first time I've submitted a PR.)

### DIFF
--- a/apps/Box86/install-64
+++ b/apps/Box86/install-64
@@ -35,7 +35,7 @@ echo "Adding key..."
 wget -qO- https://itai-nelken.github.io/weekly-box86-debs/debian/KEY.gpg | sudo apt-key add - || error "Failed to add key to box86 repo!"
 echo "Installing box86..."
 sudo dpkg --remove box86-no-binfmt-restart 2>/dev/null
-install_packages box86 || exit 1
+install_packages box86:armhf || exit 1
 
 if ! sudo systemctl restart systemd-binfmt ;then
   echo -e "\nWarning: systemd-binfmt failed to restart. Getting debug info..."
--- a/apps/Box86/install-64
+++ b/apps/Box86/install-64
@@ -43,3 +43,15 @@ if ! sudo systemctl restart systemd-binfmt ;then
   systemctl status systemd-binfmt.service
   echo "Still exiting with success, but you will have to manually run all x86 applications with box86 as binfmt isn't there to do it for you."
 fi
+
+# adding some debug output to the end of the script
+echo "dpkg --print-foreign-architectures:"
+dpkg --print-foreign-architectures
+echo "zcat /proc/config.gz | grep CONFIG_COMPAT="
+zcat /proc/config.gz | grep CONFIG_COMPAT=
+echo "dpkg -l libc6:armhf"
+dpkg -l libc6:armhf
+echo "/lib/arm-linux-gnueabihf/ld-*.so"
+/lib/arm-linux-gnueabihf/ld-*.so
+echo "/usr/local/bin/box86 --version"
+/usr/local/bin/box86 --version


theofficialgman:
would you mind showing the output of
` apt-cache show box86`

and `apt-cache show box86:armhf`

this will help in determining the information for these two packages and determine what repo they are coming from

Onfy:
Those commands show some pretty interesting (and... long) information. See the attachment
[terminaloutput.txt](https://github.com/Botspot/pi-apps/files/8559233/terminaloutput.txt)
.

theofficialgman:
so there is one arm64 box86 package... a really old one
```
Package: box86
Version: 0.2.3+20210709.3a7cc57f-1
Architecture: arm64
```
@Itai-Nelken please explain why this exists and if this old outdated package is the one supposed to be used on arm64 or if the armhf packages from the repo (of which there are many newer versions) are supposed to be used

theofficialgman:
it was specifically changed in this pr... https://github.com/Botspot/pi-apps/pull/1686
@Jai-JAP @Itai-Nelken I think you two need to get together and review exactly how this is actually supposed to be installed

Itai-Nelken:
I don't build arm64 debs, only armhf. I didn't realize that #1686 changed that...

I plan to move the builds to gh actions, but until I do that there are only armhf debs.

The arm64 ones are some old ones that were built when I was experimenting with using ubuntu arm64, but ut had a newer libc than piOS buster so I switched back.

Itai-Nelken (approved):

Itai-Nelken:
There is still #1685, I have no idea why it happens though.

Jai-JAP:
Installing box86:armhf results kn no binary being installed. But the box86:armhf builds seem to work fine

Jai-JAP:
Maybe adding `Multi-Arch: same` to debian control files will fix this issue. I have seem some cases with such results. Especially for libraspberrypi0:armhf on RPiOS arm64 buster
What do you think @Itai-Nelken

Onfy:
> There is still #1685, I have no idea why it happens though.

Now that you mention it, I believe I initially ran into the same issue. As I recall, box86 actually *did* get installed, i.e. running `whereis box86` would show it was installed correctly, but simply running box86 from the terminal would show a not found error, and running box86 from its absolute path resulted in an ELF format error. I remember having to install something manually to get it to work, but unfortunately I forget what. I should write a fresh Bullseye image to try to investigate.

(And of course after getting box86 running on my system, it was massively out of date, I ran into bugs that shouldn't exist anymore, and now we're here. Fun bugs.)

(As an aside, I'm not sure yet if this directly affects box86, [but in theory it affects anything 32-bit that outputs sound, audacity:armhf was affected too] but when I was first experimenting with armhf multiarch, I ran into an interesting problem in which a pulseaudio component in armhf flavour wasn't installed by default, either as a dependency of libraspberrypi0:armhf or box86. It made Redream [Dreamcast emulator]... exciting. [Loud.])

theofficialgman:
If you are getting an elf error when trying to run the box86 armhf binary on your arm64 install, that is because you do not have the necessary multiarch packages installed. I'm not sure if it's just glibc armhf that needs to be installed or if there is more

Onfy:
> If you are getting an elf error when trying to run the box86 armhf binary on your arm64 install, that is because you do not have the necessary multiarch packages installed. I'm not sure if it's just glibc armhf that needs to be installed or if there is more

Correct. I'm pretty sure I installed a library and then it worked... I just don't remember which or how I figured it out. I'm guessing that with the correct lib added as a dependency, box86 will start working out of the... box... for more people. (And me missing that issue and focusing on this problem... I guess that's missing the forest for the [boxed?] tree.)

Itai-Nelken:
I think the problem was that the armhf architecture wasn't added.
That happens here: https://github.com/Botspot/pi-apps/blob/master/apps/Box86/install-64#L9-L18

theofficialgman:
> I think the problem was that the armhf architecture wasn't added. That happens here: https://github.com/Botspot/pi-apps/blob/master/apps/Box86/install-64#L9-L18

Nope that's not it. The name of the package told to install was box86, it's NEVER going to install the armhf version unless you tell it to do so

Itai-Nelken:
> > I think the problem was that the armhf architecture wasn't added. That happens here: https://github.com/Botspot/pi-apps/blob/master/apps/Box86/install-64#L9-L18
> 
> Nope that's not it. The name of the package told to install was box86, it's NEVER going to install the armhf version unless you tell it to do so (box86:armhf)

I meant before the change.

theofficialgman:
added debug output to install to get required info upfront incase an issue like https://github.com/Botspot/pi-apps/issues/1685 creeps up again.

if something like this is posted, users should be instructed to submit their install log of box86 on their arm64 system.

Jai-JAP:
@Itai-Nelken can you apply the fix I mentioned [above](https://github.com/Botspot/pi-apps/pull/1725#issuecomment-1109663967) since box86:armhf still doesn't work

theofficialgman:
> @Itai-Nelken can you apply the fix I mentioned [above](https://github.com/Botspot/pi-apps/pull/1725#issuecomment-1109663967) since box86:armhf still doesn't work

output logs please for the install script

Jai-JAP:
No logs atm but i read them previously and it fails on `/usr/local/bin/box86 --version` with `/usr/local/bin/box86 : no such file or directory`

theofficialgman:
I still need the rest of the logs in order to interpret that.

Jai-JAP:
I'll post them later

Itai-Nelken:
this issue should be fixed.
I deleted all the `box86-no-binfmt-restart` packages from the repo, so apt can't find them.
anyone who has the box86 problem, try reinstalling it.

PR #1736, opened 2022-04-28
Fix update_github_script.sh To Properly Escape SED Commands

TheBrokenRail:
This should fix MCPI's auto-updating.

### DIFF
--- a/.github/workflows/updates/Minecraft Pi (Modded).sh
+++ b/.github/workflows/updates/Minecraft Pi (Modded).sh
@@ -13,5 +13,6 @@ deb_id="$(curl -s "https://jenkins.thebrokenrail.com/job/minecraft-pi-reborn/job
 # Update
 webVer="${build_id}/artifact/out/minecraft-pi-reborn-client-${deb_id}"
 # This URL is only used for validation.
-all_url="https://jenkins.thebrokenrail.com/job/minecraft-pi-reborn/job/master/${webVer}-armhf.AppImage"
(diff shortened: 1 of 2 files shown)


ryanfortner:
Has this been tested?

ryanfortner:
lgtm, just need to make sure with @Botspot before merging.

.github/workflows/updates/Minecraft Pi (Modded).sh
```diff
@@ -13,5 +13,6 @@ deb_id="$(curl -s "https://jenkins.thebrokenrail.com/job/minecraft-pi-reborn/job
 # Update
 webVer="${build_id}/artifact/out/minecraft-pi-reborn-client-${deb_id}"
 # This URL is only used for validation.
-all_url="https://jenkins.thebrokenrail.com/job/minecraft-pi-reborn/job/master/${webVer}-armhf.AppImage"
+armhf_url="https://jenkins.thebrokenrail.com/job/minecraft-pi-reborn/job/master/${webVer}-armhf.AppImage"
+arm64_url="https://jenkins.thebrokenrail.com/job/minecraft-pi-reborn/job/master/${webVer}-arm64.AppImage"
```
theofficialgman, line 17: ah yeah oops, I totally forgot I intended for this to be possible when writing the scripts.

theofficialgman:
I never intended for the version to contain a `/`
the current sed commands use this as the delimiter, hence the problem

we can either do what this PR does, or change the delimiter to something else not usable like so:
`sed -i "0,/version${version_number}=.*/s;;version${version_number}=${webVer}/g" install-64`

theofficialgman:
I committed my changes, I'd prefer to to add extra complexity to this unless necessary

PR #1744, opened 2022-04-30
Fix Hyper update script

Crilum:
Hyper has official arm64 debs, I haven't tested this (I don't use arm64), so somebody should probably test the official debs, but if the official debs work, then @Jai-JAP only has to build armhf debs.

I also bumped the version on install-64 to `3.2.3`, the latest release.

### DIFF
--- a/.github/workflows/updates/Hyper.sh
+++ b/.github/workflows/updates/Hyper.sh
@@ -7,6 +7,6 @@ source $GITHUB_WORKSPACE/.github/workflows/update_github_script.sh
 
 unset webVer
 unset armhf_url
-
+webVer=$(get_release vercel/hyper)
 arm64_url="https://github.com/vercel/hyper/releases/download/v${version}/hyper_${version}_arm64.deb"
 source $GITHUB_WORKSPACE/.github/workflows/update_github_script.sh
--- a/.github/workflows/updates/Hyper.sh
+++ b/.github/workflows/updates/Hyper.sh
@@ -1,12 +1,7 @@
 #!/bin/bash
 
 webVer=$(get_release Jai-JAP/hyper-arm-builds)
-armhf_url="https://github.com/Jai-JAP/hyper-arm-builds/releases/download/v${version}/hyper_${version}_armv7l.deb"
+armhf_url="https://github.com/Jai-JAP/hyper-arm-builds/releases/download/v${webVer}hyper_${webVer}_armv7l.deb"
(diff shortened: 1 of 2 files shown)


Crilum:
@theofficialgman How would one have arch-separate `webVer` variables? The hyper update script will have two separate repos for armhf and arm64.

theofficialgman:
you basically just call the source the updater script twice
refer to FreeTube or Zoom's updater script for an example:
https://github.com/Botspot/pi-apps/blob/master/.github/workflows/updates/FreeTube.sh

theofficialgman:
Actually both of these armhf_url arm64_url are incorrect and we're incorrect before. The substitution needs to be webVer in them, not version

Jai-JAP:
This should be closed as it will not work. Hyper arm64 debs are broken. See https://github.com/vercel/hyper/issues/6380

Crilum:
@theofficialgman @Jai-JAP Ok, I'll change this PR to change the version vars to webVer vars...

Crilum:
Ok, @theofficialgman I've fixed the updater script, and @Jai-JAP, I've reverted the arm64 urls back to your repo.
This can't be merged until @Jai-JAP has v3.2.3 builds.

.github/workflows/updates/Hyper.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 webVer=$(get_release Jai-JAP/hyper-arm-builds)
-armhf_url="https://github.com/Jai-JAP/hyper-arm-builds/releases/download/v${version}/hyper_${version}_armv7l.deb"
+armhf_url="https://github.com/Jai-JAP/hyper-arm-builds/releases/download/v${webVer}hyper_${webVer}_armv7l.deb"
 arm64_url="https://github.com/Jai-JAP/hyper-arm-builds/releases/download/v${version}/hyper_${version}_arm64.deb"
```
theofficialgman, line 5: this still has the wrong variable name. I know it was already wrong before but please fix it here as well

theofficialgman:
closing since this is not necessary, updates will come automatically if Jai-JAP updates their build. you can re-open or create a new PR if upstream fixes their builds

.github/workflows/updates/Hyper.sh
Crilum: I thought I fixed it.. Thanks for fixing it.

PR #1747, opened 2022-05-01
balenaEtcher: remove unnecessary uninstall lines

ryanfortner:
I don't think these two lines are necessary, but I wanted to make a PR just in case they are.


Itai-Nelken:
It's there in case the install script fails while downloading.
I don't think it's necessary anymore though.

PR #1760, opened 2022-05-03
Add Renoise (Demo)

Crilum:
Renoise is a complete, multi-platform and expandable Digital Audio Workstation. It lets you record, compose, edit, process and render production-quality audio using a music-tracker based approach. 

Website: https://renoise.com

This is only a demo, and Renoise isn't open source, but the devs provide Raspberry Pi packages.


Crilum:
I haven't tested `install-64`, but it should be good to go.

apps/Renoise (Demo)/uninstall
```diff
@@ -0,0 +1,13 @@
...
+wget "https://raw.githubusercontent.com/Crilum/stuff/main/install.sh" || error "Failed to download 'install.sh'!"
+chmod +x install.sh || error "Failed to make 'install.sh' executable!"
+sudo ./install.sh -u
+if [ "$?" == "1" ]; then
+  status "Uninstall finished!"
+elif [ "$?" == "0" ]; then
+  error "Failed to uninstall Renoise with 'install.sh -u'"
+else
+  warning "Unknown status code... Renoise may or may not have uninstalled successfully."
+fi
```
theofficialgman, line 13: this is unprofessional, why do you feel the need to do this? if Renoise is extracted, the the installs script always remains and you should run it RESOURCES_PATH (aka, `RESOURCES_PATH=$SYSTEM_LOCAL_SHARE/renoise-$RENOISE_VERSION`

you see here this is part of the install:
```
    for f in "install.sh" "uninstall.sh" "Installer"; do
      cp -r $SCRIPT_PATH/$f $RESOURCES_PATH || bail_out
    done
```
please edit your uninstall script to use the locally available one.

Crilum: Thanks, I'll fix this.

Crilum: Does my latest commit do what you were wanting?

theofficialgman: yup, assuming that is the correct path, that is good

Crilum: Yep, it is. I tested it.

theofficialgman:
Similar to the other PR, I don't know if we should be including demo as part of the name or not. I would like @Botspot 's feedback on that

Crilum:
Well, this *is* a Demo, [according to their website](https://www.renoise.com/download). Disabled features and all that, although it isn't time regulated...

theofficialgman:
> Well, this _is_ a Demo, [according to their website](https://www.renoise.com/download). Disabled features and all that, although it isn't time regulated...

the question I think is, does upgrading to the full version require installing something else, or is it just adding a key file somehwere/typing a key into the UI.

I'm hesitant to start adding demo to everything because you could consider even the minecraft launchers a "demo", yeah they will install, but you need to own the game to use them

Crilum:
I'll look in Renoise and see if I can find a activation tool, or something.

Crilum:
So, @theofficialgman, I looked in Renoise, and all I could find was a option in the help menu that says `Get the full version`, which opens https://renoise.com/shop.
It looks like if you buy renoise, they will give you another installer that will install an activated version. I think this should be named as a demo.

Crilum:
@Botspot This is mergable, whenever you're ready..

Botspot:
In this case we can look back to precedent. Puffin Browser is also a demo. In its case, I decided that for brevity the app would just be named "Puffin", but the description would clarify that this is actually a demo.

For Puffin's case, few users would care that it's a demo - as it's entirely usable with no noticeable limitations. For Renoise however, I looked through the list of restrictions with the demo version and it seems far more annoying. So in this case, I think all users would want to know up-front that this will install the demo version of Renoise. So I think that in this case it's best to leave "(Demo)" in the name, in addition to listing the restrictions in the description.

I'll merge this and then create the analytics link for it.

PR #1761, opened 2022-05-03
Add Reaper

Crilum:
REAPER is a complete digital audio production application for computers, offering a full multitrack audio and MIDI recording, editing, processing, mixing and mastering toolset.
REAPER supports a vast range of hardware, digital formats and plugins, and can be comprehensively extended, scripted and modified. 

Website: https://reaper.fm

Reaper is not open source, but the devs release arm builds.. This is also a 60 day free  trial, but it doesn't have anything disabled during the trial.

I have only tested `install-32` and `uninstall`, but the only thing different on 64 bit is the url, and top level dir names.


apps/Reaper (Demo)/install-32
```diff
@@ -0,0 +1,29 @@
...
+mkdir -p "$HOME/.local/share/icons/hicolor/64x64/apps/" || error "Failed to make icon directory!"
+cp "${DIRECTORY}/pi-apps/apps/Reaper (Demo)/icon-64.png" "$HOME/.local/share/icons/hicolor/64x64/apps/reaper.png" || error "Failed to copy icon!"
+echo "[Desktop Entry]
+Name: Reaper
+Comment: Reaper is a complete Digital Audio Workstation, for computers, offering a full multitrack audio and MIDI recording, editing, processing, mixing and mastering toolset. 
+Exec: reaper
+Icon: $HOME/.local/share/icons/hicolor/64x64/apps/reaper.png
+Terminal: false
+Category: AudioVideo;Audio;
+Type: Application
```
Botspot, line 24: Are you sure that placing colon characters here is supported by all desktop environments? I'm used to "`Name=`", "`Exec=`", etc.

Crilum: In the `Category`? I'm pretty sure it's normal, lots of desktop files use semicolons to separate categories.

Crilum: Oh sorry, I see what you're saying. I think you're right, I'll fix it.

Crilum:
Should be good to go now, @Botspot.

theofficialgman:
maybe we should consider removing demo from the application name.
The demo, as you said and as the website says, is the full version with a 60 day trial period. if the user buys the license then its not a demo anymore but the full version.

Crilum:
@theofficialgman @Botspot Look good now?

apps/Reaper/install-32
```diff
@@ -0,0 +1,29 @@
...
+tar -xf reaper.tar.xz || error "Failed to extract reaper!"
+cd reaper_linux_armv7l || error "Failed to cd into 'reaper_linux_armv7l'"
+status_green "Done!"
+status "Installing reaper..."
+chmod +x install-reaper.sh || error "Failed to make 'install-reaper.sh' executable!"
+sudo ./install-reaper.sh --install /opt || error "Failed to run 'sudo install-reaper.sh --install /opt'"
+sudo rm -f /usr/local/bin/reaper &>/dev/null
+sudo ln -s /opt/REAPER/reaper /usr/local/bin/reaper || error "Failed to link /opt/REAPER/reaper to /usr/local/bin/reaper!"
+mkdir -p "$HOME/.local/share/icons/hicolor/64x64/apps/" || error "Failed to make icon directory!"
+cp "${DIRECTORY}/pi-apps/apps/Reaper/icon-64.png" "$HOME/.local/share/icons/hicolor/64x64/apps/reaper.png" || error "Failed to copy icon!"
```
Crilum, line 16: @theofficialgman isn't the `$DIRECTORY` variable equivalent to `/location/of/pi-apps`? If so, shouldn't this line be:
```suggestion
cp "${DIRECTORY}/apps/Reaper/icon-64.png" "$HOME/.local/share/icons/hicolor/64x64/apps/reaper.png" || error "Failed to copy icon!"



```

theofficialgman: oh yeah it is. my bad. good catch

Crilum:
@Botspot Looks good now.. Mergeable whenever...

Crilum:
> Should be good to go now, @Botspot.



> @theofficialgman @Botspot Look good now?



> @Botspot Looks good now.. Mergeable whenever...

@Botspot @theofficialgman Sorry for so many prompts... lol

Botspot:
Just tested, and I saw this error. Is this a problem?
2022-05-07-231816_1920x1080_scrot

Crilum:
Nope, try running `qjackd` in terminal, (installed by default on rpios) and starting JACK, then running Reaper again.

Botspot:
qjackd: Command not found. This is unsurprising, as RPiOS includes ALSA and pulseaudio by default. About a year ago I heard that JACK caused problems on PiOS, but a lot has changed since then.

I think this app should ensure JACK dependencies are installed.

Crilum:
Wait @Botspot, one sec, I just opened reaper and there appears to be a new version of reaper available... 

See my edit, `qjackctl`, not `qjackd`

Botspot:
qjackctl: command not found.

Crilum:
Hmm, it's been installed by default on all my RPiOS's.. Are you using Twister?

Crilum:
You can also choose to use PulseAudio, or ALSA, in Preferences (ctrl + P)-> Audio -> Device

Crilum:
`jackd` also comes preinstalled on RPiOS..

Crilum:
@Botspot try `jackd -d alsa` (this will start a `jackd` daemon with `alsa` as the backend), then try Reaper again.

Botspot:
> `jackd` also comes preinstalled on RPiOS..

I don't recall uninstalling it, but I will assume you are right. Worst case scenario is our users having to click "Audio device" on the error dialog.

I'll merge this unless you have any more changes to make.

Crilum:
Nope, I just updated the version to the latest (released today, so we should be good to go!

Crilum:
Thanks @Botspot!

Crilum:
So @Botspot, I just reinstalled Reaper, and it appears that the latest release has a bad tar archive.. 6.56 works, but 6.57 (latest) fails because of bad tar archive:
```
Extracting reaper...
xz: (stdin): Compressed data is corrupt
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
Failed to extract reaper!

```

We probably need to revert the version to `6.56`.

PR #1673, opened 2022-04-09
Fix Minecraft Pi Modded to use AppImage

leha-code:
Closes #1666

PR #1767, opened 2022-05-04
fix Sublime Text arm64 installation script

touhidurrr:
As we know that `apt-key` is deprecated, it doesn't work on Raspberry Pi OS arm64.
The changes I made was manually tested by me so I can tell that they work.
Installing the package `gpg` is not required as it already comes default with a fresh installation of Raspberry Pi OS arm64, I tried to install it anyways for better support.

### DIFF
--- a/apps/Sublime Text/install-64
+++ b/apps/Sublime Text/install-64
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ## Install dependencies
-install_packages wget gpg || error "Failed to install dependencies"
+install_packages wget || error "Failed to install dependencies"
 
 status "Adding GPG key..."
 wget -qO- https://download.sublimetext.com/sublimehq-pub.gpg | sudo tee /etc/apt/trusted.gpg.d/sublimehq.asc || error "Failed to add GPG key for sublime-text repository!"
--- a/apps/Sublime Text/install-64
+++ b/apps/Sublime Text/install-64
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-## Install dependencies
-install_packages wget || error "Failed to install dependencies"
-
 status "Adding GPG key..."
 wget -qO- https://download.sublimetext.com/sublimehq-pub.gpg | sudo tee /etc/apt/trusted.gpg.d/sublimehq.asc || error "Failed to add GPG key for sublime-text repository!"
 echo "deb https://download.sublimetext.com/ apt/stable/" | sudo tee /etc/apt/sources.list.d/sublime-text.list
--- a/apps/Sublime Text/install-64
+++ b/apps/Sublime Text/install-64
@@ -1,6 +1,10 @@
 #!/bin/bash
-echo "Adding GPG key..."
-wget -qO- https://download.sublimetext.com/sublimehq-pub.gpg | sudo apt-key add - || error "Failed to add GPG key for sublime-text repository!"
+
+## Install dependencies
+install_packages wget gpg || error "Failed to install dependencies"
+
+status "Adding GPG key..."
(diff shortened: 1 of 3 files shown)


apps/Sublime Text/install-64
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
-echo "Adding GPG key..."
-wget -qO- https://download.sublimetext.com/sublimehq-pub.gpg | sudo apt-key add - || error "Failed to add GPG key for sublime-text repository!"
+
+## Install dependencies
+install_packages wget gpg || error "Failed to install dependencies"
+
+status "Adding GPG key..."
+gpg --keyserver keyserver.ubuntu.com --recv-keys F57D4F59BD3DF454 || error "Failed to add GPG key for sublime-text repository!"
```
Botspot, line 7: The current script gets a gpg key from https://download.sublimetext.com/sublimehq-pub.gpg, but this proposed modification would get the key from keyserver.ubuntu.com. Is this a security risk? Also, if the imported key is written to a universal location then it seems like this would need to run with `sudo.

touhidurrr: Pretty sure it should work without keyserver option too. In that case it will use the keyservers of gpg maintainers maybe.

theofficialgman: we already have two other valid methods for replacing apt-key, I'm not a fan of obtaining the key from the ubuntu server over gettting it directly from sublimetext.

now that I'm looking for it, I can't find the issue/discussion to link to where the methods are detailed but botspot knows what I'm talking about

theofficialgman: found them:
https://github.com/Botspot/pi-apps/pull/1673#issuecomment-1095351120

https://github.com/Botspot/pi-apps/pull/1673#issuecomment-1106777695

touhidurrr: About the first method, I tried the following but it didn’t work:
```bash
sudo curl https://download.sublimetext.com/sublimehq-pub.gpg -o /etc/apt/trusted.gpg.d/sublimehq-pub.gpg
sudo apt update
```
But it didn’t work

touhidurrr:
Ok @Botspot @theofficialgman, I changed the way gpg keys are added. This way to add gpg keys is probably the safest as it doesn't require sudo to get things done.
So, unless this doesn't work. You guys should accept the pr.

touhidurrr:
Also note that I have already tested this and from my tests it works.
If you wondering why it doesn't require sudo. it seems like it saves the keys to `~/.gnupg`.
Anyways, thats all.

theofficialgman:
> Also note that I have already tested this and from my tests it works. If you wondering why it doesn't require sudo. it seems like it saves the keys to `~/.gnupg`. Anyways, thats all.

@touhidurrr 
no we can't do what you did. this will break multi-user installs
the apt repo is installed for  all users, but the gpg key for authentication is only installed for the current user `~`

please use one of the methods already detailed which installs for all users

touhidurrr:
@theofficialgman used one of the methods you mentioned
Although from my tests a likely alternative was:
```bash
wget -qO- https://download.sublimetext.com/sublimehq-pub.gpg | sudo gpg --import
```
It seems like using gpg in sudo mode makes it use `/root/.gnupg` instead

theofficialgman:
I'll merge this for now, but note that botspot has expressed interest in changing all scripts over to using a new (unwritten as of now) function for adding there gpg key

touhidurrr:
> I'll merge this for now, but note that botspot has expressed interest in changing all scripts over to using a new (unwritten as of now) function for adding there gpg key

Oh, that would be good.

PR #1780, opened 2022-05-08
add greeting workflow

ryanfortner:
This GitHub action sends a friendly greeting to users when they make their first PR or issue on the repo. Inspired by https://github.com/wimpysworld/deb-get/blob/main/.github/workflows/greetings.yml

### DIFF
--- a/.github/workflows/greeting.yml
+++ b/.github/workflows/greeting.yml
@@ -12,8 +12,8 @@ jobs:
         issue-message: |
           Hello there 👋
           Thanks for submitting your first issue to the Pi-Apps project! We'll try to get back to you as soon as possible. 
-          In the meantime, we encourage you to join our [Discord server](https://discord.gg/RXSTvaUvuu), where we can help you directly. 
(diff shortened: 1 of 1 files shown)


theofficialgman:
I like this idea, but it might be a good idea to go ahead and prompt the user to read the documentation as well if its their first PR:
something like 
```
While you wait, make sure you have read the documentation: https://github.com/Botspot/pi-apps/wiki/Creating-an-app
```

Botspot:
> ```
> While you wait, make sure you have read the documentation: https://github.com/Botspot/pi-apps/wiki/Creating-an-app
> ```

I would avoid calling this "the documentation", as that would be confused with DOCUMENTATION.md.

Overall, I like this idea. But as we know, some people can become rather impatient with their issues & PRs, so inviting them to the discord server first thing may be asking for trouble. To be clear, I still think the benefits outweigh the potential risks though.

Feel free to merge this PR once its finalized.

ryanfortner:
added your suggestions. merging now.

PR #1781, opened 2022-05-08
Add ReaPack (Reaper Package Manager)

Crilum:
This is a Reaper package manager, which is home to over 1300 Reaper plugins and addons (synths, Web Interfaces, APIs, themes, and more).

Website: https://reapack.com
GH Repo: https://github.com/cfillion/reapack

This PR changes: new update script, added ReaPack to categories, added ReaPack app.


Botspot:
Only Reaper users will be interested in this. Nobody would want ReaPack without first installing Reaper. (Right?)
Thus, I suggest changing the Reaper app's scripts to include ReaPack, either optionally, or by default.

theofficialgman:
I agree with botspot, this is literally just an extra binary (which is funny, I've never seen anyone just drop `.so` files as a github release)
```
wget "https://github.com/cfillion/reapack/releases/download/v${version}/reaper_reapack-armv7l.so" || error "Failed to download ReaPack!!!"
```
I suggest just adding this to the the Reaper install script. you can put it behind a user input prompt if you want to make it optional, but honestly I don't see an issue with including it by default

Crilum:
> (which is funny, I've never seen anyone just drop .so files as a github release)

ReaPack opens inside of Reaper, so I guess Reaper uses .so files? :man_shrugging:

Crilum:
Okay... I'll close this PR and open another one to add Reapack to Reaper.

PR #1782, opened 2022-05-08
Remove hirsute as eol

Jai-JAP:
@Botspot @theofficialgman Removed ubuntu hirsute from box86 install-64 as it is eol and removed from official repos


Botspot:
This change to the install script would force an unnecessary update + reinstall for everyone else.
As such, I think it's best to wait on this change until the Box86 script requires some other, more important, change.

Jai-JAP:
So I'll close this for now

Issue #185, opened 2020-12-29
Time for an app eligibility rubric!


Botspot:
It's time to write out what makes a good Pi-App.  
And, we need a standardized method of ranking app requests.

Here's my idea below:

1. 1-5 points: How many causal GUI users would find this app useful?
3. 1-5 points: How hard is it to install outside of Pi-Apps? (just one command? Or 20 commands?)
4. 1-5 points: The awesome factor. (5 is Kerbal Space program, 4 is an exotic browser, 1 is a slightly newer version of an antivirus)

Add up the points and see what you get. Below are some special cases where points don't matter - the app is disqualified:

- The app is already available on APT. 
- ~~The app requires paying for something.~~ Edit: What was I thinking? MC Java and MC Bedrock are already on here, and [KSP](https://github.com/Botspot/pi-apps/issues/139) is on the waiting list. I think they belong in Pi-Apps, despite costing money. This point is no longer valid.
- The app is illegal to distribute.

This issue is for **brainstorming**. Let's tally up the points for a few apps currently in Pi-Apps, and see which of them rank the highest.  
Do any apps seem unfairly judged too high or too low? Let me know!

Botspot:
This really isn't an issue.

Closing. Something similar to this, minus the "points" system will be added to the README at some point.

Botspot:
Note to self for future reference:

> Hey Botspot, you can view issues that can be closed due to inactivity, by clicking here: https://github.com/Botspot/pi-apps/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-asc
> (Open issue list sorted by least recent activity)

srini-pro:
hey talktosrinivasn is open

srini-pro:
is'nt talktosrinivasan is subbmited #talktosrinivasan

Botspot:
> is'nt talktosrinivasan is subbmited #talktosrinivasan

@srini-pro STOP spamming this issue please.

PGSCOM:
#1156

PR #1790, opened 2022-05-09
Add app eligibility rules to README

Crilum:
@Botspot said in #185 these rules would eventually be added to the readme, I guess "eventually" is now. :man_shrugging:


Botspot:
I think it would be better to put this in [the App Creation Wiki](https://github.com/Botspot/pi-apps/wiki/Creating-an-app).
(At the time of #185, we didn't have the wiki enabled)

Botspot:
I've added app eligibility to the wiki now. Thank you @Crilum for the reminder to do this. :)

Crilum:
> I've added app eligibility to the wiki now. Thank you @Crilum for the reminder to do this. :)

You're welcome! :)

PR #1792, opened 2022-05-10
Add Reaper Update script, and add ReaPack to Reaper

Crilum:
closes #1781 (Oh wait, it's already closed :man_facepalming:)

@theofficialgman Isn't there a way to have two version variables? Reaper needs two (or at least could use two), one for Reaper, and one for ReaPack.


theofficialgman:
yes, see cura for an example
https://github.com/Botspot/pi-apps/blob/master/apps/Cura/install-32
https://github.com/Botspot/pi-apps/blob/master/.github/workflows/updates/Cura.sh

Crilum:
How does this look @theofficialgman?

Crilum:
We might consider having the version variables as Arrays too, so we're not limited to two updatable things per script.

etc/categories
```diff
@@ -128,6 +128,7 @@ Pycharm CE|Editors
 QEMU|Tools
 Quartz|Internet/Browsers
 Raspi2png|Tools
+ReaPack|Multimedia
```
Botspot, line 131: I think this change is unintended and unnecessary.

.github/workflows/updates/ReaPack.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
Botspot, line 1: As we're not adding any app named ReaPack, does this update script still belong here?

theofficialgman:
> We might consider having the version variables as Arrays too, so we're not limited to two updatable things per script.

It's not limited to two. webVer are arrays, version I kept as variables. It's automatic as necessary, version version2 version3 version4 .....

etc/categories
Crilum: Oh yeah, I accidentally based this branch off of the ReaPack pr one, so I had to remove ReaPack, but I guess when I did that, I missed the categories file..

.github/workflows/updates/ReaPack.sh
Crilum: No, I guess I missed that too... lol

.github/workflows/updates/Reaper.sh
```diff
@@ -0,0 +1,40 @@
...
+else
+        echo "\$armhf_url ($armhf_url) checks out! We're good to go."
+fi
+
+webVer[0]="${base_ver}${minor_ver}"
+echo "Latest version: ${webVer}"
+
+webVer[1]="$(get_release cfillion/reapack)"
+armhf_url[1]="https://github.com/cfillion/reapack/releases/download/v${webVer}/reaper_reapack-armv7l.so"
+arm64_url[1]="https://github.com/cfillion/reapack/releases/download/v${webVer}/reaper_reapack-aarch64.so"
```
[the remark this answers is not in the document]
Crilum: Yeah, I'll change that.

Crilum:
I think I've fixed the problems.

I also updated the description to note that this app installs ReaPack, and how to open it.

theofficialgman:
same thing as renoise, you might consider getting the version number directly from the website instead of trying every combination

```
curl -s https://www.reaper.fm/download.php | grep "linux_aarch64.tar.xz"
```
this one is more complicated though than renoise since its the actual webpage and not an index

theofficialgman:
also your current method will break once 7.x releases... because old download links remain and the 6.x link will be found first, and break out of the loop

Crilum:
@theofficialgman

theofficialgman:
not quite what I would suggest. you should only be obtaining the version info once in the update script
and you should be passing that into your download url the same way as you do in your install script.
sec and I'll push a change

Crilum:
:+1:

theofficialgman:
@Botspot do you happen to have a system with a username with spaces in it? I'm not sure if $HOME without being in quotes is safe to use or not

Crilum:
@theofficialgman I don't think you can have spaces in a linux username..

theofficialgman:
> @theofficialgman I don't think you can have spaces in a linux username..

oh no you definitely can have spaces

Crilum:
@theofficialgman 
```
$ sudo adduser "test test"
adduser: To avoid problems, the username should consist only of
letters, digits, underscores, periods, at signs and dashes, and not start with
a dash (as defined by IEEE Std 1003.1-2001). For compatibility with Samba
machine accounts $ is also supported at the end of the username
```

theofficialgman:
```
sudo useradd "test user" --badnames
```
from my memory... there are at least a few distros that when you create the username in the GUI setup they let you add a space, and it must be using this internally

theofficialgman:
and it is dangerous to do it without quotes in this case as confirmed by my test
```
$ echo $HOME
/home/test user
$ mkdir $HOME/hi
mkdir: cannot create directory ‘/home/test’: Permission denied
mkdir: cannot create directory ‘user/hi’: Permission denied
```

Crilum:
This means that all install/uninstall scripts that use `$HOME` should be changed to use `"$HOME"`?

theofficialgman:
> This means that all install/uninstall scripts that use `$HOME` should be changed to use `"$HOME"`?

yeah pretty much...

Crilum:
Well, it wouldn't be *that* hard to made a script the `sed`'s the pi-apps dir and make a PR...

theofficialgman:
> Well, it wouldn't be _that_ hard to made a script the `sed`'s the pi-apps dir and make a PR...

no need for a script... I can just search all in vscode and fix them

Crilum:
Just saying, a script is going to be way faster... There's probably thousands of `$HOME` uses in the `pi-apps` directory...

theofficialgman:
> Just saying, a script is going to be way faster... There's probably thousands of `$HOME` uses in the `pi-apps` directory...

only 300, and most of them are in quotes already which is why you can't script it easily

Crilum:
Yeah, that makes sense.. :+1:

Crilum:
@theofficialgman You'll want to get the `$USER` variables too..

theofficialgman:
ok this becomes a nightmare to fix with $DIRECTORY which is often used in .desktop files which have very cryptic rules for how to use special characters (like spaces https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html )

I can't figure it out and spaces in username isn't posix... lets just hope no users have them

Crilum:
It looks like you can just use double quotes.. 
> Arguments may be quoted in whole. If an argument contains a reserved character the argument must be quoted. ... Reserved characters are space (" ")...

theofficialgman:
> It looks like you can just use double quotes..

thats what the documentation wants you to think... but no
you can try, I already battled with it for 30 minutes and determined its not worth my time

Crilum:
We should put somewhere in the App Creation wiki to use `$HOME` and `$USER` with double quotes..

Crilum:
@Botspot or @theofficialgman Back to the PR subject, I think this is ready to merge..

Crilum:
Thanks  @Botspot!

Botspot:
I wonder how `~/` behaves with a username containing a space character. For `~/`, placing it in quotes actually prevents it from being expanded.

> We should put somewhere in the App Creation wiki to use `$HOME` and `$USER` with double quotes..

I'm not too concerned about keeping our scripts compliant with every possible edge case.
Designing around such limitations makes for complicated and hard-to-read code. For example, there are special scripting techniques for supporting folders that contain newline characters. While it may theoretically improve reliability, it's unlikely to ever be encountered in the real world, so it's just extra programming burden on developers.

For this case however, if several users start complain of broken apps due to a space-separated username, then we may decide to reconsider.

PR #1793, opened 2022-05-10
Add Renoise Updater

Crilum:


theofficialgman:
please rework this script. the renoise webpage is viewable (and hence, greppable)
https://files.renoise.com/demo/ there is not need to try every combination of version until a url is found

Crilum:
Can't believe I didn't think of that! :+1:

.github/workflows/updates/Renoise (Demo).sh
```diff
@@ -3,8 +3,9 @@
 echo "Checking Renoise version and getting download urls..."
 
 webVer=$(curl -s https://files.renoise.com/demo/ | grep "armhf" | sed -e 's/<a href="//' -e 's/".*//' -e 's/Renoise_//' -e 's/_Demo_Linux_armhf.tar.gz//')
-armhf_url="https://renoise.com/demo/Renoise_${webVer}_Demo_Linux_armhf.tar.gz"
-arm64_url="https://renoise.com/demo/Renoise_${webVer}_Demo_Linux_arm64.tar.gz"
+
+armhf_url="https://renoise.com/demo/$(curl -s https://files.renoise.com/demo/ | grep "armhf" | sed -e 's/<a href="//' -e 's/".*//')"
+arm64_url="https://renoise.com/demo/$(curl -s https://files.renoise.com/demo/ | grep "arm64" | sed -e 's/<a href="//' -e 's/".*//')"
```
theofficialgman, line 8: actually this doesn't make it more future proof at all, it makes it more dangerous
the url here is supposed to exactly match the url used in the install scripts.
now you allow the url format to change in the updater script but the format is fixed in the install scripts

revert this because if it breaks, then the github actions will notify us and that is intentional

Crilum:
@theofficialgman

PR #1798, opened 2022-05-11
Add LMMS

Crilum:
LMMS aims to be a free alternative to popular (but commercial and closed-source) programs like FruityLoops, Cubase and Logic giving you the ability of producing music with your computer by creating cool loops, synthesizing and mixing sounds, arranging samples, having more fun with your MIDI-keyboard and much more...


theofficialgman:
is the name correct? the official ubuntu and debian packages have it named `Linux Multimedia Studio`
the application seems to be named this and I can not find any reference to "lets make music"

theofficialgman:
AHAH, LOL, thats a slogan that exists only on their website, not a name at all.
Please fix to just include LMMS or LMMS (Linux Multimedia Studio)

Crilum:
Ok. I saw that (Let's Make Music) on some linux blog, lol.

theofficialgman:
also please add this to the categories file. you forgot the last couple of times.

after the new PR merges, we will likely need to make new categories to better reflect all the apps that we now have in pi-apps

Botspot:
Would it increase clarity by naming this app "Linux Multimedia Studio"?

Crilum:
Like, changing the whole name to `Linux Multimedia Studio`? Yes, but then you're not putting the real name on it.. I think it would be more clear to rename it to `LMMS (Linux Multimedia Studio)`.

Botspot:
I just looked at the website and it cleared up my concerns. The name `LMMS` is fine.

PR #1799, opened 2022-05-11
Add Waveform

Crilum:
Waveform, another DAW.
This is a 90 day free trial, but you can get a free (or `Pro` paid) license. (follow the instructions in `description` if you want one.)


apps/Waveform/install-64
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+install_packages "https://cdn.tracktion.com/file/tracktiondownload/waveform/1157/waveform_pi_64bit_v11.5.7.deb"
```
theofficialgman, line 3: technically this will still prompt an error report if install_packages errors (because is the last thing in the script) but its still better to put `|| exit 1`

Crilum: Oh yeah, sorry, I forgot about that... I usually do that, but for some reason I just missed it...

Crilum:
Thanks @Botspot!

Issue #1446, opened 2022-01-28
Split `Editors` and `Tools` Categories


theofficialgman:
### Cool ideas?

The number of apps in the editors category is getting quite large. Maybe we should consider splitting the category into multiple different categories. Examples below:

```
Code Editors
Office Suites
CAD/Manufacturing
Creative Arts Editors
```

My main issue with the current setup is I don't really think the CAD/Manufacturing programs fall into the Editors category at all:
LinuxCNC LibrePCB Eagle CAD FreeCAD Cura OpenSCAD PrusaSlicer

Similarly, the `Tools` category is massive as well but the variety there is so large I can't think of good ways to split it.

open-sorcerer64:
Yeah, I totally agree. But imo 
```
Programming (Apps like codex aren't a code editor) 
Office (Libreoffice themes and AbiWord aren't office "suites")
Graphics (can accommodate both of them)
```

open-sorcerer64:
@Botspot

Botspot:
The other day, I was thinking of a new and totally different category solution: based on the human-level action.

Categories would include "create", "connect", "listen", "watch", "write", "play", etc.
It would certainly be unique, and arguably more precise and intuitive.

Botspot:
> Maybe there could be category themes, that can be changed from Pi-Apps Settings..

Category names can be changed, so what do you mean by "Category themes"? Are you referring to the category icons found in `icons/categories`?

theofficialgman:
> The other day, I was thinking of a new and totally different category solution: based on the human-level action.
> 
> Categories would include "create", "connect", "listen", "watch", "write", "play", etc. It would certainly be unique, and arguably more precise and intuitive.

I don't like this at all...
Way too many apps would fall into multiple categories and it would be confusing for users.
For instance, in YouTube buddy you connect, listen, and watch. The Minecraft scripts, you create and play. For the editors, you create, write, etc.
It just doesn't make any sense, not intuitive at all

Crilum:
@Botspot No, like multiple `categories` files,  the categories you suggested could be one theme, and the current categories would be another theme, and you could switch them out in Settings.

PR #1813, opened 2022-05-14
add and update categories

theofficialgman:
Removed Categories (number of apps in that category):
Editors (40)

Added Categories:
Programming (19)
System Management (16)
Creative Arts (10)
CAD and Manufacturing (10)
Terminals (7)
Office (4)

New subcategory under tools:
Emulation (5)

these are just ideas. looking for feedback. 

I've considered having System Management as a subcategory under tools instead of its own primary category, same goes for Terminals. 

Programming definitely deserves its own category. Appearance has shrunk because of apps moving to better fit categories (12 apps now). 

Office and CAD and Manufacturing are on the small side, but I fee like they deserve primary categories, or otherwise put them as subcategories under tools. I wanted to move apps our of tools because it was such a massive category before, 62 apps, and shrunk to 39 apps now.

closes https://github.com/Botspot/pi-apps/issues/1446

### DIFF
--- a/etc/categories
+++ b/etc/categories
@@ -1,97 +1,97 @@
-Alacritty Terminal|Tools
-All Is Well|Tools
+Alacritty Terminal|Terminals
+All Is Well|System Management
 Amiberry|Games
 AndroidBuddy|Tools
 Angry IP scanner|Internet
 AntiMicroX|Tools
 AnyDesk|Internet
-Arduino|Editors
+Arduino|Programming
 AstroMenace|Games
 Audacious|Multimedia
 Audacity|Multimedia
-Autostar|Tools
+Autostar|System Management
 BalenaEtcher|Tools
-BleachBit|Tools
-BlockBench|Editors
-BlockPi|Editors
-BlueJ Java IDE|Editors
+BleachBit|System Management
+BlockBench|Creative Arts
+BlockPi|Programming
+BlueJ Java IDE|Programming
 Bongo Cam|Multimedia
-Box64|Tools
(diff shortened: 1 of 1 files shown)


theofficialgman:
oh oops... apparently I need to make sure the github actions wiki script only runs on the master branch

etc/categories
```diff
@@ -184,8 +184,8 @@ WhatsApp|Internet
 Windows 10 Theme|Appearance
 Windows Flasher|Tools
 Windows Screensavers|Appearance
-Wine (x86)|Tools
-WPS Office|Editors
+Wine (x86)|Tools/Emulation
```
Botspot, line 187: It's pretty sneaky of you to define an app by the exact thing its name defies. (WINE = Wine Is Not Emulator)
Not that I have a problem with this though...

Botspot:
"CAD and Manufacturing" seems long. I think it could be replaced with "Engineering."

Other than that, this looks very well thought-out. As new types of apps are added, an occasional re-factor is good.

etc/categories
theofficialgman: I know lol, but wine itself IS running under box86 which is an emulator. so I felt like it was fitting regardless

theofficialgman:
> "CAD and Manufacturing" seems long. I think it could be replaced with "Engineering."

ah duh... why didn't I just think of that

Botspot:
Let's merge this when #1580 is merged.

theofficialgman:
I picked to the GUI update. hope thats ok, I'll close this

PR #1822, opened 2022-05-15
update_apps: use github api token to get release

ryanfortner:
GitHub's API is rate-limited to 600 requests per hour, per IP. This may cause issues in the future if we have too many scripts fetching the latest release/prerelease. The limit can be increased to 5,000 requests per hour if an API token is used.

I have used this same method in the project afterburner [update scripts](https://github.com/ProjectAfterBurner/update-scripts/blob/main/api).

### DIFF
--- a/.github/workflows/update_apps.yml
+++ b/.github/workflows/update_apps.yml
@@ -32,11 +32,11 @@ jobs:
 
           #add special functions
           get_release() {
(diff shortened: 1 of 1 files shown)


Botspot:
I was wondering how this would be implemented and left the docs feeling confused. Thank you for your contribution, @ryanfortner.

theofficialgman:
oh nice, I was confused when I saw the commit popup. I guess we should probably do this at the analytics repo as well

PR #1824, opened 2022-05-16
firefox: use tee for 99bionic-updates

ryanfortner:


ryanfortner:
@Botspot if you don't mind me asking, what does `>/dev/null` do?

Crilum:
> @Botspot if you don't mind me asking, what does `>/dev/null` do?

`>/dev/null` redirects the output to /dev/null so you don't get the output of the command:
```
pi@AwePi:/tmp $ ls
 antimicrox
 anydesk
 dhcpcd-pi
 gimp
 hsperfdata_pi
'Inc+LMFiDB5H3FMuIVMTLv65UFhsAwQM1OyGOlIq0_M='
 pi-apps
 pkgs
 qipc_sharedmemory_IncLMFiDBHFMuIVMTLvUFhsAwQMOyGOlIqMe592369488333abacad329599e0a9469500c9fa3
 qipc_systemsem_dbcbeabecdfff5c3ae65d4469c947b3f33ff560f169d2f78088a
 qipc_systemsem_IncLMFiDBHFMuIVMTLvUFhsAwQMOyGOlIqMe592369488333abacad329599e0a9469500c9fa3
 scoped_dir6EdMZd
 ssh-DIBL9ADPLYhe
 ssh-EHiTvBfGmbHl
 systemd-private-00bf8db1f1b14230953a4b584b076f1e-colord.service-UFew2h
 systemd-private-00bf8db1f1b14230953a4b584b076f1e-ModemManager.service-IMgqbj
 systemd-private-00bf8db1f1b14230953a4b584b076f1e-systemd-logind.service-f2u1Xh
 systemd-private-00bf8db1f1b14230953a4b584b076f1e-systemd-timesyncd.service-h075Wi
 systemd-private-00bf8db1f1b14230953a4b584b076f1e-tor@default.service-sRdn2e
 tmp.5buLKVP1YG
 tmp.5UMiRpsbAg
 www-data-temp-aspnet-0
pi@AwePi:/tmp $ ls >/dev/null
pi@AwePi:/tmp $
```

ryanfortner:
Thanks!

Botspot:
> @Botspot if you don't mind me asking, what does `>/dev/null` do?

/dev/null is the place to send unwanted output.

In this case, we're trying to write to a root-owned file. Under normal circumstances, we could do `echo 'content' > /path/to/file`, but that would fail as this file is owned by root.
Even `sudo echo 'content' > /path/to/file` would fail as redirection (`>`) is still a non-root process, even though `echo` is a root process. One possible solution would be to start a new shell and run the entire echo command in there. `sudo bash -c "echo 'content' > /path/to/file"`, but that's a bit overkill.

To work around bash output-redirection ignoring `sudo`, we will avoid using it. The `tee` command *splits* a pipe. It will write to whatever file you give it, _and_ it will also forward the information to Standard Output. In our case, we don't want to see the standard output - we only want `tee` to write to the file without any other visible output. So the `> /dev/null` will redirect the Standard Output to this special "virtual file" that discards all data written to it.

PR #1827, opened 2022-05-16
Update AnyDesk description to be more informational

Crilum:
The current description doesn't really tell users how to run AnyDesk. It starts on boot, but I thought it (the description) could be more informational on that subject.


apps/PrusaSlicer/install-32.save
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
Botspot, line 1: I don't think you meant to create this file.

Crilum: Oh, no I didn't.. I force checked out a new branch, and I thought it would delete all the changed files, but I guess not..

PR #1846, opened 2022-05-20
feature: support movable GUI

theofficialgman:
creating as a draft since this needs testing on piOS and many DEs to see if they all behave. yes obtaining the window size as well as the headers/buffer sizes separately was necessary.

this obtains the current main yad window position, and size

this relies on the window manager not lying so it needs to be tested on multiple systems

this can be extended to allow the user to resize the main UI as well to automatically change the size of the second yad window (like if the user makes the main UI window taller/smaller, then the second window can be automatically made taller/smaller)

https://user-images.githubusercontent.com/28281419/169632418-35bf79e8-fe50-4889-a56b-c93d17755b7f.mp4

### DIFF
--- a/gui
+++ b/gui
@@ -686,8 +686,8 @@ if [[ "$guimode" == yad* ]];then
       --list --no-headers --column=:IMG  --column=Name --column=Sysname:HD --column=tip:HD --column=@fore@:HD \
       --separator='\n' --window-icon="${DIRECTORY}/icons/logo.png" \
       --tooltip-column=4 \
-      --dclick-action "env pipe=$pipe geometry2=$geometry2 '${DIRECTORY}/gui'" --select-action "env pipe=$pipe geometry2=$geometry2 '${DIRECTORY}/gui'" \
-      --button="!${DIRECTORY}/icons/search.png"!'Search':"env pipe=$pipe geometry2=$geometry2 '${DIRECTORY}/gui' '' '' Search/ Search/" \
+      --dclick-action "env pipe=$pipe '${DIRECTORY}/gui'" --select-action "env pipe=$pipe '${DIRECTORY}/gui'" \
+      --button="!${DIRECTORY}/icons/search.png"!'Search':"env pipe=$pipe '${DIRECTORY}/gui' '' '' Search/ Search/" \
       --button="!${DIRECTORY}/icons/options.png"!'Settings':2 \
       "$geometry1"
     
--- a/gui
+++ b/gui
@@ -508,37 +508,32 @@ get_positions() {
     main_yad_window=$YAD_XID
     if [ -n "${main_yad_window}" ]; then
       window_info=$(xwininfo -id $main_yad_window)
-      main_yad_window_pos=($(echo "$window_info" | grep "Corners:" | awk '{print $2}' | tr '+' '\n' | tr '-' '-\n' | sed '/^$/d'))
-      main_yad_window_width=$(echo "$window_info" | grep 'Width:' | awk '{print $2}')
-      main_yad_window_height=$(echo "$window_info" | grep 'Height:' | awk '{print $2}')
-      extents=$(xprop _NET_FRAME_EXTENTS -id "$main_yad_window" | grep "NET_FRAME_EXTENTS" | cut -d '=' -f 2 | tr -d ' ')
-      bl=$(echo $extents | cut -d ',' -f 1) # width of left border
-      br=$(echo $extents | cut -d ',' -f 2) # width of right border
-      t=$(echo $extents | cut -d ',' -f 3)  # height of title bar
-      bb=$(echo $extents | cut -d ',' -f 4) # height of bottom border
-      main_yad_window_pos[0]=$((main_yad_window_pos[0]-bl))
-      main_yad_window_pos[1]=$((main_yad_window_pos[1]-t))
-      #xdotool getwindowgeometry $main_yad_window | awk '{print $2}' | sed -n 2p | tr ',' '\n'
-      #xwininfo -id $main_yad_window | grep "Corners:" | awk '{print $2}' | tr '+' '\n' | tr '-' '-\n' | sed '/^$/d'
-      echo "${main_yad_window_pos[@]}" >/dev/stderr
-      echo "$main_yad_window" >/dev/stderr
+      # for speed of execution, we are assuming the order of matches from xwininfo
+      # this is safe as the output order has been standardized for over a decade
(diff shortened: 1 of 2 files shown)


theofficialgman:
I just want to clarify that I've marked this as a draft @Botspot because I want testing on more DEs than my own. If it works as is, then I will add the resizing window function to it as well.

Its currently been tested on budgie xorg and KDE plasma wayland and works correctly in both.

Crilum:
It doesn't really work on RPiOS 32-bit:

https://user-images.githubusercontent.com/91354257/169633828-3567a11a-8ac9-4f8c-b31b-94e0676f63da.mp4

theofficialgman:
@Crilum unrecognized arg?
can you output this please
```
xwininfo --help
```

also do 
```
xdotool --help
```

theofficialgman:
oh @Crilum you appear to have two matches for "Pi-Apps", can you please close out other instances before running this?

theofficialgman:
@Crilum btw, you shouldn't be running two instances anyway, but I have found a way to support it (see above commit)

Crilum:
Oh, lol, yeah, the would be the problem.. I'll test it again.

Crilum:
It works perfectly, even with two Pi-Apps windows open. If you put the Pi-Apps window too close to the edge of the screen, it obviously only draws the info window up to the edge of the screen, but that isn't a problem.

theofficialgman:
https://user-images.githubusercontent.com/28281419/169711511-74918d10-2846-4156-8497-2170a67f61d0.mp4

theofficialgman:
also I've tested myself on LXDE and it works fine. my LXDE has window borders non-zero (2 pixels on the left and right) which is how I discovered the border issue. pi-apps master has this problem as well

theofficialgman:
I'd like for one final test and feedback from @Botspot . then I'll remove the debug output and I will merge

Botspot:
> also I've tested myself on LXDE and it works fine. my LXDE has window borders non-zero (2 pixels on the left and right) which is how I discovered the border issue. pi-apps master has this problem as well, you can even see it in the picture on the readme https://github.com/Botspot/pi-apps/blob/master/icons/screenshots/details%20window.png?raw=true.

I noticed that when using PiXflat theme, but how much could 2 pixels matter?

> I decided to make it so the overall dual window width is constant (to the hardcoded value that botspot set), and if the user makes the main window smaller/larger, then the second window becomes larger/smaller to compensate. if the height of the main window is changed, the second window height is changed to match. if that is not desirable, that can change.

I like the idea of someone expanding/shrinking their Pi-Apps window to their liking. It may be more straightforward to expand window2 the the same percentage as window1 was expanded, or something like that.

No offence, but this whole feature seems a bit gimmicky to me. Sure, it may be nice, but if we are willing to reverse-engineer window managers, why not build a "proper" dual pane python gui at this point? I think a python gui would be a bad move, but maybe this PR is a bad move too - yet another feature that could malfunction.

theofficialgman:
> I noticed that when using PiXflat theme, but how much could 2 pixels matter?

for 2 pixels it doesn't matter much. but the window borders could be bigger which would make it more noticeable.

> I like the idea of someone expanding/shrinking their Pi-Apps window to their liking. It may be more straightforward to expand window2 the the same percentage as window1 was expanded, or something like that.

thats one thing I considered.

> No offence, but this whole feature seems a bit gimmicky to me. Sure, it may be nice, but if we are willing to reverse-engineer window managers

I wouldn't consider this reverse engineering window managers, this is just xorg spec. I feel like this is a necessity of the new GUI, I've already seen multiple users move the main window (and resize it) in photos and that pretty much breaks the experience.

>  why not build a "proper" dual pane python gui at this point? I think a python gui would be a bad move

I agree on the python not being the right move, at least not now since I don't think anyone here has experience with it, and it would be more complex.

> but maybe this PR is a bad move too - yet another feature that could malfunction.

yad always populates window id in the YAD_XID variable, I got that from the man page. The window info is always available since thats gotten directly from xorg, no window manager involved.

I say we merge, it works everywhere and actually is independent of the window manager, I just originally though it might have been until I learned more about xorg

Botspot:
> > I like the idea of someone expanding/shrinking their Pi-Apps window to their liking. It may be more straightforward to expand window2 the the same percentage as window1 was expanded, or something like that.
> 
> thats one thing I considered.

I think that would be a good change to make. Or as a simpler alternative, the width of window2 could be hardcoded rather than taking up the slack from window1.

Also, I'd like your input on whether these position settings should be saved for future launches. On the one hand, the user may appreciate it being a particular way, but this risks there being some malfunction, resulting in the window appearing partway off-screen from then on.
VLC had this exact problem a few years back on Pi OS, and with openbox WM you could bring it back with the Alt key. Not anymore with Mutter. (though mutter might prevent off-screen app-launches)

> I wouldn't consider this reverse engineering window managers, this is just xorg spec. I feel like this is a necessity of the new GUI, I've already seen multiple users move the main window (and resize it) in photos and that pretty much breaks the experience.

Good point. I guess this feature doesn't really lock us down any. I had similar concerns for @Itai-Nelken's genapplist-yad program, but when it was time to redo the app list, there was no additional difficulty in working around it.

theofficialgman:
> I think that would be a good change to make. Or as a simpler alternative, the width of window2 could be hardcoded rather than taking up the slack from window1.

yeah I think hardcoding the second window is probably a good idea. to be clear, I don't expect the main window width to change much, its just that users might change it slightly if they drag from the corner like you see in the videos (instead of on the top/bottom/sides)

> Also, I'd like your input on whether these position settings should be saved for future launches. On the one hand, the user may appreciate it being a particular way, but this risks there being some malfunction, resulting in the window appearing partway off-screen from then on.

right now the positions are not saved for future launches even within pi-apps, thats why you see when I go into settings and the main yad window closes, the main window re-opens at the center again when the user leaves the settings. I definitely do not want to save the position between the user actually closing and re-opening pi-apps, the user could have changed displays or resolutions by then.

> VLC had this exact problem a few years back on Pi OS, and with openbox WM you could bring it back with the Alt key. Not anymore with Mutter. (though mutter might prevent off-screen app-launches)

mutter will not let the window spawn with anything outside the drawable area. it moves it within the drawable area before starting. LXDE does not do this, it will let it go outside the drawable area

Crilum:
>No offence, but this whole feature seems a bit gimmicky to me. Sure, it may be nice, but if we are willing to reverse-engineer window managers, why not build a "proper" dual pane python gui at this point? I think a python gui would be a bad move, but maybe this PR is a bad move too - yet another feature that could malfunction.

@Botspot Earlier today I after testing this PR and going back to the current version, I was already missing it. I think this PR is really a good idea.

PR #1849, opened 2022-05-21
Use proper Debian Version for Flameshot

touhidurrr:
Flameshot releases separate packages for Debian 10 & Debian 11.
For example in current latest [release](https://github.com/flameshot-org/flameshot/releases/tag/v11.0.0), there is 2 armhf deb files:
[flameshot-11.0.0-1.debian-10.armhf.deb](https://github.com/flameshot-org/flameshot/releases/download/v11.0.0/flameshot-11.0.0-1.debian-10.armhf.deb) &
[flameshot-11.0.0-1.debian-11.armhf.deb](https://github.com/flameshot-org/flameshot/releases/download/v11.0.0/flameshot-11.0.0-1.debian-11.armhf.deb)
However we are only using the Debian 10 version for some reason.
This pr reads debian version from `/etc/debian_version` and uses it to provide the version number.
**EDIT:** Now it uses `/etc/os-release` for suggestions.

### DIFF
--- a/apps/Flameshot/install-32
+++ b/apps/Flameshot/install-32
@@ -1,5 +1,4 @@
 #!/bin/bash
 
 version=11.0.0
-debian_version="$(cat /etc/os-release | grep -oPm1 '(?<=VERSION=")\d+')"
-install_packages https://github.com/flameshot-org/flameshot/releases/download/v${version}/flameshot-${version}-1.debian-${debian_version}.armhf.deb || exit 1
+install_packages https://github.com/flameshot-org/flameshot/releases/download/v${get_codename}/flameshot-${version}-1.debian-${get_codename}.armhf.deb || exit 1
--- a/apps/Flameshot/install-32
+++ b/apps/Flameshot/install-32
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 version=11.0.0
-debian_version="$(cat /etc/debian_version | grep -o '^.[1-9]*')"
+debian_version="$(cat /etc/os-release | grep -om1 '[0-9]*')"
 install_packages https://github.com/flameshot-org/flameshot/releases/download/v${version}/flameshot-${version}-1.debian-${debian_version}.armhf.deb || exit 1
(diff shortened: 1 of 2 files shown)


touhidurrr:
Note: I tested [flameshot-11.0.0-1.debian-11.armhf.deb](https://github.com/flameshot-org/flameshot/releases/download/v11.0.0/flameshot-11.0.0-1.debian-11.armhf.deb) and it works.

Crilum:
I think it's better (more universal) to get the version from `/etc/os-release`.

touhidurrr:
I settled on `cat /etc/os-release | grep -oPm1 '(?<=VERSION=")\d+'` because according to `man os-release`:
```kt
VERSION=
    A string identifying the operating system version, excluding any OS
    name information, possibly including a release code name, and
    suitable for presentation to the user. This field is optional.
    Example: "VERSION=17" or "VERSION="17 (Beefy Miracle)""
```
Also about Regex `(?<=VERSION=")\d+'` it just matches the first set of digits after `VERSION=`.

ryanfortner:
Use the `get_codename` function built into the pi-apps API. check out the [obs-studio scripts](https://github.com/Botspot/pi-apps/blob/master/apps/OBS%20Studio/install-64) for an example.

ryanfortner (approved):

apps/Flameshot/install-32
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 version=11.0.0
-install_packages https://github.com/flameshot-org/flameshot/releases/download/v${version}/flameshot-${version}-1.debian-10.armhf.deb || exit 1
+install_packages https://github.com/flameshot-org/flameshot/releases/download/v${version}/flameshot-${version}-1.debian-${get_codename}.armhf.deb || exit 1
```
Botspot, line 4: This does not appear to be tested.
First of all, incorrect syntax is used to call the `get_codename` function, and also, that function will return values like 'buster' or 'bullseye', NOT 10 or 11.
Also, this change will break Flameshot for our Ubuntu users.

touhidurrr:
Sigh. Maybe I need to redo it and make sure that it doesn't break for Ubuntu also. So, closing this.

PR #1858, opened 2022-05-23
Add Notejot

Crilum:
### [Notejot](https://github.com/lainsce/notejot)

A minimalist note-taking app

This is only 64-bit because (like Dot Matrix) there are no 32-bit flatpaks.


apps/Notejot/uninstall
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+flatpak_remove io.github.lainsce.Notesjot || exit 1
```
Botspot, line 3: I think the app name is misspelled here.

Botspot:
> This is only 64-bit because (like Dot Matrix) there are no 32-bit flatpaks.

I found a thread on the flatpak forum and it appears that armhf flatpaks were deprecated due to server costs. Perhaps some people from the Pi-Apps team can organize something or offer a pi as a build server.

Alternatively, would it be possible to avoid the Flatpak middle-man and use github directly for an armhf version?

Crilum:
> > This is only 64-bit because (like Dot Matrix) there are no 32-bit flatpaks.
> 
> I found a thread on the flatpak forum and it appears that armhf flatpaks were deprecated due to server costs. Perhaps some people from the Pi-Apps team could organize funding or offer a pi as a build server. After all, other than the Raspberry Pi Foundation, we are the only other major organization with an interest in armhf linux apps.

I would donate a little.. I don't have any method of payment though, besides [Duino-Coin](https://duinocoin.com) (and I don't even have half a cent worth of it).. 🤣 

> Alternatively, would it be possible to avoid the Flatpak middle-man and use github directly for an armhf version?

I don't know, when I initially saw Dot Matrix, I tried building it (on armhf), and got some weird code syntax error, yet it works on arm64.. Maybe I just missed something..

apps/Notejot/install-64
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
```
Botspot, line 2: Please add `install_packages flatpak || exit 1`

Crilum: I added install_packages to `install-64`, as well as `purge_packages` to the uninstall script I assume that's what you want..

Botspot:
> > Alternatively, would it be possible to avoid the Flatpak middle-man and use github directly for an armhf version?
> 
> I don't know, when I initially saw Dot Matrix, I tried building it (on armhf), and got some weird code syntax error, yet it works on arm64..

That would probably vary on a case-by-case basis.

Crilum:
Most of this stuff is built with GTK4, which isn't in Debian yet, which is why these flatpaks are important..

Crilum:
Thanks, @Botspot!

PR #1860, opened 2022-05-23
Add Quilter

Crilum:
## [Quilter](https://github.com/lainsce/quilter)

Quilter is a minimalist markdown editor

Quilter is (technically) deprecated, but it still works great.


ryanfortner:
This seems like a nice app but I don’t think it’s a good idea to add an app that has been abandoned, since the maintainer won’t provide support anymore.

Crilum:
> This seems like a nice app but I don’t think it’s a good idea to add an app that has been abandoned, since the maintainer won’t provide support anymore.

Yeah, I guess that's probably true... If it starts having problems we can just take it off, but since it's a Flatpak, I don't know why it would start having problems, if it works now. I guess maybe the GNOME Platform it depends on could be eventually depreciated, and removed from Flathub..

Botspot:
All flatpak apps on pi-apps need `install_packages flatpak || exit 1` in the install script.

ryanfortner:
@Crilum I noticed two extra files in your branch, can you remove them please?
image

theofficialgman:
```
Quilter has stopped development, to merge with Apostrophe, I advise to use Apostrophe. slightly_smiling_face
```
this is from the dev as seen in the issues at quilter ^

the dev is referring to this app: https://flathub.org/apps/details/org.gnome.gitlab.somas.Apostrophe
https://gitlab.gnome.org/World/apostrophe

I would not suggest adding quilter as has been said before

PR #1867, opened 2022-05-24
Add Option to Pi-Apps Settings App

Red-exe-Engineer:
Changes will add a new option to the settings app allowing users to select their prefered terminal

Also assigns a terminal if none is found (and I hope it also detects if the terminal isn't in /usr/bin)

This has not been tested a lot but it seems to work for me, feel free to make changes if you like :p


Red-exe-Engineer:
Oh yeah, here is a video
https://user-images.githubusercontent.com/90420989/169941096-d04513ff-22af-4ccf-acef-7b6318acbd9e.mp4

etc/terminal-run
```diff
@@ -21,44 +21,83 @@ function error {
...
+  elif [ -f /usr/bin/gnome-terminal ];then
+    printf gnome-terminal > ~/pi-apps/data/settings/Terminal
+
+  elif [ -f /usr/bin/qterminal ];then
+    printf qterminal > ~/pi-apps/data/settings/Terminal
+
+  elif [ -f /usr/bin/x-terminal-emulator ];then
+    printf x-terminal-emulator > ~/pi-apps/data/settings/Terminal
+
+  else:
```
Itai-Nelken, line 54:
```suggestion
  else
```

Itai-Nelken (changes requested):

Crilum:
I think this might cause problems, because if you don't have the terminal you selected installed, it won't work. I suggest using `apt list` to see if the terminal is installed, when you save the settings.

etc/terminal-run
```diff
@@ -21,44 +21,83 @@ function error {
   exit 1
 }
 
-if [ -f /usr/bin/lxterminal ];then
+# Check if the current terminal is in /usr/bin - Wallee
+if ! [ -f /usr/bin/$(cat ~/pi-apps/data/settings/Terminal) ];then
+
+  if [ -f /usr/bin/lxterminal ];then
```
Crilum, line 27:
```suggestion
  if command -v lxterminal &>/dev/null; then
```

This will check to see if the command exists, regardless of the location, as long as it is on path.
It would probably be a good idea to use this for all the `elif`s too.

etc/terminal-run
```diff
@@ -21,44 +21,83 @@ function error {
   exit 1
 }
 
-if [ -f /usr/bin/lxterminal ];then
+# Check if the current terminal is in /usr/bin - Wallee
+if ! [ -f /usr/bin/$(cat ~/pi-apps/data/settings/Terminal) ];then
```
Botspot, line 25: The `~/pi-apps` part should be changed to $DIRECTORY. This variable allows you to rename the pi-apps folder to anything you want - something we did a lot when testing the new gui.

Botspot:
I'm not sure it is a good idea to tie the `terminal-run` script with other pi-apps components. Until now, it has been a separate script that can be used elsewhere without modification.
Making terminal-run obey the user's preference of terminal ***using this method*** would complicate the script and make it les useful outside of Pi-Apps.

I would prefer a more system-wide terminal selection approach. x-terminal-emulator could do this nicely, in which case `terminal-run` could be changed to obey the choice of x-terminal-emulator.

Botspot:
The `update-alternatives` command changes the preferred terminal, and it is simple for terminal-run to determine the user's current choice.
```
readlink -f /usr/bin/x-terminal-emulator
#or
readlink -f $(command -v x-terminal-emulator)
```

theofficialgman:
> The `update-alternatives` command changes the preferred terminal, and it is simple for terminal-run to determine the user's current choice.
> 
> ```
> readlink -f /usr/bin/x-terminal-emulator
> #or
> readlink -f $(command -v x-terminal-emulator)
> ```
> 
> To change the terminal:
> 
> ```
> sudo update-alternatives --config x-terminal-emulator
> ```

I don't see how this has anything to do with pi-apps. the default terminal emulator under x-terminal-emulator is usually left in auto mode, so that the booted DE can set its own terminal to the highest priority and use that.
changing the terminal from anything but auto will break that unless you set it back to auto.

theofficialgman:
> I'm not sure it is a good idea to tie the `terminal-run` script with other pi-apps components. Until now, it has been a separate script that can be used elsewhere without modification. Making terminal-run obey the user's preference of terminal _**using this method**_ would make it less useful outside of Pi-Apps.

More importantly, the changes here actually BREAK terminal-run used by other scripts. DIRECTORY is NOT set and and so any .desktop file or other script outside of pi-apps that calls terminal-run will fail with:
```
else
  echo "Terminal emulator not found!"
  exit 1
fi
```
given this is a breaking change, I've marked this PR as draft to prevent accidental merges

Botspot:
> I don't see how this has anything to do with pi-apps. the default terminal emulator under x-terminal-emulator is usually left in auto mode, so that the booted DE can set its own terminal to the highest priority and use that.

On Pi OS, lxterminal is the default terminal in x-terminal-emulator. But once I install xfce4-terminal, running x-terminal-emulator would now launch xfce4-terminal. Uninstalling xfce4-terminal would cause update-alternatives to fall back to lxterminal.

My DE does not appear to have any say in the default terminal. Yours may be different, @theofficialgman.
Most desktop environments treat x-terminal-emulator as the default choice of terminal, so that is what menus will use if a .desktop file says `Terminal=true`. If desktop environments can use x-terminal-emulator, I think terminal-run can too.

> changing the terminal from anything but auto will break that unless you set it back to auto.

Could you elaborate? I've manually set my preferred terminal with update-alternatives and nothing "broke".

theofficialgman:
> My DE does not appear to have any say in the default terminal. Your DE may be different, @theofficialgman. Most desktop environments treat x-terminal-emulator as the default choice of terminal, so that is what menus will use if a .desktop file says `Terminal=true`. If desktop environments can use x-terminal-emulator, I think terminal-run can too.

sorry I should have been more clear. when you install a new DE, it often comes with that DEs own terminal emulator (LXTerminal). at the time of installation, it gets added as an option to x-terminal-emulator with the highest priority.

> Could you elaborate? I've manually set my preferred terminal with update-alternatives and nothing "broke".

I'm saying that if you change the preferred terminal, this is a system wide setting, not just pi-apps, and installing new terminals or new DEs with their own new terminals will not automatically change the preferred terminal anymore.

Basically what I would worry about, if this was made a pi-apps GUI setting, would be a user opens the menu, changes the terminal (aka: setting a manual update alternatives override), and then then forgets about this or maybe even uninstalls pi-apps. Now the system default terminal has been overridden and the user doesn't know how to change it back.

etc/terminal-run
Red-exe-Engineer: Yeah I removed it for testing and I could not figure out how to reimplement it, took me 2 and a half hours to get the line to "work" in general xD

Botspot:
> I'm saying that if you change the preferred terminal, this is a system wide setting, not just pi-apps, and installing new terminals or new DEs with their own new terminals will not automatically change the preferred terminal anymore.
> 
> it just seems strange to me to make a pi-apps setting, actually change a system setting.

I think of `terminal-run` as a separate helper script that can be used in any project. Terminal-run is also in Pi-Power-Tools, and [in DOCUMENTATION.md](https://github.com/Botspot/pi-apps/blob/master/DOCUMENTATION.md#the-terminal-run-script) I say it is useful for the larger open source world. I don't think of it as a pi-apps component, but as a more capable version of x-terminal-emulator.

With that mindset, it makes perfect sense to have terminal-run inherit the preference of x-terminal-emulator.
For Pi-Apps' case, I would rather it follow system-wide preferences than to have its own terminal setting.

Red-exe-Engineer:
> > I'm saying that if you change the preferred terminal, this is a system wide setting, not just pi-apps, and installing new terminals or new DEs with their own new terminals will not automatically change the preferred terminal anymore.
> > it just seems strange to me to make a pi-apps setting, actually change a system setting.
> 
> I think of `terminal-run` as a separate helper script that can be used in any project. Terminal-run is also in Pi-Power-Tools, and [in DOCUMENTATION.md](https://github.com/Botspot/pi-apps/blob/master/DOCUMENTATION.md#the-terminal-run-script) I say it is useful for the larger open source world. I don't think of it as a pi-apps component, but as a more capable version of x-terminal-emulator.
> 
> With that mindset, it makes perfect sense to have terminal-run inherit the preference of x-terminal-emulator. For Pi-Apps' case, I would rather it follow system-wide preferences than to have its own terminal setting.

Oh, I see.

In that case the setting can use/change the system terminal :))) (maybe  IDK)

theofficialgman:
> With that mindset, it makes perfect sense to have terminal-run inherit the preference of x-terminal-emulator. For Pi-Apps' case, I would rather it follow system-wide preferences than to have its own terminal setting.

I'm fine with terminal-run and pi-apps following the system preference (just directly use x-terminal-emulator). I just don't want a pi-apps setting to be allowed to change the system preference. its not called "System Settings", its called "Pi-Apps" settings. I was under the impression that you said here: https://github.com/Botspot/pi-apps/pull/1867#issuecomment-1135399351 that you wanted to have pi-apps change that system setting

Botspot:
> I just don't want a pi-apps setting to be allowed to change the system preference. its not called "System Settings", its called "Pi-Apps" settings. I was under the impression that you said here: [#1867 (comment)](https://github.com/Botspot/pi-apps/pull/1867#issuecomment-1135399351) that you wanted to have pi-apps change that system setting

I can read my past messages and can see how you saw it that way. (It's not what I meant)
I agree that it would be bad for Pi-Apps Settings to change the update-alternatives choice of terminal.

Botspot:
I've made the changes in commit https://github.com/Botspot/pi-apps/commit/a1d068075d7f4fe34be2eac994c0ee1b2cd64961.

Good thing this PR was NOT merged. If it was, pi-apps would have been severely broken. I nearly missed this, but it looks like somehow @Red-exe-Engineer managed to change the file-permissions of `terminal-run` from 775 to 644. The 644 permissions don't allow execution, so if this had been merged Pi-Apps would not have been able to run the terminal-run script anymore.
2022-05-24-224423_1920x1080_scrot

Fortunately, if this ever occurs, it won't be disastrous. Early on I recognized a possibility of breaking the terminal implementation and made Pi-Apps update its core files outside of the terminal.

PR #1869, opened 2022-05-24
total rewrite of Doom3 script

theofficialgman:
upstream @techcoder20 was not fixing their app and the repo there is separate from pi-apps

move the install to pi-apps, fix multiple broken things, and enable 64bit (since it should have always worked). this is necessary as Doom3's broken installed sdl2 packages have been blocking multiple other pi-apps apps from being merged

edit: also, yes, apparently I can't spell lately. the branch is misspelled


theofficialgman:
a lot of these packages appear to be unnecessary, and were likely to due to compiling SDL2 which is not done anymore (its in the deb repos)

apps/Doom 3/install
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+# fix broken install due to @techcoder20
+sudo apt purge -y sdl2-image sdl2-mixer sdl2-ttf || true
```
Botspot, line 4: What would happen if these packages were not remove? Would there be conflicts?

theofficialgman: these packages don't exist in repos. they were custom made during the old install script and cause conflicts with official repo packages (due to overwriting files)

this is the main reason for the re-write

theofficialgman: > What would happen if these packages were not removed? Would there be conflicts?

oh I think I misread your question. yes they must be purged. my intention here was just to not error if the packages happen to not be installed (like on a clean system) let me fix this

Botspot:
Please hide these errors:
```
E: Unable to locate package sdl2-image
E: Unable to locate package sdl2-mixer
E: Unable to locate package sdl2-ttf
```
And also, please change the script to ask for user input right away, rather than waiting until all packages are installed before asking.

apps/Doom 3/install
```diff
@@ -0,0 +1,82 @@
...
+  cp base.so d3xp.so dhewm3 libidlib.a ~/Doom3Demo || error "Failed to copy necessary files to Doom3Demo Folder"
+  echo "[Desktop Entry]
+Version=1.0
+Type=Application
+Name=Doom3Demo
+Path=$HOME/Doom3Demo
+Icon=${DIRECTORY}/apps/Doom 3/icon-64.png
+Exec=$HOME/Doom3Demo/dhewm3
+Categories=Game;
+Terminal=false" | tee $HOME/.local/share/applications/Doom3Demo.desktop >/dev/null || error "Failed to create menu button!"
```
Botspot, line 60: Does this tee command serve a purpose here? AFAIK, tee is only necessary when writing to a root-owned file. Tee can be run with sudo, while output-redirection cannot without starting another shell.

apps/Doom 3/uninstall
```diff
@@ -1,12 +1,4 @@
...
-if [ -d ~/RPIDoom3Installer ];then
-  cd ~/RPIDoom3Installer || "Failed to Change directory"
-  echo "Running uninstall script..."
-  ./uninstall.sh || error "Failed to execute uninstall script :("
-else
-  echo "WARNING: Doing nothing as ~/RPIDoom3Installer folder does not exist!"
-fi
-
-exit 0
+sudo rm -rf ~/RPIDoom3Installer ~/Doom3GameFiles ~/Doom3Demo ~/Desktop/Doom3Demo.desktop ~/.local/share/applications/Doom3Demo.desktop ~/.local/share/applications/Doom3.desktop
```
Botspot, line 4: During an update reinstall, it's probably best to avoid deleting the user's gamefiles.

theofficialgman: I will note that this can't be helped now. because this is how the previous script did it. I can change that for the future but I believe on update, the old uninstall script is run.
is that correct?

apps/Doom 3/install
theofficialgman: I generally use this format even when writing to non-sudo owned files
its just easier to keep on format rather than changing between two

apps/Doom 3/uninstall
theofficialgman: ignore what I said. saves aren't stored in these folders. they reside in

``` "%s/.local/share/dhewm3", getenv("HOME"));```

File: apps/Doom 3/install (hunk 2 above)
Comment (Botspot) [reply apps/Doom 3/install:60]: OK that's fine. I was just curious if this was out of habit or if there was some obscure POSIX-related reason to doing it.

**Comment by Botspot** [comment]
I think this looks good. Not sure why this PR is a draft, but you can merge when ready.

**Comment by theofficialgman** [comment]
its a draft because I want somebody running piOS buster 32bit (as well as 64bit on buster or bullseye) to test out running doom3 before merging

you can use the demo, thats fine for testing.

**Comment by theofficialgman** [comment]
@Botspot keep in mind I've also marked multiple of the new PRs as draft due to Doom 3 improper sdl2 packages causing conflicts. this needs to be tested, merged, and a runonce added to fix these issues

**Comment by Botspot** [comment]
> this PR needs to be tested, merged, and a runonce added

Wow. I was not aware the current script caused so many problems! Doom 3 has 13k users, so I'm honestly surprised we haven't received more complaints about this.

**Comment by Botspot** [comment]
> and a runonce added (to remove the packages even if the user doesn't 

Does this look good?
```bash
#Doom3 app by techcoder20 installed problematic SDL2 packages. (sdl2-image sdl2-mixer sdl2-ttf)
#Nothing else uses them, and they just cause problems with many other games, so remove them
runonce <<"EOF"
  if dpkg -l sdl2-image sdl2-mixer sdl2-ttf &>/dev/null ;then
    sudo_popup apt purge -y sdl2-image sdl2-mixer sdl2-ttf
  fi
EOF
```

theofficialgman:
> ```shell
>   if dpkg -l sdl2-image sdl2-mixer sdl2-ttf &>/dev/null ;then
>     sudo_popup apt purge -y sdl2-image sdl2-mixer sdl2-ttf
>   fi
> ```

this only works if all of them are installed. for the edge case where only one or two of them are installed, thats why I did what I did in the script (I know its ugly)
https://github.com/Botspot/pi-apps/pull/1869/files#diff-2654fa8ffaa944920866f93e087b47b91ecfdab13e23f965fa27722bd8bd7f89R14-R26

Botspot:
> > ```shell
> >   if dpkg -l sdl2-image sdl2-mixer sdl2-ttf &>/dev/null ;then
> >     sudo_popup apt purge -y sdl2-image sdl2-mixer sdl2-ttf
> >   fi
> > ```
> 
> this only works if all of them are installed. for the edge case where only one or two of them are installed

Good point. If any of the 3 packages are missing then dpkg will return a non-zero exit code. I will do this then:
```bash
runonce <<"EOF"
  if dpkg -l sdl2-image &>/dev/null || dpkg -l sdl2-mixer &>/dev/null || dpkg -l sdl2-ttf &>/dev/null ;then
    sudo_popup apt purge -y sdl2-image sdl2-mixer sdl2-ttf
  fi
EOF
```

theofficialgman:
> > > ```shell
> > >   if dpkg -l sdl2-image sdl2-mixer sdl2-ttf &>/dev/null ;then
> > >     sudo_popup apt purge -y sdl2-image sdl2-mixer sdl2-ttf
> > >   fi
> > > ```
> > 
> > 
> > this only works if all of them are installed. for the edge case where only one or two of them are installed
> 
> Good point. If any of the 3 packages are missing then dpkg will return a non-zero exit code. I will do this then:
> 
> ```shell
> runonce <<"EOF"
>   if dpkg -l sdl2-image &>/dev/null || dpkg -l sdl2-mixer &>/dev/null || dpkg -l sdl2-ttf &>/dev/null ;then
>     sudo_popup apt purge -y sdl2-image sdl2-mixer sdl2-ttf
>   fi
> EOF
> ```

@Botspot that still will not work. purge requires that all of those packages be installed in order to purge. again, thats why I did what I did

Botspot:
I've added the runonce in commit https://github.com/Botspot/pi-apps/commit/b0dbf99637cc1680a31271fd2a0fc416239b7e64

Can this PR be merged now?

Botspot:
> its a draft because I want somebody running piOS buster 32bit (as well as 64bit on buster or bullseye) to test out running doom3 before merging

I forgot to mention earlier: I tested this PR and it worked fine, so merging this now.

theofficialgman:
> I forgot to mention earlier: I tested this PR and it worked fine, so merging this now.

ah... well thats what I was waiting on. with the amount of users doom3 has, we shouldn't have to wait long if this doesn't work for some reason

PR #1871, opened 2022-05-24
Simplify installation script of Unciv

touhidurrr:
Doesn't create `Unciv.sh` anymore, Instead use `Path=Working Directory`
Also Considering starting to use Temurin 11 if @Botspot Agree's
Otherwise merge as is

### DIFF
--- a/apps/Unciv/install
+++ b/apps/Unciv/install
@@ -61,22 +61,15 @@ mv /tmp/Unciv.jar ~/Unciv/Unciv.jar || exit 1
 echo '- removing /tmp/Unciv.jar'
 rm -f /tmp/Unciv.jar
 
-status "Creating runner script ($HOME/Unciv/Unciv.sh)"
-# This is necessary because some special characters are supported in Desktop Entry Files.
-# Without running Unciv.jar from ~/Unciv, the Game Data will be saved in ~ by default (not ~/Unciv).
-# see https://github.com/yairm210/Unciv/blob/master/desktop/linuxFilesForJar/Unciv.sh
-echo "\
-cd $HOME/Unciv
-$(echo /usr/lib/jvm/temurin-8-*/jre/bin/java) -jar ./Unciv.jar" \
-> ~/Unciv/Unciv.sh || error "Failed to write $HOME/Unciv/Unciv.sh!"
-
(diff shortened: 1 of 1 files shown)


Botspot:
This looks good, but changing the `install` script in any way will force a reinstall of Unciv on everyone's system who installed it from Pi-Apps.

Does a code cleanup like this really matter enough to have everybody reinstall it?
If you think so, then I can merge it.

touhidurrr:
> Does a code cleanup like this really matter enough to have everybody reinstall it?
If you think so, then I can merge it.

I want to make it use Temurin 11 also, I doubt any app in Pi-Apps except this uses Temurin 8. So, doing so will save space for everyone. Then we can make the exec command `java -jar Unciv.jar`.

Botspot:
> I want to make it use Temurin 11 also, I doubt any app in Pi-Apps except this uses Temurin 8. So, doing so will save space for everyone. Then we can make the exec command `java -jar Unciv.jar`.

Your reasons make sense, but I don't see those changes here yet. Are you going to make them?

touhidurrr:
> Your reasons make sense, but I don't see those changes here yet. Are you going to make them?

I was asking if I should add it.
Now done with https://github.com/Botspot/pi-apps/pull/1871/commits/1fa458cb9d434fa1cc8b3c203db1a93eb56a1913.
Anyways. If no app uses temurin-8-jdk then should we remove it also?

theofficialgman:
> I want to make it use Temurin 11 also, I doubt any app in Pi-Apps except this uses Temurin 8. So, doing so will save space for everyone. Then we can make the exec command `java -jar Unciv.jar`.

actually nothing uses temurin 11 in pi-apps. currently on minecraft java server uses temurin 8 or 17 (depending on user selection) and multimc uses temurin 8 and 17

touhidurrr:
> actually nothing uses temurin 11 in pi-apps. currently on minecraft java server uses temurin 8 or 17 (depending on user selection) and multimc uses temurin 8 and 17

Oh, then I guess I its not necessay.

theofficialgman:
and a few apps also use `openjdk-11-jre` or `openjdk-11-jdk` since this version isn't broken in raspbian (unlike openjdk-8 which was broken on piOS 32bit with 64bit kernel)

Issue #1865, opened 2022-05-23
PeaZip


ryanfortner:
### What is the name of the app?

PeaZip

### (Optional) Where is the app hosted?

https://github.com/ryanfortner/peazip-arm/releases/download/8.6.0-fp2/peazip_8.6.0_fp1_arm64.deb

I'm opening an issue instead of a pr first because I would like for the package to be tested by more users than just me beforehand. Also, there is an option to use Flatpak for installing instead of a deb, if people would prefer that.

### About the app

PeaZip is a powerful free and open-source archive utility. It supports over 200 archive formats and counting. https://peazip.github.io/

![image](https://peazip.github.io/peazip.png)

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.
- [X] I have confirmed that this app can run on Raspberry Pi.

theofficialgman:
```
Unpacking peazip (8.6.0~fp) ...
dpkg: error processing archive /tmp/apt-dpkg-install-gne6sK/5-peazip_8.6.0_fp_arm64.deb (--unpack):
 trying to overwrite '/usr/include/zdict.h', which is also in package libzstd-dev:arm64 1.3.3+dfsg-2ubuntu1.2
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
Errors were encountered while processing:
 /tmp/apt-dpkg-install-gne6sK/5-peazip_8.6.0_fp_arm64.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

the package is improperly packaged with conflicting files. this can not be installed in pi-apps as it will cause issues on upgrade. please notify the upstream developers to fix their deb

ryanfortner:
> ```
> Unpacking peazip (8.6.0~fp) ...
> dpkg: error processing archive /tmp/apt-dpkg-install-gne6sK/5-peazip_8.6.0_fp_arm64.deb (--unpack):
>  trying to overwrite '/usr/include/zdict.h', which is also in package libzstd-dev:arm64 1.3.3+dfsg-2ubuntu1.2
> dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
> Errors were encountered while processing:
>  /tmp/apt-dpkg-install-gne6sK/5-peazip_8.6.0_fp_arm64.deb
[...]

PR #1873, opened 2022-05-24
add peazip app

ryanfortner:
Closes https://github.com/Botspot/pi-apps/issues/1865


Botspot:
I think `install` needs to be renamed to `install-64`, as I justr tried this on my armhf system and the error was `Nothing matches io.github.peazip.PeaZip in remote flathub`

ryanfortner:
> I think `install` needs to be renamed to `install-64`, as I just tried this on my armhf system and the error was `Nothing matches io.github.peazip.PeaZip in remote flathub`

Done. Thanks for noticing that @Botspot .

I think this is ready to be merged. I've tested it on ubuntu 20.04 arm64 and afterburnermod, as well as rpios buster 64 and all work flawlessly.

theofficialgman:
looks good and I think its fine to merge (worked fine on bionic when I tested it)

but just a thought for the future, since this is "bascially" a "package app" do you think we should add a specifically script name for for flatpaks? we have install(-32/-64) and package, but maybe we should add a flatpak

just a thought incase pi-apps starts getting lots of flatpaks

Botspot:
> looks good and I think its fine to merge (worked fine on bionic when I tested it). the only thing I noticed is the gtk theme doesn't match my theme, but thats a flatpak issue in general
> 
> but just a thought for the future, since this is "bascially" a "package app" do you think we should add a specifically script name for for flatpaks? we have install(-32/-64) and package, but maybe we should add a flatpak
> 
> just a thought incase pi-apps starts getting lots of flatpaks

The reason for package-apps was two-fold: first it allowed for more efficient app-creation, but also it allowed for additional features that would be impossible with script-type apps.
With the limited nature of package-apps, it becomes trivial to check which ones are already installed, thereby setting app status accordingly. This is handled by `refresh_pkgapp_status`.

A similar argument could be made for flatpak apps: make a flatpak-app type so it's possible to detect which ones are installed with a similar `refresh_flatapp_status` function. We will have to discuss this and decide if the additional complexity of a third app_type is worth it. For now, flatpak is less prevalent than apt.

Issue #674, opened 2021-05-21
make vivaldi run faster and better


Itai-Nelken:
Idea by [craigevil from the botspot discord server #general channel](https://discord.com/channels/770629697909424159/770629697909424162/845129611574640641)

change the 'Exec' part of the desktop shortcut to: `Exec=/usr/bin/vivaldi-stable --use-gl=desktop %U`

Pythonic456:
Seems to change nothing except to output this error:
`[13092:13092:0525/190818.091097:ERROR:gl_surface_presentation_helper.cc(259)] GetVSyncParametersIfAvailable() failed for 1 times!`
And repeats this every minute or so.
Is there anything else I need to change?

Itai-Nelken:
I don't know..

theofficialgman:
closing, performance in browsers isn't a real concern of pi-apps.
the browser gets to decide which backend is most performant and stable for a platform (whether that be a desktop GL backend, gles backend, or vulkan)

Botspot:
In addition, `--use-gl=desktop` started causing problems for me on RPi's modified Chromium. It took a long time to identify the cause of my problems.

A similar thing could happen in the future with Vivaldi.

However I wouldn't go as far to say that Pi-Apps is not interested in performance web-browsing. In the future I may release an app to boost Chromium performance on the older RPi3 model. (HW video deciding, smooth 720p youtube, partial RAMfs chromium cache & config, and more)

PR #1877, opened 2022-05-25
vivaldi: proprietary media fix

ryanfortner:
Closes https://github.com/Botspot/pi-apps/pull/1382
Possibly reopens https://github.com/Botspot/pi-apps/issues/674

I'm creating a new PR from the original one because I haven't gotten a response from @open-sorcerer64 for a long while, they seem to be inactive.


theofficialgman:
I don't think this is a good idea. vivaldi is a chromium based browser. as you know a special patched ffmpeg source is compiled along with chromium browsers and is called libffmpeg. the chromium patches actually change quite a bit between versions, to the point where a previous version of libffmpeg will no longer work with newer browsers depending on the api changes.

I don't think this is a good idea. if you want vivaldi to be built with the proprietary video/audio formats, you should ask upstream to enable it in their builds or provide an explanation for why they don't. this "fix" is just guaranteed to fail after some time and leave the user with NO functional decoding

ryanfortner:
Thanks, I'll take a look.

PR #1889, opened 2022-05-27
box86: use /usr/share/keyrings

ryanfortner:
`apt-key` is deprecated in Debian Bullseye, so we'll need to save the key in `/usr/share/keyrings` now.


theofficialgman:
I think Botspot wanted to add a function for this. it hasn't been added yet (to my knowledge)

we have quite a few scripts that still need this done. might be best to lump them all into one PR.

PR #1890, opened 2022-05-27
add jetson overrides file

theofficialgman:
add `category-overrides-jetson-18.04`

this system extends the current overrides format allowing pi-apps developers to hide (or create custom categories) for apps.

other systems or OSs can be added by simply creating their own custom category-overrides file, this could also be used to hide apps on buster or bullseye in general if desired. currently only one model override is allowed per device

~~I don't have time (till next week) to respond about this PR, but~~ this was my idea for implementing an overrides file that can be easily changed. Like if new apps are added that are incompatible or if app compatibly is brought to an app that was previously incompatible.

I went through the apps list a couple weeks ago and found all of these apps either non-functional (pi-specific, libraries/dependencies issues, etc) or not desirable.

Some of them I already have fixes that can easily be integrated into pi-apps, but for now they are hidden.

Others will likely be re-enabled if I implement the "buildscript" alternative install method for these apps.

here was my reasoning for disabling the apps. if an app doesn't have a description in (), its was disabled due to dependency issues due to being built on buster.

<details><summary>Disabled Apps</summary>
<p>

```
eDEX-UI
Lightpad
Mac OS Theme
Windows 10 Theme
Windows Screensavers
XSnow
BlockBench
BlockPi
Boxy SVG
Codex
Cura
Drawing
OpenSCAD
Processing IDE
Remarkable
StackEdit
Sublime Text
Amiberry
Minecraft Pi (Modded)
PPSSPP (PSP emulator)
Stunt Rally
Tetris CLI
Caprine (incorrect ownership of chrome-sandbox, needs upstream update)
Cawbird
Deskreen
Zoom
Pale Moon
Quartz
FreeTube (crappy compared to other options we have)
OBS Studio
Vivaldi (not available)
Renoise (Demo)
YouTubuddy (crappy compared to other options we have)
AntiMicroX
Autostar (unnecessary, other better options are pre-installed on ubuntus DE)
BalenaEtcher (incorrect chrome sandbox and other issues, needs upstream update)
Box64 (build on buster so is incompatible, and scripted through the megascript)
Box86 (no armhf gpu libs, will not run GUI apps so not helpful)
CommanderPi
Flameshot
KeePassXC
More RAM (no need, we have zram setup as part of initramfs)
Node.js (too hacky, nodejs isn't hard to install through the official instructions)
PiKISS GUI (don't want another app store to regulate)
piKiss (don't want another app store to regulate)
Pi Power Tools (pi only stuff)
QEMU
Raspi2png
Snapdrop
Snap Store (no, snaps do not have hardware accelerated gpu on non-mesa)
Temps (no, arm64 only for jetsons)
USBImager
VeraCrypt
Windows Flasher
```


Botspot:
We already have two ways to selectively hide apps for compatibility reasons.

1. install-32 and install-64 scripts indicate what CPU architecture an app can be installed on. If the app is not compatible, it will be hidden when running `list_apps cpu_installable`.
2. For TwisterOS, certain apps were hidden to prevent users from messing up their systems. This was achieved with a `runonce` and I'm not too fond of using this method further.

One advantage to using separate scripts to denote compatibility was that **it avoided a centralized list** of apps compatible on each platform. All information vital to the app is self-contained.
This allows the possibility for "sideloading" 3rd-party repositories for apps that we cannot officially endorse for one reason or another.
For a while, apps in Pi-Apps were completely decentralized. But now with categories, there is a centralized `etc/categories` file. 

That being said, I'm not too fond of the idea of maintaining multiple centralized categories files when it would be better to have none at all.
Perhaps each app could contain a list of incompatible systems?

theofficialgman:
> Perhaps each app could contain a list of incompatible systems?

I think the problems with that would be:
1. potential speed issues going through every app folder checking if it is compatible or not
2. each app needs an additional file rather than just maintaining one list
3. what is a "system"? we don't currently have system names or definitions. this update checks for a "jetson" as well as a matching ubuntu version "18.04" and defines that as a system. but this method doesn't have to be used for all systems. you add an additional elif  to define a unique system that maybe just has pi in the name, or switch, or bionic, or buster, or bullseye, or twisterOS.

I don't plan on using this file to specify unique categories (though it could be used for that) but more of a "disabled/hidden" apps list. This was the simplest and most straightforward way of accomplishing that.

theofficialgman:
> 2. For TwisterOS, certain apps were hidden to prevent users from messing up their systems. This was achieved with a `runonce` and I'm not too fond of using this method further.

I agree, I don't like that method (writing to the users overrides file). it would be better, imho, to make a list of twister such as the one here.

theofficialgman:
> it would be better to have none at all.

if you have none, then you have no categories. you must replace it with something. so is your idea there to make an additional two files in each app? one for the intended category
eg:
```
Internet/Browsers
```
and then one for incompatible systems
eg:
```
jetson-18.04
```

this has the downside of if we ever want to change the category name, we have to change each file in every app folder

theofficialgman:
> Yes, I was thinking of something along those lines.

ok... I still feel like a central file (like is done now with categories) is better but thats your choice.

I suggest not having an `incompatible` file but rather an `override` file

```
jetson-18.04|hidden
```
before the `|` is the system name, after the `|` is the action. this allows for apps to be moved around in categories (or a custom category created) if desired, rather than just being an "incompatible/hidden" toggle list


if we are talking about implementations, I'm going to suggest that we continue to use a master categories file that contains all of the app category information, but instead of shipping it with pi-apps, it is generated at runtime based on the `category` and `override` files inside each app directory. this would mean less script rewriting and would only require changing the function `read_category_files`

theofficialgman:
however.. changing to an inside an app folder implementation as opposed to a central file will trigger an update for every single app... which users could disable if they wanted (de-select the checkbox) and result in apps missing category data.

Botspot:
I've been thinking this over, and I had an idea. On Pi-Apps, there are and always will be apps that are specifically for the Raspberry Pi. On another ARM Linux system, RPi-specific apps won't work or won't be relevant.
So, rather than hiding problematic apps per non-pi system, why not try to remain more general and hide all apps that are not compatible with a generic ARM OS?
That would seem to take care of most cases, and when more granularity is required it could be accomplished with a `runonce`.
Some apps may also be improved to make them compatible with other SBCs.

An explanation of my hesitation for this PR:
In the long run, the only way I see for Pi-Apps to continue its expansion is by reducing overhead. And the best way to do that is to streamline the app-addition process and allow apps to be imported from 3rd-party Pi-Apps repositories. Once this happens, each Pi-Apps user may have varying app lists, as it would no longer be centralized. At that point, it would not be possible for us to maintain a centralized categories file.
Therefore, adding more categories files would seem to solve a short-term problem but worsen a long-term one. This is why I'm not too fond of this PR, even though it solves a tangible problem.

theofficialgman:
> That would seem to take care of most cases, and when more granularity is required it could be accomplished with a runonce.
Some apps may also be improved to make them compatible with other SBCs.

no, a runonce is bad, as you have already said with twister. having a file which is updatable (for when app support is dropped for a system or added for a system) is necessary.

> In the long run, the only way I see for Pi-Apps to continue its expansion is by reducing overhead. And the best way to do that is to streamline the app-addition process and allow apps to be imported from 3rd-party Pi-Apps repositories. Once this happens, each Pi-Apps user may have varying app lists, as it would no longer be centralized. At that point, it would not be possible for us to maintain a centralized categories file. Therefore, adding more categories files would seem to solve a short-term problem but worsen a long-term one. This is why I'm not too fond of this PR, even though it solves a tangible problem.

I have already stated why I don't like this. pi-apps hosts centralized apps. currently, we allow users to select and deselect which apps they want to update. if we decide to change the category system such that a each app folder contains its own "category" data, then we risk a user having lots of broken app categories on update to this new system: https://github.com/Botspot/pi-apps/pull/1890#issuecomment-1142282956 . once again, the current update system is our enemy (and we can't really change it)

here is my solution which I think is the best of both. how about we keep a "centralized" category file system for apps within the "pi-apps repo" (this github repo). For other apps that a user imports into pi-apps (3rd party repos, however that gets implemented) we allow them to have their own per app category data in the app folder itself. this solves the "update" issue described above, and allows us to keep easily defining our central category file.

keep in mind I am aware this PR solves a temporary issue, but it also doesn't impose a permanent solution. its a very low tech and easily changeable solution on purpose. my goal here is to make pi-apps a more "recommendable" solution for switch/jetson/bionic users as soon as possible. I don't want to point users to something where 60 apps are broken or irrelevant. that is why this PR exists

api
```diff
@@ -1889,6 +1898,54 @@ get_device_info() { #returns information about current install and hardware
   
 }
 
+get_model() {
```
Botspot, line 1901: Is this function specifically for Jetson devices? In any case, it's probably best to explain its usage in a comment.

theofficialgman: its a partial copy of the get_device_info function but with extra stuff that slowed it down removed.

the below part is just for detecting if we are running on a jetson, yes, which I wrote originally for the megascript and retropie.

theofficialgman: added with https://github.com/Botspot/pi-apps/pull/1890/commits/081e8219c3b273432acc677795e6ed362019c8ef

Botspot: While better, this function does not have a summary stating what it does or how it's used.
I assume that it is not outputting to stdout because you set the `$model` _and_ `$jetson_model` variables?

Botspot:
I know that some of the apps you've listed would fail on non-Pi hardware, but what about others like Windows Flasher, More RAM, Cura, Boxy SVG, Snapdrop, Flameshot, or Lightpad?
Are some of the apps in this list merely considered irrelevant on your platform?

theofficialgman:
Refer to my original post. It has a drop-down with the apps that are disabled and the reason for why https://github.com/Botspot/pi-apps/pull/1890#issue-1250321116

api
theofficialgman: right, the idea was this would be a function that could just be used to get some identifying model information for the hardware the user is running. right now its just the generic model and specifically the jetson_model (if applicable)

but it could be extended as necessary kinda like what is done over at retropie for the systems they support: 
https://github.com/RetroPie/RetroPie-Setup/blob/8e2fda6a2273222f50a39b490cec85cfbc3cba2f/scriptmodules/system.sh#L349-L437

Botspot:
> > what about others like Windows Flasher, More RAM, Cura, Boxy SVG, Snapdrop, Flameshot, or Lightpad?
> Refer to my original post. It has a drop-down with the apps that are disabled and the reason for why [#1890 (comment)](https://github.com/Botspot/pi-apps/pull/1890#issue-1250321116)

Apart from More RAM, you do not appear the reasons for removing any of the other apps I asked about.

theofficialgman:
for the apps that don't have a specific reason, they don't work due to dependency issues. the majority are because they are compiled on buster and have glibc 2.28+ requirement (bionic only has glibc 2.27).

they would work if recompiled on bionic (and would still be compatible with buster/bullseye). for example, I have already asked the cura dev to compile on bionic to solve this and they plan on doing it when they get the time https://github.com/smartavionics/Cura/issues/175

Botspot:
I guess that just leaves Windows Flasher. I suppose you removed it because there was little point in including it?

theofficialgman:
also windows flasher should be self explanatory
https://github.com/Botspot/pi-apps/blob/master/apps/Windows%20Flasher/description

I don't want users to get confused and think that they can run windows on the jetson/switch

Botspot:
I gave this more thought and now agree that it doesn't pose any hindrance to the future plans of decentralization.
This file won't have to be updated with every new app, so although it's really a category file, it works like a hidelist because the main categories file is still used.

Rak1ta:
@theofficialgman

> hide apps on buster or bullseye in general

Will removed apps be brought back due to lack of Bullseye support? If it is now possible to hide applications. 
https://github.com/Botspot/pi-apps/blob/90cd31fbe0187e9acb72040ebae81207ec43cfff/gui#L383

PR #1900, opened 2022-05-27
Add Linux Wifi Hotspot

Crilum:
## [Linux Wifi Hotspot](https://github.com/lakinduakash/linux-wifi-hotspot)
A tool to create and manage Wifi Hotspots.

I noticed as I was adding this to the `categories` file, that `LinuxCNC` is still in there.. Just thought I'd mention that.

(Wow, I made the 1900th PR/Issue..)


Crilum:
@Botspot or @theofficialgman I think this is mergeable..

theofficialgman:
tested, works fine

apps/Linux Wifi Hotspot/uninstall
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+#Allow packages required by this app to be uninstalled
+purge_packages || exit 1
+
+# Uninstall Linux Wifi Hotspot
+cd "$HOME"/linux-wifi-hotspot || error "Failed to cd into 'linux-wifi-hotspot'!"
```
Botspot, line 7: This will fail if the user first removes ~/linux-wifi-hotspot.
Also, this uninstall script does not remove ~/linux-wifi-hotspot.

A better solution would be to remove this folder during install, then make the uninstall script replicate whatever `sudo make uninstall` would do. This is easily done by looking removing all files mentioned in the folder's `installManifext.txt` file. (But please don't try to read that file on uninstall)

theofficialgman: there isn't an installmanifest (this app isn't quite a traditional cmake one)

it has hardcoded scripts
```
uninstall:
	rm -f $(DESTDIR)$(BINDIR)/create_ap
	rm -f $(DESTDIR)/etc/create_ap.conf
	rm -f $(DESTDIR)$(PREFIX)/lib/systemd/system/create_ap.service
	rm -f $(DESTDIR)$(PREFIX)/share/bash-completion/completions/create_ap
	rm -f $(DESTDIR)$(PREFIX)/share/doc/create_ap/README.md
	rm -f $(DESTDIR)$(BINDIR)/wihotspot
```

Botspot:
If this app does what I think it does, I will be quite happy. Several times I thought it would be useful to quickly share my Pi's Ethernet connection with a nearby device, but was disappointed with the complexity of online tutorials. Some even said to download a custom OS for a WiFi hotspot!

Crilum:
> If this app does what I think it does, I will be quite happy. Several times I thought it would be useful to quickly share my Pi's Ethernet connection with a nearby device, but was disappointed with the complexity of online tutorials. Some even said to download a custom OS for a WiFi hotspot!

Yeah, I had the same problem.. Luckily I found this app! You can even share your WiFi, according to the README, although I think I had a problem with it when I tried it on a Pi..

apps/Linux Wifi Hotspot/uninstall
Crilum: I think I fixed it with [f51846f](https://github.com/Botspot/pi-apps/pull/1900/commits/f51846f59491e2b74418bcc2878fc86649ed167b).. Does this look good?

theofficialgman:
I was very surprised that this app "just worked" on my switch. considering we use custom broadcom firmware pulled from the default SwitchOS, i didn't even realize hotspots were possible with it.

apps/Linux Wifi Hotspot/install
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Install dependencies
+install_packages bash util-linux procps hostapd iproute2 iw haveged dnsmasq iptables make gcc g++ build-essential pkg-config libgtk-3-0 libgtk-3-dev libqrencode-dev libpng-dev || exit 1
+
+# Clone and cd into repo
+cd "$HOME"
+git_clone https://github.com/lakinduakash/linux-wifi-hotspot || error "Failed to clone repository!"
+cd linux-wifi-hotspot || error "Failed to cd into 'linux-wifi-hotspot'!"
```
theofficialgman, line 9: now that this folder isn't used for the uninstall. can you make the building happen in the /tmp folder and delete this folder once the install is done?

Crilum:
@Botspot and/or @theofficialgman Do you want me to remove `LinuxCNC` from the `categories` file?

theofficialgman:
> @Botspot and/or @theofficialgman Do you want me to remove `LinuxCNC` from the `categories` file?

you can leave it. linuxCNC may come back in the future as a package app. the devs are currently working on bringing it to debian officially and its currently build in debian unstable for many architectures

Botspot:
This worked for me. The only hiccup I encountered was that it failed to create the hotspot when I had turned WiFi off from the panel.

I'll try to investigate how this works and see if a fix can be made.

Botspot:
> I'll try to investigate how this works and see if a fix can be made.

Found it in RPi's lxpanel lxplug-network [source code](https://github.com/raspberrypi-ui/lxplug-network/blob/9d1cf8c553ee2bb34f00693a88c3b6ae7f1a953d/plugins/dhcpcd-gtk/menu.c#L540-L546).
The command is:
```
/usr/sbin/rfkill unblock wifi
```
This looks like a RPi-specific command, though I still think it would be useful enough to add it to the .desktop file, for those of us who had WiFi disabled beforehand.

PR #1902, opened 2022-05-28
api: add add_external_repo and rm_external_repo

ryanfortner:
As suggested in https://github.com/Botspot/pi-apps/pull/1889, this is a function that adds external apt repositories when provided:

- an external repo path (such as `https://ryanfortner.github.io/box64-debs/debian ./`) 
- a repo name (such as `box64`)
- a url to the repo's gpg pubkey (such as `https://ryanfortner.github.io/box64-debs/KEY.gpg`)

Example of the usage of these functions:
```bash
add_external_repo "https://ryanfortner.github.io/box64-debs/debian ./" "box64" "https://ryanfortner.github.io/box64-debs/KEY.gpg" || exit 1

rm_external_repo "box64" || exit 1 # this will remove the repo if nothing is used in it
rm_external_repo "box64" "force" || exit 1 # this will remove the repo even if software is installed from it
```

This is my first try at creating a function for pi-apps and I would appreciate any feedback I receive!

### DIFF
--- a/api
+++ b/api
@@ -207,10 +207,16 @@ remove_repofile_if_unused() { #Given a sources.list.d file, delete it if nothing
   [ -z "$file" ] && error "remove_repo_if_unused: no sources.list.d file specified!"
   #return now if the list file does not exist
   [ -f "$file" ] || return 0
-  
-  #determine what repo-urls are in the file
-  local urls="$(cat "$file" | grep -v '^#' | tr ' ' '\n' | grep '://')"
-  
+
+  if [ "${file##*.}" == "list" ]; then
+    #determine what repo-urls are in the file
+    local urls="$(cat "$file" | grep -v '^#' | tr ' ' '\n' | grep '://')"
+  elif [ "${file##*.}" == "sources" ]; then
+    #determine what repo-urls are in the file
+    local urls="$(cat "$file" | grep -v '^#' | grep '^URIs: ' | sed 's/URIs: //g')"
+  else
+    error "$file was not of apt list or sources type"
+  fi
   #there could be multiple urls in one file. Check each url and set the in_use variable to 1 if any packages are found
   local IFS=$'\n'
   local in_use=0
@@ -225,7 +231,7 @@ remove_repofile_if_unused() { #Given a sources.list.d file, delete it if nothing
   if [ "$in_use" == 0 ] && [ "$testmode" == test ];then
     echo "The given repository is not in use and can be deleted:"$'\n'"$file" 1>&2
   elif [ "$in_use" == 0 ];then
-    status "Removing the $(basename "$file" | sed 's/.list$//g') repo as it is not being used"
+    status "Removing the $(basename "$file" | sed 's/.list$//g' | sed 's/.sources$//g') repo as it is not being used"
     sudo rm -f "$file"
     [ -f "$key" ] && sudo rm -f "$key" || true
   fi
@@ -704,19 +710,25 @@ debian_ppa_installer() { #setup a PPA on a Debian distro. Arguments: ppa_name di
 }
 
 add_external_repo() { # add an external apt repo and its gpg key.
-  # follows https://wiki.debian.org/DebianRepository/UseThirdParty specification with one-line-style format https://repolib.readthedocs.io/en/latest/deb822-format.html#one-line-format
+  # follows https://wiki.debian.org/DebianRepository/UseThirdParty specification with deb822 format https://repolib.readthedocs.io/en/latest/deb822-format.html
+  # required inputs
   local reponame="$1"
-  # "URI SUITE COMPONENT"
-  local repopath="$2"
-  local pubkeyurl="$3"
+  local pubkeyurl="$2"
+  local uris="$3"
+  local suites="$4"
+  # potentially optional inputs
+  # components is not used when suite is an absolute path
+  local components="$5"
+  # additional options can be specified as the 6th, 7th, 8th, etc argument (eg: "Architectures: arm64")
 
-  # check if all needed vars are set
-  [ -z "$repopath" ] && error "add_external_repo: repopath not set"
+  # check if all needed vars are set  
   [ -z "$reponame" ] && error "add_external_repo: reponame not set"
+  [ -z "$uris" ] && error "add_external_repo: uris not set"
+  [ -z "$suites" ] && error "add_external_repo: suites not set"
   [ -z "$pubkeyurl" ] && error "add_external_repo: pubkeyurl not set"
 
-  # exit if reponame contains space, since apt doesn't accept .list files with spaces in filename or keyname.
-  if [[ $reponame = *" "* ]]; then
+  # exit if reponame or uri or suite contains space
+  if [[ $reponame = *" "* ]] || [[ $uris = *" "* ]] || [[ $suites = *" "* ]]; then
     error "add_external_repo: provided reponame contains a space."
   fi
 
@@ -733,6 +745,11 @@ add_external_repo() { # add an external apt repo and its gpg key.
     sudo rm -f /etc/apt/sources.list.d/${reponame}.list || error "add_external_repo: failed to remove conflicting .list file."
   fi
 
+  # check if .sources file already exists
+  if [ -f /etc/apt/sources.list.d/${reponame}.sources ]; then
+    sudo rm -f /etc/apt/sources.list.d/${reponame}.sources || error "add_external_repo: failed to remove conflicting .sources file."
+  fi
+
   # download gpg key from specified url
   if [ -f /usr/share/keyrings/${reponame}-archive-keyring.gpg ]; then
     sudo rm -f /usr/share/keyrings/${reponame}-archive-keyring.gpg
@@ -740,16 +757,25 @@ add_external_repo() { # add an external apt repo and its gpg key.
   wget -qO- "$pubkeyurl" | sudo gpg --dearmor -o /usr/share/keyrings/${reponame}-archive-keyring.gpg
 
   if [ $? != 0 ];then
-    sudo rm -f /etc/apt/sources.list.d/${reponame}.list
+    sudo rm -f /etc/apt/sources.list.d/${reponame}.sources
     sudo rm -f /usr/share/keyrings/${reponame}-archive-keyring.gpg
     error "add_external_repo: download from specified pubkeyurl failed."
   fi
 
-  # create .list file
-  echo "deb [signed-by=/usr/share/keyrings/${reponame}-archive-keyring.gpg] $repopath" | sudo tee /etc/apt/sources.list.d/${reponame}.list >/dev/null
+  # create .sources file
+  echo "Types: deb
+URIs: $uris
+Suites: $suites" | sudo tee /etc/apt/sources.list.d/${reponame}.sources >/dev/null
+  if [ ! -z "$components" ]; then
+    echo "Components: $components" | sudo tee -a /etc/apt/sources.list.d/${reponame}.sources >/dev/null
+  fi
+  for input in "${@: 6}"; do
+    echo "$input" | sudo tee -a /etc/apt/sources.list.d/${reponame}.sources >/dev/null
+  done
+  echo "Signed-By: /usr/share/keyrings/${reponame}-archive-keyring.gpg" | sudo tee -a /etc/apt/sources.list.d/${reponame}.sources >/dev/null
 
   if [ $? != 0 ];then
-    sudo rm -f /etc/apt/sources.list.d/${reponame}.list
+    sudo rm -f /etc/apt/sources.list.d/${reponame}.sources
     sudo rm -f /usr/share/keyrings/${reponame}-archive-keyring.gpg
     error "add_external_repo: failed to create ${reponame}.list file"
   fi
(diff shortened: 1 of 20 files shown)


theofficialgman:
We already have a remove repo if unused function. You should extend that to remove the key if used as well.
https://github.com/Botspot/pi-apps/blob/d20a6330924ce6523612717e7aec33b965a807fc/api#L214

api
```diff
@@ -351,6 +351,38 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
   echo "pi-apps-$(echo "$app" | md5sum | cut -c1-8 | awk '{print $1}')"
 }
 
+add_external_repo() { # add an external apt repo and its gpg key
+  repopath="$1"
+  reponame="$2"
+  pubkeyurl="$3"
```
Botspot, line 357: I recommend adding something to check that each of these values are populated, and that the URLs are reachable.

api
```diff
@@ -351,6 +351,38 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+  if [ ! -d /tmp/pi-apps-apt-keys ]; then
+    mkdir -p /tmp/pi-apps-apt-keys || error "add_external_repo: failed to create pi-apps key storage dir."
+  fi
+
+  # check if .list file already exists
+  [ -f /etc/apt/sources.list.d/${reponame}.list ] && error "add_external_repo: /etc/apt/sources.list.d/${reponame}.list already exists."
+
+  # download gpg key from specified url
+  rm -f /tmp/pi-apps-apt-keys/${reponame}-keyring.gpg
+  wget -q "$pubkeyurl" -O /tmp/pi-apps-apt-keys/${reponame}-keyring.gpg || error "add_external_repo: download from specified pubkeyurl failed." && rm -f /tmp/pi-apps-apt-keys/${reponame}-keyring.gpg
```
Botspot, line 369: I recommend testing the conditionals in this line.
If wget succeeds, the `||` will be skipped but the `&&` will run `rm`, which immediately removes the file `wget` just downloaded. Is that intended?
Conversely, if wget fails, the `||` will trigger, which runs the `error` function and exit the script. In this case, the `rm` command won't run, when I think you intended it to run.

api
```diff
@@ -351,6 +351,38 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+
+  # create .list file
+  echo "deb [signed-by=/tmp/pi-apps-apt-keys/$reponame-keyring.gpg] $repopath" | sudo tee /etc/apt/sources.list.d/${reponame}.list >/dev/null || error "add_external_repo: failed to create .list file"
```
Botspot, line 372: You seem to be treating files in /tmp as though they are persistent. They are not - every once in a while old contents will be cleared, and a reboot will clear it too.
GPG keys should be stored in a persistent, root-owned location.

api
```diff
@@ -351,6 +351,38 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+
+  # test if the repo works
+  sudo apt-get update -qq || error "add_external_repo: apt-get update failed. something is wrong with apt on your system."
```
Botspot, line 375: I don't think it is good to have apt update within this function. `install_packages` will likely be run right after this function, and it already does an apt update anyway.

api
ryanfortner: I have added a check for each of the variables, as well as a check for if `pubkeyurl` is reachable.

However I'm not sure how we could check if `repopath` is reachable... maybe test if the `Packages` file exists on the server?

api
ryanfortner: I changed the location from `/tmp` to `/opt/pi-apps-apt-keys`.

api
ryanfortner: removed.

api
ryanfortner: My intention was that if wget fails, then the incomplete download file should be removed as well as the error printed.

I have removed the && part anyways, because the function already checks if the url is reachable so it isn't likely that there will be download issues.

api
Botspot: @theofficialgman, where are system-wide GPG keys stored for things like apt repositories? Wouldn't it be better to store these keys in a standard location than a new one?

Crilum: > @theofficialgman, where are system-wide GPG keys usually stored for things like apt repositories? Wouldn't it be better to store these keys in a standard location than a new one?

It looks like `/usr/share/keyrings/`.. Don't quote me on that.. 😆

Jai-JAP: Yup thats the correct location for all apt keyrings

api
```diff
@@ -351,6 +351,43 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+  [ -z "$reponame" ] && error "add_external_packages: reponame not set"
+  [ -z "$pubkeyurl" ] && error "add_external_packages: pubkeyurl not set"
+
+  # check if links are valid
+  wget -q --spider "$pubkeyurl" || error "add_external_packages: pubkeyurl isn't a valid link"
+
+  # make pi-apps key storage dir if it doesn't exist
+  if [ ! -d /opt/pi-apps-apt-keys ]; then
+    mkdir -p /opt/pi-apps-apt-keys || error "add_external_repo: failed to create pi-apps key storage dir."
+  fi
```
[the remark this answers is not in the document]
theofficialgman: oops... I had this review pending so you didn't see my comments from earlier today

my bad

api
```diff
@@ -351,6 +351,43 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+  [ -z "$repopath" ] && error "add_external_packages: repopath not set"
+  [ -z "$reponame" ] && error "add_external_packages: reponame not set"
+  [ -z "$pubkeyurl" ] && error "add_external_packages: pubkeyurl not set"
+
+  # check if links are valid
+  wget -q --spider "$pubkeyurl" || error "add_external_packages: pubkeyurl isn't a valid link"
+
+  # make pi-apps key storage dir if it doesn't exist
+  if [ ! -d /usr/share/keyrings ]; then
+    mkdir -p /usr/share/keyrings || error "add_external_repo: failed to create pi-apps key storage dir."
```
theofficialgman, line 369: all files/folders in /usr are root owned, this would fail without sudo if it gets called

api
ryanfortner: changed the location of the keys to `/usr/share/keyrings`.

ryanfortner:
I just made a few changes. If everything looks good, I can go ahead and change the scripts for each of the apps that need the functions in this branch, so we can get it all done at once.

api
```diff
@@ -351,6 +351,55 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+  if [ ! -d /usr/share/keyrings ]; then
+    sudo mkdir -p /usr/share/keyrings || error "add_external_repo: failed to create pi-apps key storage dir."
+  fi
+
+  # check if .list file already exists
+  [ -f /etc/apt/sources.list.d/${reponame}.list ] && error "add_external_repo: /etc/apt/sources.list.d/${reponame}.list already exists."
+
+  # download gpg key from specified url
+  rm -f /usr/share/keyrings/${reponame}-keyring.gpg
+  curl -fsSL "$pubkeyurl" | sudo gpg --dearmor -o /usr/share/keyrings${reponame}-keyring.gpg
```
theofficialgman, line 377: missing `/`

api
```diff
@@ -351,6 +351,55 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+
+  if [ $? != 0 ];then
+    sudo rm -f /etc/apt/sources.list.d/${reponame}.list
+    sudo rm -f /usr/share/keyrings/${reponame}-keyring.gpg
+    error "add_external_repo: download from specified pubkeyurl failed."
+  fi
+
+  # create .list file
+  echo "deb [signed-by=/usr/share/keyrings/$reponame-keyring.gpg] $repopath" | sudo tee /etc/apt/sources.list.d/${reponame}.list >/dev/null
```
theofficialgman, line 386: I think we still shouldn't specify a specific key. apt will search the whole directory for a valid key. in the chance that a repo updates its own key (in a different filename) the key that is hardcoded here may be invalid

api
```diff
@@ -351,6 +351,55 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+  fi
+}
+
+rm_external_repo() { # remove an external apt repo and its gpg key
+  reponame="$1"
+
+  rm -f /usr/share/keyrings/${reponame}-keyring.gpg || error "rm_external_repo: removal of ${reponame}-keyring.gpg failed"
+
+  sudo rm -f /etc/apt/sources.list.d/${reponame}.list || error "rm_external_repo: removal of ${reponame}.list failed"
+}
```
theofficialgman, line 401: I still think this function should be made to be much more like "remove_repofile_if_unused" (thats an existing function)

we shouldn't always remove a repo because it may be in use in multiple apps.

ryanfortner: makes sense. I will make some changes later tomorrow, it's getting late here.

ryanfortner:
Alright @theofficialgman I changed the `rm_external_repo` function to be more like the `remore_repofile_if_unused` function. 

Just in case the repo would need to be removed no matter what, I added a `force` option where the repo will be removed regardless of what packages are installed from it. (see the top comment)

theofficialgman:
this might be overkill, but spaces are not allowed in apt repo filename or keyname. this function doesn't do anything to prevent the script writer from making that mistake or document it.

at minimum some documents should be made to specify that repopath (the second input) should never contain spaces

ryanfortner:
I added a check for spaces and updated documentation.

api
```diff
@@ -351,6 +350,94 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+  # check if links are valid
+  wget -q --spider "$pubkeyurl" || error "add_external_packages: pubkeyurl isn't a valid link"
+
+  # make pi-apps key storage dir if it doesn't exist
+  if [ ! -d /usr/share/keyrings ]; then
+    sudo mkdir -p /usr/share/keyrings || error "add_external_repo: failed to create pi-apps key storage dir."
+  fi
+
+  # check if .list file already exists
+  [ -f /etc/apt/sources.list.d/${reponame}.list ] && error "add_external_repo: /etc/apt/sources.list.d/${reponame}.list already exists."
```
Botspot, line 377: I feel like this error would be triggered if this function is somehow run twice, which is likely to happen.
Multiple apps may use the same repo, and I try to support the same app being installed multiple times without an uninstall in between.

api
```diff
@@ -351,6 +350,94 @@ app_to_pkgname() { #given an app-name, convert it to a unique, valid package-nam
...
+
+  # download gpg key from specified url
+  rm -f /usr/share/keyrings/${reponame}-keyring.gpg
+  curl -fsSL "$pubkeyurl" | sudo gpg --dearmor -o /usr/share/keyrings/${reponame}-keyring.gpg
```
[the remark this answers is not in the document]
Botspot: Also, `sudo` is used to create the ${reponame}-keyring.gpg file, but no sudo is used to remove it. I haven't tested this, but it seems like rm would fail.

api
Botspot: Probably best to first remove it.

theofficialgman:
marking this as draft since I have been doing some testing and found that some repos need the signed-by to specify the key location, and some others don't need dearmor... this is going to be complicated to do universally because of this and it might be best to have a set of guidelines and something done on a script by script basis

theofficialgman:
Deb822 format (.sources) has been supported for many years now (and all modern distros are switching to it) and it is probably best that if we were to add functions for adding and removing repos that we also do this with the Deb822 format rather than the older .list format.

https://manpages.debian.org/bookworm/apt/sources.list.5.en.html#THE_DEB_AND_DEB-SRC_TYPES:_GENERAL_FORMAT

One nice thing about deb822 is it supported external public key files as well as embedding the public key directly within the .sources file.

theofficialgman:
I have revived this PR.
However the test case in the initial comment https://github.com/Botspot/pi-apps/pull/1902#issue-1251632420 will not work as it does not have a debian suite and component defined which is required according to the debian spec for third party repos which these new functions follow.

theofficialgman:
A new test case:
```bash
add_external_repo "https://adoptium.jfrog.io/artifactory/deb" "adoptium" "bionic" "main" "https://adoptium.jfrog.io/artifactory/api/security/keypair/default-gpg-key/public" || exit 1

rm_external_repo "adoptium" || exit 1 # this will remove the repo if nothing is used in it
rm_external_repo "adoptium" "force" || exit 1 # this will remove the repo even if software is installed from it
```

theofficialgman:
I have switched all applications that can be switched to use these generic functions.

The other applications (vivaldi and brave) cannot use these generic functions as they use their own custom non-standard key and repofile names and update them themselves as part of other debs.

theofficialgman:
A new test case (based on the above edits):
```bash
add_external_repo "adoptium" "https://adoptium.jfrog.io/artifactory/api/security/keypair/default-gpg-key/public" "https://adoptium.jfrog.io/artifactory/deb" "bionic" "main" || exit 1

rm_external_repo "adoptium" || exit 1 # this will remove the repo if nothing is used in it
rm_external_repo "adoptium" "force" || exit 1 # this will remove the repo even if software is installed from it
```

the original test case now becomes:
```bash
add_external_repo "box64" "https://ryanfortner.github.io/box64-debs/KEY.gpg" "https://ryanfortner.github.io/box64-debs/debian" "./" || exit 1

rm_external_repo "box64" || exit 1 # this will remove the repo if nothing is used in it
rm_external_repo "box64" "force" || exit 1 # this will remove the repo even if software is installed from it
```

theofficialgman:
@ryanfortner thanks for your original contribution! I will merge the current iteration of this PR. Sorry it took so long to get back around to this.

Between when this PR was opened and now pi-apps moved documentation from this repo to our website so I will rewrite that information there soon.

Issue #1894, opened 2022-05-27
Xfburn


Rak1ta:
### What is the name of the app?

Xfburn

### About the app

_Xfburn is a tool to help burning CDs and DVDs. It fits well in the Xfce Desktop Environment but can be used anywhere._

https://docs.xfce.org/apps/xfburn/start
I don't have an external drive at the moment, but there were no complaints on the rpi forum. This is package-app.
[Xfburn.zip](https://github.com/Botspot/pi-apps/files/8785000/Xfburn.zip)

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.
- [ ] I have confirmed that this app can run on Raspberry Pi.

ryanfortner:
I actually just used this the other day to burn an ISO to a cd drive, it worked perfectly. 

I haven’t tried it using rpios’ default desktop environment though. Are you sure xfburn can be installed without the entire xfce desktop suite along with it?

Rak1ta:
@ryanfortner are you talking about dependencies?
![Безымянный](https://user-images.githubusercontent.com/88277343/171050613-a8431df6-ccd4-4a1c-a5b8-e77e7ca0a84e.png)

Crilum:
```
$ sudo apt install xfburn
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  libburn4 libexo-2-0 libexo-common libisofs6 libjte2 libxfce4ui-2-0
  libxfce4ui-common
Suggested packages:
  devhelp
The following NEW packages will be installed:
  libburn4 libexo-2-0 libexo-common libisofs6 libjte2 libxfce4ui-2-0
  libxfce4ui-common xfburn
0 upgraded, 8 newly installed, 0 to remove and 5 not upgraded.
Need to get 1,947 kB of archives.
After this operation, 6,903 kB of additional disk space will be used.
Do you want to continue? [Y/n]
```

On RPiOS 11

ryanfortner:
Thanks for confirming @Crilum and @Rak1ta , looks like it's safe to install on non-XFCE desktops. I have opened a pr, just waiting for botspot's approval. https://github.com/Botspot/pi-apps/pull/1916

PR #1916, opened 2022-05-30
add xfburn package

ryanfortner:
Closes https://github.com/Botspot/pi-apps/issues/1894


theofficialgman:
@Botspot would you might adding additional "rules" to the rubric specifically regarding package apps. https://github.com/Botspot/pi-apps/issues/185

I have said before that I don't like package apps and don't feel like they really fit pi-apps but that isn't the matter of discussion here (I can just disable them with the settings toggle which is enough for me).

What I would like clarified is what type of package apps should be eligible/included in pi-apps. I feel like we are reaching the point where package apps are becoming a significant portion of the app list and rules for their inclusion should be listed.

Botspot:
> @Botspot would you might adding additional "rules" to the rubric specifically regarding package apps. #185
> 
> I have said before that I don't like package apps and don't feel like they really fit pi-apps but that isn't the matter of discussion here (I can just disable them with the settings toggle which is enough for me).
> 
> What I would like clarified is what type of package apps should be eligible/included in pi-apps. I feel like we are reaching the point where package apps are becoming a significant portion of the app list and rules for their inclusion should be listed.

A bit of background:
I never intended for Pi-Apps to include apt packages. But I also never intended it to be a primary installation method for some users.
What ended up happening was that apps available on apt began having a *disadvantage*.
If a GUI user wants to install an image editor from a curated list of apps (which is what Pi-Apps has become), it would be a shame to only include non-apt apps, while excluding potentially better options just because they were from apt.

There is definitely a potential for this package-app feature to be taken too far. @theofficialgman, if you have any particular package-apps in mind that don't seem to fit, by all means let me know.

Crilum:
The purpose of Pi-Apps is definitely _not_ an Apt GUI..

theofficialgman:
> What ended up happening was that apps available on apt began having a _disadvantage_. If a GUI user wants to install an image editor from a curated list of apps (which is what Pi-Apps has become), it would be a shame to only include non-apt apps, while excluding potentially better options just because they were from apt.

I like this explanation, it makes sense. Given this, do you think it would be a good idea for our "rules" then to say that package apps should only be included if there is already a similar alternative for that package already included in pi-apps?


under this "rule" the xfburn package app would be fine to include because there is similar non-package app software already pi-apps (USBImager)

Crilum:
> there is already a similar alternative for that package already included in pi-apps?

is or isn't?

theofficialgman:
> is or isn't?

is. I think we should only include a package app IF there is already an install(32/64) type software for that in pi-apps. basically, no new random software from package apps that has no pi-apps unique (install32/64) alternative

Botspot:
> I like this explanation, it makes sense. Given this, do you think it would be a good idea for our "rules" then to say that package apps should only be included if there is already a similar alternative for that package already included in pi-apps?

This sounds reasonable to me, though it might make for some silly situations down the road.
For example, a potential app-submitter may try to add a script app for the sole purpose of adding a package-app afterwards.
Or, the app-submitter will make a script-app to install a slightly newer version of an already available apt package, "just because" it's allowed. But I guess this one has already happened a few times. 😂

PR #1928, opened 2022-05-31
Speedtest cli deb

theofficialgman:
I'm cleaning out my home directory and would prefer less files in it.

lets use the deb since its available. the resulting deb is the same for all supported distros, so I've just chosen bionic

### DIFF
--- a/apps/SpeedTest-CLI/uninstall
+++ b/apps/SpeedTest-CLI/uninstall
@@ -1,4 +1,6 @@
 #!/bin/bash
-rm -f ~/.local/share/applications/speedtest.desktop || error "failed to remove menu shortcut!"
-rm -rf ~/speedtest
-sudo rm -f /usr/local/bin/speedtest
+
+purge_packages || exit 1
+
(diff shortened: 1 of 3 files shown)


apps/SpeedTest-CLI/uninstall
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
-rm -f ~/.local/share/applications/speedtest.desktop || error "failed to remove menu shortcut!"
-rm -rf ~/speedtest
-sudo rm -f /usr/local/bin/speedtest
+
+purge_packages || exit 1
+
+sudo rm -rf ~/.local/share/applications/speedtest.desktop /usr/share/applications/speedtest.desktop || error "failed to remove menu shortcut!"
+sudo rm -rf /usr/local/bin/speedtest ~/speedtest
```
Itai-Nelken, line 6: I think it's a good idea to add a comment explaining why those files are being removed (cleaning up after the old script).

theofficialgman: done

Issue #1842, opened 2022-05-19
Blank list - No Categories


DaveDischord:
### Confirmations

- [X] I searched for similar bug reports in https://github.com/Botspot/pi-apps/issues and found none were relevant.
- [X] I tried updating Pi Apps and the issue is still not fixed. [How to update Pi Apps?](https://github.com/Botspot/pi-apps#to-update-pi-apps)

### What happened?

App does not display list of categories after launch.

### Description

It started after an Pi-Apps update.  I went ahead and uninstalled Pi-apps and removed the ~/pi-apps folder before doing a fresh install with the same results.  I've even changed themes between light and dark and went to Settings -> Categories -> All and Saved with no change
![Screen Shot 05-19-22 at 09 19 AM](https://user-images.githubusercontent.com/14235646/169318890-fb20a0c2-1168-49e7-881b-51c087a55656.PNG)
.

### What is your operating system?

TwisterOS on Pi4

DaveDischord:
The only thing out of the ordinary from the looks of it is the **#00AA00** on Line 28
[output.txt](https://github.com/Botspot/pi-apps/files/8730377/output.txt)
2

Crilum:
This has happened to me a couple of times too..

Crilum:
I did, when I had the issue, it looked like `preload` issues..

DaveDischord:
Just looks like its hanging on 'Reading list file...' unless that's normal.
![Screen Shot 05-19-22 at 12 48 PM](https://user-images.githubusercontent.com/14235646/169365055-8e7ac51d-b51f-472b-be96-d022437ad764.PNG)

Botspot:
> Just looks like its hanging on 'Reading list file...' unless that's normal. ![Screen Shot 05-19-22 at 12 48 PM](https://user-images.githubusercontent.com/14235646/169365055-8e7ac51d-b51f-472b-be96-d022437ad764.PNG)

Do you have a discord account? I'd like to test some things myself on your system through AnyDesk, and we need a private place to share credentials.

DaveDischord:
Yes, but I can't seem to be able to install AnyDesk. Using `sudo pi-apps/apps/AnyDesk/install-32` returns `pi-apps/apps/AnyDesk/install-32: line 5: install_packages: command not found`

Botspot:
[...]

PR #1931, opened 2022-06-01
regenerate yad list if has one form of corruption

theofficialgman:
we have seen a couple of cases (as have I myself) where a yad list gets corrupted (probably by us writing to it twice at once).

this adds a simple check for a yad list to verify it is a multiple of 5. if it is not, then regenerate it.

### DIFF
--- a/preload
+++ b/preload
@@ -111,6 +111,18 @@ fi
 if [ ! -f "$listfile" ] || [ -z "$(cat "$listfile")" ];then
   echo "list file for $prefix does not exist." 1>&2
   reloadlist=1
+else
+  list_lenght=$(wc -l "$listfile" | awk '{print $1}')
+fi
+
+if [ -n "$list_lenght" ] && [ "$format" == yad ] && (( $list_lenght % 5 )) ; then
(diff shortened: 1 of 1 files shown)


theofficialgman:
for example, somehow this users list is corrupted here https://github.com/Botspot/pi-apps/issues/1842#issuecomment-1131817099 . this PR will re-gen that list which should hopefully fix it before it is ever used

preload
```diff
@@ -108,6 +108,18 @@ else
...
+fi
+
+if [ -n "$list_lenght" ] && [ "$format" == yad ] && (( $list_lenght % 5 )) ; then
+  echo "yad lists are a multiple of 5, $prefix does not have a list a multiple of 5 so it may be corrupted and needs regenerating." 1>&2
+  reloadlist=1
+fi
+
 if [ ! -f "$listfile" ] || [ -z "$(cat "$listfile")" ];then
   echo "list file for $prefix does not exist." 1>&2
   reloadlist=1
```
Botspot, line 125: You probably intended to remove these lines. Otherwise, they are duplicates.

Issue #1825, opened 2022-05-16
GUI update suggestions


Rak1ta:
1. Is it possible to put the menu item "back" additionally at the end of the list of programs? In order not to scroll back to return.
2. When you try to move the program description window or expand the window to view the text - everything disappears. Is it possible to make it so that it is removed only at the request of the user? I open the program script - I close Geany - the window with the description disappears.
3. Black popup borders / menu item (zoom PWA) is overridden

![Безымянный](https://user-images.githubusercontent.com/88277343/168654769-3b29bcb5-aa4d-49bb-a685-889abd3f5599.png)
![Безымянный2](https://user-images.githubusercontent.com/88277343/168644563-e8ca418f-b954-45a4-bd9c-e57544a8f7bf.png)
![Безымянный4](https://user-images.githubusercontent.com/88277343/168645500-87219df7-0db5-4303-8c14-bef5766253df.png)

Crilum:
@Botspot @theofficialgman Can the columns be made a little narrower? (Or the whole window a little wider, it seems the minimum width with no scrollbar is about 260px.) I don't see any reason for the x dimension scrollbar.. And it just causes problems as seen here.

Crilum:
@Rak1ta I don't have the black tooltip corners (as seen below), but then again, I don't have the default RPiOS PiXflat theme.
![2022-05-16-115112_1024x768_scrot](https://user-images.githubusercontent.com/91354257/168654337-7d01af68-a388-4a42-a7d2-95489c9b3b36.png)

I don't think this is a problem Pi-Apps can fix, this is probably a problem with GTK, or YAD.

theofficialgman:
my right hand side scrollbar display ON TOP OF the list, meaning it doesn't take up any list width.
this is not the case here, it looks like the users scrollbar on the right is taking up horizontal space. the Persepolis Download Manager is "just a bit" too wide to fit in that case and the scrollbar gets added.

Rak1ta:
> the GUI is not intended to be moved since its position can not be obtained
But the external window "Updates available" works as I described. :-(

theofficialgman:
> > the GUI is not intended to be moved since its position can not be obtained
> 
> But the external window "Updates available" works as I described. :-(

yes, this window does not close automatically on losing focus. the new GUI would not function as intended if the window didn't close automatically, users would click on an app, then click on another app, and the previous apps details page would still be visible

Botspot:
[...]

PR #1932, opened 2022-06-01
preload bugfixes and optimizations

theofficialgman:
if the lists timestamps do not match, the list generation is done at the time the user clicks on the category. previously, each line of the list was directly echoed to the pipe, meaning if the user clicked the back button, they could send data at the same time that other data was still loading in.

change it so that all data is generated and then read the list file into the pipe at the very end.

### DIFF
--- a/api
+++ b/api
@@ -1766,11 +1766,11 @@ send_error_report() { #non-interactively send a Pi-Apps error log file to the Bo
 runonce() { #run command only if it's never been run before. Useful for one-time migration or setting changes.
   #Runs a script in the form of stdin
   
-  script="$(cat /dev/stdin)"
+  script="$(< /dev/stdin)"
   
   runonce_hash="$(sha1sum <<<"$script" | awk '{print $1}')"
   
-  if grep -qxF "$runonce_hash" "${DIRECTORY}/data/runonce_hashes" ;then
+  if [ -s "${DIRECTORY}/data/runonce_hashes" ] && while read line; do [[ $line == "$runonce_hash" ]] && break; done < "${DIRECTORY}/data/runonce_hashes"; then
     #hash found
     #echo "runonce: '$script' already run before. Skipping."
     true
--- a/api
+++ b/api
@@ -1765,10 +1765,8 @@ send_error_report() { #non-interactively send a Pi-Apps error log file to the Bo
 #miscellaneous low-level functions below
 runonce() { #run command only if it's never been run before. Useful for one-time migration or setting changes.
   #Runs a script in the form of stdin
-  
-  script="$(< /dev/stdin)"
-  
-  runonce_hash="$(sha1sum <<<"$script" | awk '{print $1}')"
+    
+  runonce_hash="$(sha1sum <<<"$(< /dev/stdin)" | awk '{print $1}')"
   
   if [ -s "${DIRECTORY}/data/runonce_hashes" ] && while read line; do [[ $line == "$runonce_hash" ]] && break; done < "${DIRECTORY}/data/runonce_hashes"; then
     #hash found
@@ -1776,13 +1774,13 @@ runonce() { #run command only if it's never been run before. Useful for one-time
     true
   else
     #run the script.
-    bash <(echo "$script")
+    bash <(echo "$(< /dev/stdin)")
     #if it succeeds, add the hash to the list to never run it again
     if [ $? == 0 ];then
       echo "$runonce_hash" >> "${DIRECTORY}/data/runonce_hashes"
-      echo "runonce(): '$script' succeeded. Added to list."
+      echo "runonce(): '$(< /dev/stdin)' succeeded. Added to list."
     else
-      echo "runonce(): '$script' failed. Not adding hash to list."
+      echo "runonce(): '$(< /dev/stdin)' failed. Not adding hash to list."
     fi
     
   fi
--- a/api
+++ b/api
@@ -777,7 +777,7 @@ read_category_files() { #Generates a combined categories-list from several sourc
   (cat "${DIRECTORY}/data/category-overrides" "${DIRECTORY}/etc/categories" ; echo ; list_apps local | sed 's/$/|/g') | awk -F'|' '!seen[$1]++' | grep .
 }
 
-app_prefix_category() { #lists all apps in a virtual filesystem based on categories file. Format: "category/app"
+app_prefix_category() { #lists all apps in a category with formast "category/app", or if $1 is left blank, then list the full structure of all categories
   
   #Subtract a type of app if enabled
   local show_apps_setting="$(cat "${DIRECTORY}/data/settings/Show apps")"
--- a/api
+++ b/api
@@ -1991,6 +1991,7 @@ $(head -n1 "${DIRECTORY}/apps/$app/description")"
     
   fi
 }
+#end of miscellaneous low-level functions
 
 #command interceptors - functions that enhance a command
 enable_module() { #Permanent equivalent to modprobe, this will load the module on every future startup
--- a/api
+++ b/api
@@ -777,37 +777,6 @@ read_category_files() { #Generates a combined categories-list from several sourc
   (cat "${DIRECTORY}/data/category-overrides" "${DIRECTORY}/etc/categories" ; echo ; list_apps local | sed 's/$/|/g') | awk -F'|' '!seen[$1]++' | grep .
 }
 
-app_categories() { #lists all apps in a virtual filesystem based on categories file. Format: "category/app"
-  
-  #Subtract a type of app if enabled
-  local show_apps_setting="$(cat "${DIRECTORY}/data/settings/Show apps")"
-  local filter=()
-  if [ "$show_apps_setting" == 'standard' ];then
-    #if only showing standard apps, hide package apps
-    filter=(list_subtract "$(list_apps package | sed 's+^+.*/+g')")
-  elif [ "$show_apps_setting" == 'packages' ];then
-    #if only showing package apps, hide standard apps
-    filter=(list_subtract "$(list_apps standard | sed 's+^+.*/+g')")
-  else
-    #is the setting is "all" or missing, don't filter.
-    filter=(cat)
-  fi
-  
-  #show normal categories
-  read_category_files | grep . | awk -F'|' '{print $2"/"$1}' | sed 's+^/++g' | "${filter[@]}"
-  
-  #show special "Installed" category - don't filter it
-  list_apps installed | sed 's+^+Installed/+g'
-  
-  #show special "Packages" category
-  if [ "$show_apps_setting" != standard ];then
-    list_apps package | list_subtract "$(list_apps hidden)" | sed 's+^+Packages/+g'
-  fi
-  
-  #show special "All Apps" category
-  list_apps cpu_installable | list_subtract "$(list_apps hidden)" | sed 's+^+All Apps/+g' | "${filter[@]}"
-}
-
 app_prefix_category() { #lists all apps in a virtual filesystem based on categories file. Format: "category/app"
   
   #Subtract a type of app if enabled
@@ -832,6 +801,20 @@ app_prefix_category() { #lists all apps in a virtual filesystem based on categor
     list_apps package | list_subtract "$(list_apps hidden)" | sed 's+^+Packages/+g'
   #show special "All Apps" category
   elif [ "$1" == "All Apps" ]; then
+    list_apps cpu_installable | list_subtract "$(list_apps hidden)" | sed 's+^+All Apps/+g' | "${filter[@]}"
+  # show all categories
+  elif [ "$1" == "" ]; then
+    #show normal categories
+    read_category_files | grep . | awk -F'|' '{print $2"/"$1}' | sed 's+^/++g' | "${filter[@]}"
+    
+    #show special "Installed" category - don't filter it
+    list_apps installed | sed 's+^+Installed/+g'
+    
+    #show special "Packages" category
+    if [ "$show_apps_setting" != standard ];then
+      list_apps package | list_subtract "$(list_apps hidden)" | sed 's+^+Packages/+g'
+    fi
+    #show special "All Apps" category
     list_apps cpu_installable | list_subtract "$(list_apps hidden)" | sed 's+^+All Apps/+g' | "${filter[@]}"
   #show normal categories
   else
--- a/api
+++ b/api
@@ -803,7 +803,7 @@ app_prefix_category() { #lists all apps in a category with formast "category/app
   elif [ "$1" == "All Apps" ]; then
     list_apps cpu_installable | list_subtract "$(list_apps hidden)" | sed 's+^+All Apps/+g' | "${filter[@]}"
   # show all categories
-  elif [ "$1" == "" ]; then
+  elif [ -z "$1" ];; then
     #show normal categories
     read_category_files | grep . | awk -F'|' '{print $2"/"$1}' | sed 's+^/++g' | "${filter[@]}"
     
--- a/api
+++ b/api
@@ -767,14 +767,14 @@ read_category_files() { #Generates a combined categories-list from several sourc
   #remove app category if app folder not found
   local IFS=$'\n'
   local app
-  for app in $(cat "${DIRECTORY}/data/category-overrides"); do 
+  for app in $(cat "${DIRECTORY}/data/category-overrides" 2>/dev/null); do
     if ! [ -d "${DIRECTORY}/apps/$(echo "$app" | sed 's/|.*//')" ] &>/dev/null; then 
       sed -i "/$app/d"  "${DIRECTORY}/data/category-overrides"
     fi
   done
   
   #list the user-overrides file              and the global categories file          |-----and all apps------------|     filter out duplicates   no '\n\n'
-  (cat "${DIRECTORY}/data/category-overrides" "${DIRECTORY}/etc/categories" ; echo ; list_apps local | sed 's/$/|/g') | awk -F'|' '!seen[$1]++' | grep .
+  (cat "${DIRECTORY}/data/category-overrides" "${DIRECTORY}/etc/categories" 2>/dev/null ; echo ; list_apps local | sed 's/$/|/g') | awk -F'|' '!seen[$1]++' | grep .
 }
 
 app_prefix_category() { #lists all apps in a category with format "category/app", or if $1 is left blank, then list the full structure of all categories
--- a/api
+++ b/api
@@ -1098,13 +1098,8 @@ refresh_pkgapp_status() { #for the specified package-app, if dpkg thinks it's in
       bitly_link "$app" uninstall &
     fi
   else
-    #this app is trying to install a package that's not on the repository. Hide the app.
-    #only mark as hidden if its not already marked as hidden, this prevents changing the timestamp of the file, causing the preload script to re-load
-    #a common issue on All-Apps
-    if ! grep -Fxq "$app|hidden" "${DIRECTORY}/data/category-overrides" ;then
-      echo "Marking $app as hidden"
-      "${DIRECTORY}/etc/categoryedit" "$app" 'hidden' >/dev/null
-    fi
+    echo "Marking $app as hidden"
+    "${DIRECTORY}/etc/categoryedit" "$app" 'hidden' >/dev/null
   fi
 }
 
--- a/api
+++ b/api
@@ -777,7 +777,7 @@ read_category_files() { #Generates a combined categories-list from several sourc
   (cat "${DIRECTORY}/data/category-overrides" "${DIRECTORY}/etc/categories" ; echo ; list_apps local | sed 's/$/|/g') | awk -F'|' '!seen[$1]++' | grep .
 }
 
-app_prefix_category() { #lists all apps in a category with formast "category/app", or if $1 is left blank, then list the full structure of all categories
+app_prefix_category() { #lists all apps in a category with format "category/app", or if $1 is left blank, then list the full structure of all categories
   
   #Subtract a type of app if enabled
   local show_apps_setting="$(cat "${DIRECTORY}/data/settings/Show apps")"
@@ -1081,6 +1081,7 @@ refresh_pkgapp_status() { #for the specified package-app, if dpkg thinks it's in
       bitly_link "$app" uninstall &
     fi
   else
+    #this app is trying to install a package that's not on the repository. Hide the app.
     echo "Marking $app as hidden"
     "${DIRECTORY}/etc/categoryedit" "$app" 'hidden' >/dev/null
   fi
--- a/api
+++ b/api
@@ -1090,13 +1090,12 @@ refresh_pkgapp_status() { #for the specified package-app, if dpkg thinks it's in
 refresh_all_pkgapp_status() { #for every package-app, if dpkg thinks it's installed, then mark it as installed.
   #repeat for every package-type app
   local IFS=$'\n'
+  local PIDS=()
   for app in $(list_apps package) ;do
-    refresh_pkgapp_status "$app" &  declare PID_$(TMP=${app//-/_};echo ${TMP// /_})=$!
+    refresh_pkgapp_status "$app" &
+    PIDS+=("$!")
   done
-  for app in $(list_apps package) ;do
-    varname=PID_$(TMP=${app//-/_};echo ${TMP// /_})
-    wait $(echo ${!varname})
-  done
+  wait "${PIDS[@]}" #wait for all subprocesses to complete
 }
 
 refresh_app_list() { #Force-regenerate the app list
--- a/api
+++ b/api
@@ -1070,8 +1070,12 @@ refresh_pkgapp_status() { #for the specified package-app, if dpkg thinks it's in
     fi
   else
     #this app is trying to install a package that's not on the repository. Hide the app.
-    echo "Marking $app as hidden"
-    "${DIRECTORY}/etc/categoryedit" "$app" 'hidden' >/dev/null
+    #only mark as hidden if its not already marked as hidden, this prevents changing the timestamp of the file, causing the preload script to re-load
+    #a common issue on All-Apps
+    if ! grep -Fxq "$app|hidden" "${DIRECTORY}/data/category-overrides" ;then
+      echo "Marking $app as hidden"
+      "${DIRECTORY}/etc/categoryedit" "$app" 'hidden' >/dev/null
+    fi
   fi
 }
 
--- a/api
+++ b/api
@@ -1091,7 +1091,11 @@ refresh_all_pkgapp_status() { #for every package-app, if dpkg thinks it's instal
   #repeat for every package-type app
   local IFS=$'\n'
   for app in $(list_apps package) ;do
-    refresh_pkgapp_status "$app"
+    refresh_pkgapp_status "$app" &  declare PID_$(TMP=${app//-/_};echo ${TMP// /_})=$!
+  done
+  for app in $(list_apps package) ;do
+    varname=PID_$(TMP=${app//-/_};echo ${TMP// /_})
+    wait $(echo ${!varname})
   done
 }
 
--- a/api
+++ b/api
@@ -1731,8 +1731,10 @@ send_error_report() { #non-interactively send a Pi-Apps error log file to the Bo
 #miscellaneous low-level functions below
 runonce() { #run command only if it's never been run before. Useful for one-time migration or setting changes.
   #Runs a script in the form of stdin
-    
-  runonce_hash="$(sha1sum <<<"$(< /dev/stdin)" | awk '{print $1}')"
+  
+  script="$(< /dev/stdin)"
+  
+  runonce_hash="$(sha1sum <<<"$script" | awk '{print $1}')"
   
   if [ -s "${DIRECTORY}/data/runonce_hashes" ] && while read line; do [[ $line == "$runonce_hash" ]] && break; done < "${DIRECTORY}/data/runonce_hashes"; then
     #hash found
@@ -1740,13 +1742,13 @@ runonce() { #run command only if it's never been run before. Useful for one-time
     true
   else
     #run the script.
-    bash <(echo "$(< /dev/stdin)")
+    bash <(echo "$script")
     #if it succeeds, add the hash to the list to never run it again
     if [ $? == 0 ];then
       echo "$runonce_hash" >> "${DIRECTORY}/data/runonce_hashes"
-      echo "runonce(): '$(< /dev/stdin)' succeeded. Added to list."
+      echo "runonce(): '$script' succeeded. Added to list."
     else
-      echo "runonce(): '$(< /dev/stdin)' failed. Not adding hash to list."
+      echo "runonce(): '$script' failed. Not adding hash to list."
     fi
     
   fi
--- a/api
+++ b/api
@@ -57,40 +57,6 @@ generate_logo() { #display colorized Pi-Apps logo in terminal
 \e[0m\e[0m"
   fi
 }
-
-generate_splashscreen() { #Place 10 random app icons on the loading screen for the next time Pi-Apps launches
-  
-  #get 10 random apps. Some are exempted because they don't display well.
-  local apps="$(list_apps local | grep -v 'Raspi2png\|template\|YouTubuddy' | shuf)"
-  
-  local app1="$(sed -n 1p <<<"$apps")"
-  local app2="$(sed -n 2p <<<"$apps")"
-  local app3="$(sed -n 3p <<<"$apps")"
-  local app4="$(sed -n 4p <<<"$apps")"
-  local app5="$(sed -n 5p <<<"$apps")"
-  local app6="$(sed -n 6p <<<"$apps")"
-  local app7="$(sed -n 7p <<<"$apps")"
-  local app8="$(sed -n 8p <<<"$apps")"
-  local app9="$(sed -n 9p <<<"$apps")"
-  local app10="$(sed -n 10p <<<"$apps")"
-  
-  #display the chosen apps
-  #echo "$apps" | head -n 10
-  
-  sed "s_IMAGE0_$(base64 "${DIRECTORY}/apps/${app1}/icon-64.png" -w 0)_g ; \
-  s_IMAGE1_$(base64 "${DIRECTORY}/apps/${app2}/icon-64.png" -w 0)_g ; \
-  s_IMAGE2_$(base64 "${DIRECTORY}/apps/${app3}/icon-64.png" -w 0)_g ; \
-  s_IMAGE3_$(base64 "${DIRECTORY}/apps/${app4}/icon-64.png" -w 0)_g ; \
-  s_IMAGE4_$(base64 "${DIRECTORY}/apps/${app5}/icon-64.png" -w 0)_g ; \
-  s_IMAGE5_$(base64 "${DIRECTORY}/apps/${app6}/icon-64.png" -w 0)_g ; \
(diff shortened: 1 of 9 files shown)


theofficialgman:
https://github.com/Botspot/pi-apps/issues/1825#issuecomment-1134727587
here is another example of what happens because of the previous format

theofficialgman:
and another one: https://github.com/Botspot/pi-apps/issues/1825#issuecomment-1137947204

theofficialgman:
the whole point of the `preload-daemon` is to keep the lists updated so they DONT have to be generated when the user clicks a button. if you find that the lists are out of date frequently (meaning you get this `Timestamps don't match. Generating list for...`, then decrease the 30 second sleep

theofficialgman:
So far these changes have resulted in a first time start of pi-apps to fully generated `~/pi-apps/data/preload/` lists reduction of 15 seconds, down to 8 seconds on my system.

api
```diff
@@ -755,13 +755,11 @@ list_apps() { # $1 can be: installed, uninstalled, corrupted, cpu_installable, h
 }
 
 list_intersect() { #Outputs only the apps that appear in both stdin and in $1
-  #                      change \n to \|     |   remove last "\|"
-  grep -x "$(echo "$1" | sed -z 's/\n/\\|/g' | sed -z 's/\\|$/\n/g')"
+  comm -12 - <(echo "$1" | sort)
```
Botspot, line 758: This is fantastic! @theofficialgman how did you come across the `comm` command?

preload
```diff
@@ -14,7 +14,6 @@ ${DIRECTORY}/data/settings
 ${DIRECTORY}/data/status
 ${DIRECTORY}/etc
 ${DIRECTORY}/icons/categories
-${DIRECTORY}/apps
```
Botspot, line 17: Why did you remove ${DIRECTORY}/apps from `mktimestamps`?
It would seem that this change would regenerate the list if a new app was added.

theofficialgman: Look up, it was in the list rwice

api
theofficialgman: its one of the coreutils packages. coreutils are the gnu untils installed on every system

documentation for all coreutils packages are here: https://www.gnu.org/software/coreutils/manual/

you can also just use the filelist and look at the binaries included in the deb https://packages.debian.org/bullseye/amd64/coreutils/filelist

Crilum: > its one of the coreutils packages. coreutils are the gnu untils installed on every system

We should probably still add it as a dependency of Pi-Apps..

theofficialgman:
https://github.com/Botspot/pi-apps/blob/e26db0253f3e0d688b8d3c6b1ab9e20fff389bd8/gui#L503

@Botspot so I noticed the runonce scripts in the GUI run asynchronously to the rest of the GUI (but synchronously in order) and on first run (if you have the dependencies already installed) it take like 10 seconds to get through all of them.

this means that often times, the preload has already finished most/all of the first time LIST generation before the runonce which calls refresh_all_pkgapp_status gets run, invalidating the timestamps due to refresh_all_pkgapp_status marking some already installed packages as installed

theofficialgman:
also I personally think that the dependencies runonce should block the GUI from loading (aka, put it outside the &), if the dependencies aren't there, we have no buisness showing the GUI yet

Botspot:
> also I personally think that the dependencies runonce should block the GUI from loading (aka, put it outside the &), if the dependencies aren't there, we have no business showing the GUI yet

The downside to making all runonce entries complete before the gui appears is the recurrent slowdown caused by it. Runonce takes time to hash the commands and compare that hash to the runonce-hashes file, so making the gui wait until this is finished will increase launch time.

theofficialgman:
> The downside to making all runonce entries complete before the gui appears is the recurrent slowdown caused by it.

I didn't say all, I just want the dependencies runonce to happen before the GUI appears and block it. we shouldn't let users run the GUI without having the script dependencies because they could ignore the prompt (or cancel out) of the password input and never actually install them

etc/preload-daemon
```diff
@@ -29,10 +29,10 @@ source "${DIRECTORY}/preload" "source"
 #variable 2 is 'once', if you only want this to run once and exit.
 
 #generate list of folders to preload, including special folders
-folders="$(read_category_files | awk -F '|' '{print $2}' | sort | uniq | grep .)
+folders="All Apps
+$(read_category_files | awk -F '|' '{print $2}' | sort | uniq | grep .)
 Installed
-Packages
-All Apps"
+Packages"
```
Botspot, line 35: I'd be interested to know your reasoning behind this change. This will preload `All Apps` first, to the detriment of all the others. It would seem logical to preload the smaller folders first.

theofficialgman: The logic was that a user may be more likely to click on All Apps than any other single category.

In my mind the most likely categories would go All Apps -> Installed -> everything else

You can change it back though, it doesn't really matter

theofficialgman:
@Botspot you will have to tell me if you are comfortable with my previous commit. I will explain it here and why its so much faster

```
refresh_all_pkgapp_status() { #for every package-app, if dpkg thinks it's installed, then mark it as installed.
  #repeat for every package-type app
  local IFS=$'\n'
  for app in $(list_apps package) ;do
    refresh_pkgapp_status "$app" &  declare PID_$(TMP=${app//-/_};echo ${TMP// /_})=$!
  done
  for app in $(list_apps package) ;do
    varname=PID_$(TMP=${app//-/_};echo ${TMP// /_})
    wait $(echo ${!varname})
  done
}
```

the basics are that we are sending the refresh_pkgapp_status to the background so that multiple apps can have their status refreshed at once. this is perfectly fine for all apps that simply read the local lists in /var/apt/list and set their own status.

but, we don't want to exit refresh_all_pkgapp_status before all the apps status are updated, so we obtain the PID value of each background process, store it in a variable based on the name of the app (with spaces and - converted to _) and wait for each one of them to exit before exiting refresh_all_pkgapp_status.

the only edge case has to do with the package apps that need to be hidden. they call the `etc/categoryedit` script and multiple instances of this can not be run at once, otherwise you end up overwriting the file overrides file with data that is missing some of the apps.

etc/categoryedit
```diff
@@ -9,6 +9,12 @@ function error {
   exit 1
 }
 
+#Only one instance of this script should run at a time. wait if another process other than this one is running
+while [[ $(pgrep categoryedit | sort |  head -n 1) != $$ ]]; do
+  # waiting for another instance of categoryedit to finish
+  sleep 0.02
+done
+
```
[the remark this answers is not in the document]
theofficialgman: hmm.. I'll have to think about what should be done in this case. (will get back to it in a few hours)

api
```diff
@@ -1091,7 +1091,11 @@ refresh_all_pkgapp_status() { #for every package-app, if dpkg thinks it's instal
   #repeat for every package-type app
   local IFS=$'\n'
   for app in $(list_apps package) ;do
-    refresh_pkgapp_status "$app"
+    refresh_pkgapp_status "$app" &  declare PID_$(TMP=${app//-/_};echo ${TMP// /_})=$!
+  done
+  for app in $(list_apps package) ;do
+    varname=PID_$(TMP=${app//-/_};echo ${TMP// /_})
+    wait $(echo ${!varname})
```
Botspot, line 1098: This seems overdesigned. On my system it took 4.2 seconds.
The single-threaded version took 2.6 seconds on my system.
I found a more efficient way to do multithreading:
```bash
refresh_all_pkgapp_status() { #for every package-app, if dpkg thinks it's installed, then mark it as installed.
  #repeat for every package-type app
  local IFS=$'\n'
  local PIDS
  for app in $(list_apps package) ;do
    refresh_pkgapp_status "$app" &
    PIDS+="$!"$'\n'
  done
  for PID in $PIDS ;do
    wait $PID
  done
}
```

Botspot: Further optimized:
```bash
refresh_all_pkgapp_status() { #for every package-app, if dpkg thinks it's installed, then mark it as installed.
  #repeat for every package-type app
  local IFS=$'\n'
  local PIDS=()
  for app in $(list_apps package) ;do
    refresh_pkgapp_status "$app" &
    PIDS+=("$!")
  done
  wait "${PIDS[@]}" #wait for all subprocesses to complete
}
```

Botspot: I've gone ahead and improved the branch with my changes.

theofficialgman: > This seems overdesigned. On my system it took 4.2 seconds. The single-threaded version took 2.6 seconds on my system. I found a more efficient way to do multithreading:

did you report these numbers backward? you said the old method (where each refresh had to wait on another) took 2.6 seconds, and my new method took 4.2 seconds

theofficialgman: regardless, you method seems to work fine and is definitely a cleaner way of writing it

etc/categoryedit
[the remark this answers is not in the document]
theofficialgman: could we break out the GUI part of this file into its own file? that way the Category editor could just source `etc/categoryedit` to obtain the necessary functions and we can still use it here like this

api
Botspot: > > This seems overdesigned. On my system it took 4.2 seconds. The single-threaded version took 2.6 seconds on my system. I found a more efficient way to do multithreading:
> 
> did you report these numbers backward? you said the old method (where each refresh had to wait on another) took 2.6 seconds, and my new method took 4.2 seconds

Yes, your multithreaded method actually took 4.2 seconds. This could be attributed to me testing your multithreaded version first, then the original afterwards. The first time could have been a "cold-start" and the dpkg file first needed to be read into cache or something.
Subsequent tests of your method are 2.2 seconds on average, where my multithreaded method is 1.5 on average, on my system.

theofficialgman:
@Botspot I applied the same technique to the daemon. it now multi-thread preloads all the categories https://github.com/Botspot/pi-apps/pull/1932/commits/9b07a92e6fb3b083c6b741e6079b9d7c6e412914

etc/categoryedit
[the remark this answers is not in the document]
Botspot: > could we break out the GUI part of this file (basically the stuff in the while loop) into its own file? that way the Category editor could just source `etc/categoryedit` to obtain the necessary functions and we can still use it here like this

Your suggestion gave me a different idea. There could be a category function in `api` that categoryedit then uses. The category function would contain a waiting function.

Ont other way to solve this would be to implement a lockfile.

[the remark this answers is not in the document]
theofficialgman: > Your suggestion gave me a different idea. There could be a category function in `api` that categoryedit then uses. The category function would contain a waiting function.

well the waiting function in use now wouldn't work, since it relies on the PID of the categoryedit file (which is different from the script that spawned it). I'd rather not use a lockfile since that could result in somehow being permanently locked while I don't think that could really ever happen with the PID check

preload
```diff
@@ -136,56 +134,58 @@ fi
...
   APPS="$(grep -v '/' <<<"$vfiles" | list_intersect "$(list_apps cpu_installable)")"
   
-  #get list of folders - excluding apps - and hide the hidden folder.
-  DIRS="$(grep '/' <<<"$vfiles" | tr -d '/' | grep -vFx "hidden")"
+  #get list of folders - excluding apps.
+  DIRS="$(grep '/' <<<"$vfiles" | tr -d '/')"
   
   #shuffle the list if enabled
-  if [ "$(cat "${DIRECTORY}/data/settings/Shuffle App list")" == 'Yes' ];then
+  if [ "$(while read line; do echo $line; done 2>/dev/null < "${DIRECTORY}/data/settings/Shuffle App list")" == 'Yes' ];then
```
Botspot, line 159: What is the purpose of this change? This seems like an over-complicated way to read one line from a file.

theofficialgman: this loop is the theoretically fastest way to read a file from bash, since while, read, and echo are builtin functions while cat is an external program which has to be loaded into memory

basically its just reducing syscalls in what I found was a "hot" codepath.

etc/preload-daemon
```diff
@@ -2,41 +2,33 @@
 
 #runs in the background and refreshes all the list files
 
+#Only one instance of this script should run at a time. If this is removed, preload will form a fairly good fork bomb.
+if [ "$(pgrep preload-daemon | wc -l)" -gt 2 ];then
```
theofficialgman, line 6: this confuses me... why are there two allowed now? when I moved this up in the file (in the commit that I reverted) the preload-daemon just wouldn't run for some reason...

theofficialgman: oh I see... you just got rid of the reloading entirely

preload
Botspot: > this loop is the theoretically fastest way to read a file from bash, since while, read, and echo are builtin functions while cat is an external program which has to be loaded into memory
> 
> basically its just reducing syscalls in what I found was a "hot" codepath.

I see. Well if that's what you're trying to do, then I'd like to introduce you to your new favorite method!
```
variable="$(< /home/pi/pi-apps/etc/categories)"
echo "$variable"
```
I figured this out by accident. 😂 I'm not sure if this usage is documented, but it appears to work.

theofficialgman: actually I already have another method I employed already for getting the first line of a file (so no while loop)
```
read -r 2>/dev/null line <"${DIRECTORY}/apps/${app}/description"
```

pretty similar, but no variable setting

Botspot: I've gone ahead and removed your while loops and read commands from preload. This probably reduces overhead a little, and it definitely improves readibility.

theofficialgman: your method is slower.... by a lot
in fact, mine is infinitly fast

```
garrett@garrett-usb:~$ time status="$(< "${DIRECTORY}/data/status/${app}")"; echo $status

real	0m0.002s
user	0m0.000s
sys	0m0.002s
installed
garrett@garrett-usb:~$ unset status
garrett@garrett-usb:~$ time read -r 2>/dev/null status <"${DIRECTORY}/data/status/${app}"; echo $status

real	0m0.000s
user	0m0.000s
sys	0m0.000s
installed
```

theofficialgman: (also no this is running on sd card, not a usb drive... I am too lazy to change the system name 😆 )

theofficialgman:
whelp this is the fastest and best for each case. now all the methods are in use

preload
```diff
@@ -285,7 +285,7 @@ $prefix/$app
     for app in $APPS
     do
       if [ -f "${DIRECTORY}/data/status/${app}" ];then
-        echo "${app} ($(< "${DIRECTORY}/data/status/${app}" 2>/dev/null));${DIRECTORY}/apps/${app}/icon-64.png;:exec "\""echo '$prefix/${app}'"\"""
+        echo "${app} ($(while read line; do echo $line; done 2>/dev/null < "${DIRECTORY}/data/status/${app}" 2>/dev/null));${DIRECTORY}/apps/${app}/icon-64.png;:exec "\""echo '$prefix/${app}'"\"""
```
Botspot, line 288: Is there a reason to read all lines from the app's status file when it's always one line anyway?
For cases when you do want to read an entire file, that still wouldn't need a while loop.
```
read -d '\0'
```

theofficialgman: I get that you don't need a while loop just to read one line. the problem is it is the most efficient method.

this is all part of the speedup, when you go from something that takes 0.002 second to something literally less than 0.000 seconds, and do this over 500 times (each app category and all apps in that category run this), that turns into over a second of timesave

theofficialgman:
> Runonce takes time to hash the commands and compare that hash to the runonce-hashes file

actually I went ahead and checked this, its not that bad.... on my system it takes 0.235 sec to run through all the runonce if they are already completed. I reduced this to 0.147 sec without changing the hash (by replacing the grep and sed).

Botspot:
@theofficialgman I'm pretty sure you actually broke `runonce`. Every time you want to read `$script`, you now read stdin. But unless I have misunderstood how I/O works, stdin will only be there to read **once**. Trying to read /dev/stdin a second time will yield nothing.
Here's an example of this in practice:
```
$ echo sleep 10 | ~/preload-speedup/api runonce
runonce(): '' succeeded. Added to list.
```

theofficialgman:
actually your example works fine as well... don't know what your problem is
```
echo sleep 10 | runonce
runonce(): 'sleep 10' succeeded. Added to list.
```

theofficialgman:
oh I see your problem, you are piping to the script api, not runonce. that worked before, yes, but does not now because if you pipe to the api, the stdin does get consumed immediatly.

how scripts use runonce, they pipe to runonce function directly

etc/categoryedit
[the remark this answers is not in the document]
theofficialgman: @Botspot so... I solved this but it didn't need to be solved. you can't have the Category Editor running in the background. the GUI won't load till you exit it, it is blocking 😆 

I have pushed my changes anyway just so you can see what it took to filter out the GUI vs non-gui launch of categoryeditor https://github.com/Botspot/pi-apps/pull/1932/commits/74f0293167f101b031b939be97ffee4f4d848e05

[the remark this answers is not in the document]
theofficialgman: also keep in mind, you can't test out this standalone without an instance of categoryedit running
```
pgrep categoryedit | xargs ps -o pid,cmd h | awk '{print $1, $4}' | awk '$2 != ""' | awk '{print $1}' | sort | head -n 1
```
otherwise it produces false output (it outputs all PIDs since nothing gets piped to xargs)

this is fine because there is always be at least one PID found, since this is running in categoryedit itself

install
```diff
@@ -101,61 +101,10 @@ X-GNOME-Autostart-enabled=true
...
 
-mkdir -p "${DIRECTORY}/data" && cd "${DIRECTORY}/data" || error "Failed to make and enter ${DIRECTORY}/data directory!"
-mkdir -p installed-packages preload settings status update-status categories
-cd $HOME
-
-#hide template app by default
-"${DIRECTORY}/etc/categoryedit" "template" 'hidden' &>/dev/null
-
-#hide duplicates if running in twisteros
-if [ -f /usr/local/bin/twistver ] && [[ $(twistver) == "Twister OS version"* ]]; then
```
theofficialgman, line 112: what is your plan with removing this from the installer?
have you given into my category overrides PR? or have you just given up on the twister duplicates

theofficialgman: @Botspot I would like to have this merged soon. can you answer the above so I know what changes need to be made (if any)?

Botspot:
> @Botspot your lock file will not work. Please revert, I will explain later (don't have time rn). The short of it is that the calculations will be done based on wrong data and the lock file only waits to write it, instead of the previous implementation waiting to calculate

What about also waiting for lockfile when running `read_category_files`?
My main concern with waiting for another process name is that it would respond to unrelated Pi-Apps directories. I sometimes run multiple versions of Pi-Apps at a time, and others within chroots. If we search for a name of a process, the pi-apps versions will see each other and end up waiting for each other. A lockfile solves this issue, which is partially the reason why I did it in terminal-run and manage daemon.

theofficialgman:
> > @Botspot your lock file will not work. Please revert, I will explain later (don't have time rn). The short of it is that the calculations will be done based on wrong data and the lock file only waits to write it, instead of the previous implementation waiting to calculate
> 
> What about also waiting for lockfile when running `read_category_files`? My main concern with waiting for another process name is that it would respond to unrelated Pi-Apps directories. I sometimes run multiple versions of Pi-Apps at a time, and others within chroots. If we search for a name of a process, the pi-apps versions will see each other and end up waiting for each other. A lockfile solves this issue, which is partially the reason why I did it in terminal-run and manage daemon.

Unrelated pi-apps directories are an easy fix. The command I use to filter outputs the full filepath of categories in use. The output is PID FILEPATH ARGS.
Example:

1111 /home/Garrett/pi-apps/etc/categoryedit Audacity hidden

theofficialgman:
> A lockfile solves this issue, which is partially the reason why I did it in terminal-run and manage daemon.

I don't like lockfile because there is always the chance that the lockfile remains, preventing all future instances from functioning (like if the user experiences an unexpected shutdown, sd card removal, out of ram killer, SIGKILL, etc)

> the pi-apps versions will see each other and end up waiting for each other. 

Also keep in mind this isn't a big deal at all to wait anyway. we don't wait at all for the GUI category editor (those are filtered out), and the other processes finish within less than a second. so worst case scenario if you have a process in a chroot and one in normal userspace and they both execute the CLI categoryedit script at the same time, once of them will have to wait a cycle or two of the while loop (0.02 seconds each time). given that the CLI is already executed async, this is perfectly fine

anyway, for normal instances where two pi-apps are launched from different absolute directories, categoryedit is able to filter other directories out https://github.com/Botspot/pi-apps/pull/1932/commits/b9a1e0299a9fb30ce9ad9878a1a806c25cf6776c

Botspot:
> > A lockfile solves this issue, which is partially the reason why I did it in terminal-run and manage daemon.
> 
> I don't like lockfile because there is always the chance that the lockfile remains, preventing all future instances from functioning (like if the user experiences an unexpected shutdown, sd card removal, out of ram killer, SIGKILL, etc)

I simply don't understand how that would be possible. This not an ordinary lockfile like you would see with an office program, it contains a PID. If that PID does not exist, then the script removes the lockfile and proceeds. An unexpected shutdown will not cause issues because on next boot that PID will not exist.
This is exactly like your PID-searching method, except it's not sensitive to outside interference.

> Also keep in mind this isn't a big deal at all to wait anyway.

There's one other reason why I don't like it: a couple of Pi-Apps users have reported that their system froze after launching Pi-Apps. None of us on Github ever figured out what was going on and blamed it on unstable overclock.
Well I actually reproduced this. Once.
I was in a minimal Debian arm64 chroot, and ended up installing Pi-Apps for something. Launching Pi-Apps quickly consumed all CPU, and RAM was filling up fast. Luckily, I have 8GB of it, so I managed to get to tty1 and kill everything I could think of.
`killall preload preload-daemon` fixed it.
Preload-daemon became a ***fork bomb***, as each preload process would run preload-daemon and somehow preload-daemon would fail to detect the pre-existing processes. To this day I'm not sure what went wrong, but the key takeaway was that a process-search can fail.

If at all possible, could a PID-based lockfile work in this case? If `categoryedit` waited for the lockfile to disappear, would that not be equivalent?

theofficialgman:
> > > A lockfile solves this issue, which is partially the reason why I did it in terminal-run and manage daemon.
> > 
> > 
> > I don't like lockfile because there is always the chance that the lockfile remains, preventing all future instances from functioning (like if the user experiences an unexpected shutdown, sd card removal, out of ram killer, SIGKILL, etc)
> 
> I simply don't understand how that would be possible. This not an ordinary lockfile like you would see with an office program, it contains a PID. If that PID does not exist, then the script removes the lockfile and proceeds. An unexpected shutdown will not cause issues because on next boot that PID will not exist. This is exactly like your PID-searching method, except it's not sensitive to outside interference.
> 
> > Also keep in mind this isn't a big deal at all to wait anyway.
> 
> There's one other reason why I don't like it: a couple of Pi-Apps users have reported that their system froze after launching Pi-Apps. None of us on Github ever figured out what was going on and blamed it on unstable overclock. Well I actually reproduced this. Once. I was in a minimal Debian arm64 chroot, and ended up installing Pi-Apps for something. Launching Pi-Apps quickly consumed all CPU, and RAM was filling up fast. Luckily, I have 8GB of it, so I managed to get to tty1 and kill everything I could think of. `killall preload preload-daemon` fixed it. Preload-daemon became a _**fork bomb**_, as each preload process would run preload-daemon and somehow preload-daemon would fail to detect the pre-existing processes. To this day I'm not sure what went wrong, but the key takeaway was that a process-search can fail.
> 
> If at all possible, could a PID-based lockfile work in this case? If `categoryedit` waited for the lockfile to disappear, would that not be equivalent?



> > > A lockfile solves this issue, which is partially the reason why I did it in terminal-run and manage daemon.
> > 
> > 
> > I don't like lockfile because there is always the chance that the lockfile remains, preventing all future instances from functioning (like if the user experiences an unexpected shutdown, sd card removal, out of ram killer, SIGKILL, etc)
> 
> I simply don't understand how that would be possible. This not an ordinary lockfile like you would see with an office program, it contains a PID. If that PID does not exist, then the script removes the lockfile and proceeds. An unexpected shutdown will not cause issues because on next boot that PID will not exist. This is exactly like your PID-searching method, except it's not sensitive to outside interference.

it is sensitive to outside interference, PIDs can (and are) re-used after finished. it can easily happen that lets say the categoryedit gets PID 1000, then an unexpected shutdown occurs, leaving the lockfile in place, and then on next boot some service or script grabs PID 1000. this would prevent any future use of categoryedit.

> If at all possible, could a PID-based lockfile work in this case? If `categoryedit` waited for the lockfile to disappear, would that not be equivalent?

to avoid the above case, you could make it such that the entire script was lockfile based, but check if the PID in the lockfile matches a categoryedit process (easy with the command currently in used with the script). this would prevent the case. again though.... I don't really see much of an issue with not using a lockfile as it is now. The fork bomb is impossible now anyway, since preload-daemon doesn't work the same as it did before.

theofficialgman:
also I have attempted locally to implement a lockfile method and I don't think it can be done. the problem is the lockfile will not exist on first run, and since all processes are "effectively" started at the same time, most of them see that there isn't a lockfile and just pass the wait. the only way to actually see that there is another process running immediately is via the process ID using pgrep

theofficialgman:
> > Also keep in mind this isn't a big deal at all to wait anyway.
> 
> There's one other reason why I don't like it: a couple of Pi-Apps users have reported that their system froze after launching Pi-Apps. None of us on Github ever figured out what was going on and blamed it on unstable overclock. Well I actually reproduced this. Once. I was in a minimal Debian arm64 chroot, and ended up installing Pi-Apps for something. Launching Pi-Apps quickly consumed all CPU, and RAM was filling up fast. Luckily, I have 8GB of it, so I managed to get to tty1 and kill everything I could think of. `killall preload preload-daemon` fixed it. Preload-daemon became a _**fork bomb**_, as each preload process would run preload-daemon and somehow preload-daemon would fail to detect the pre-existing processes. To this day I'm not sure what went wrong, but the key takeaway was that a process-search can fail.

unrelated. there is no forking in categoryedit so this is a non-issue. like I said, the max delay that could happen from this is an additional cycle of the while loop (or however long it takes a categoryedit hidden action to complete). the VERY MINOR delay would only happen if it happened to be that categoryedit was executed by two instances of pi-apps (one in a chroot and another not or in a different chroot in the same directory)

Botspot:
Valid points on the lockfile idea for categoryedit.

I will review the rest of this conversation, but is there anything else to be done before this is merged, @theofficialgman?

Botspot:
I would like to request for `runonce` to be reverted.
```
pi@raspberrypi:~/preload-speedup $ ./gui
    🭊▅▅▅▅🬿
 🭈▂▂█▂▂▂▂█▂▂🬽
 ▌  ▗▖▗▖▗▖  ▐   █▀▀🭍 ▄
 ▌  ▗▖▗▖▗▖  ▐   █▄▄🭞 ▄ ▄▄ 🭂▀▀█ █▀▀🭍 █▀▀🭍 🭂🬰🬰🬰
 ▌  ▗▖▗▖▗▖  ▐   █    █    🭓▄▄█ █▄▄🭞 █▄▄🭞 ▄▄▄🭞
 🭓🬭🬭🬰🬰🬰🬰🬰🬰🬭🬭🭞                  █    █

runonce(): '' succeeded. Added to list.
cat: '/home/pi/preload-speedup/data/settings/App List Style': No such file or directory
runonce(): '' succeeded. Added to list.
Downloading message to /home/pi/preload-speedup/data/announcements... 
runonce(): '' succeeded. Added to list.
runonce(): '' succeeded. Added to list.
runonce(): '' succeeded. Added to list.
runonce(): '' succeeded. Added to list.
runonce(): '' succeeded. Added to list.
runonce(): '' succeeded. Added to list.

Done
pipe is /tmp/tmp.3ry3Azrxmu
runonce(): '' succeeded. Added to list.
runonce(): '' succeeded. Added to list.
```

theofficialgman:
> I would like to request for `runonce` to be reverted.
> 
> ```
> pi@raspberrypi:~/preload-speedup $ ./gui
>     🭊▅▅▅▅🬿
>  🭈▂▂█▂▂▂▂█▂▂🬽
>  ▌  ▗▖▗▖▗▖  ▐   █▀▀🭍 ▄
>  ▌  ▗▖▗▖▗▖  ▐   █▄▄🭞 ▄ ▄▄ 🭂▀▀█ █▀▀🭍 █▀▀🭍 🭂🬰🬰🬰
>  ▌  ▗▖▗▖▗▖  ▐   █    █    🭓▄▄█ █▄▄🭞 █▄▄🭞 ▄▄▄🭞
>  🭓🬭🬭🬰🬰🬰🬰🬰🬰🬭🬭🭞                  █    █
> 
> runonce(): '' succeeded. Added to list.
> cat: '/home/pi/preload-speedup/data/settings/App List Style': No such file or directory
> runonce(): '' succeeded. Added to list.
> Downloading message to /home/pi/preload-speedup/data/announcements... 
> runonce(): '' succeeded. Added to list.
> runonce(): '' succeeded. Added to list.
> runonce(): '' succeeded. Added to list.
> runonce(): '' succeeded. Added to list.
> runonce(): '' succeeded. Added to list.
> runonce(): '' succeeded. Added to list.
> 
> Done
> pipe is /tmp/tmp.3ry3Azrxmu
> runonce(): '' succeeded. Added to list.
> runonce(): '' succeeded. Added to list.
> ```

I do not understand how this can happen... but I have reverted the offending change (I think)
let me know if it works now

theofficialgman:
> I will review the rest of this conversation, but is there anything else to be done before this is merged, @theofficialgman?

besides the issue you just brought up, no. everything should be fine to merge. if it works now for you, you can merge

Botspot:
> I do not understand how this can happen unless you are running old code. I have tested it many times and it always worked... but I have reverted the changes let me know if it works now

I ran `git pull` right before this and there were no conflicts. It's possible that I had made local changes to `runonce`, but then `git` would have said I needed to stash them, right?

theofficialgman:
> I ran `git pull` right before this and there were no conflicts. It's possible that I had made local changes to `runonce`, but then `git` would have said I needed to stash them, right?

no. git would say there are conflicts only if it needed to pull in new changes. if you already had the latest commit, then made local changes, then pulled again, nothing would happen and you local changes would still be there.

regardless its fine, this optimization just made things silly and didn't add to a real speedup. so its fine to have gone.

you should always run `git reset --hard` which will reset any local changes in files that also exist in the git.

Botspot:
This PR, while good, does reduce the UI responsiveness in some scenarios.
When running `gui` from master without any preexisting preload folder, it takes about 1.5 seconds to begin displaying `All Apps`.
But on this branch, `preload` writes to the file and then reads it. As a result, it takes over 3 seconds to begin displaying `All Apps`. (without any preexisting preload folder)

I understand the reason for this change, but could another solution be found for the preload pipe be made synchronous, without risking yad corruption? I would think a lock file could solve this, though there may be unforeseen complications like with `categoryedit`. 🤣

theofficialgman:
> This PR, while good, does reduce the UI responsiveness in some scenarios. When running `gui` from master without any preexisting preload folder, it takes about 1.5 seconds to begin displaying `All Apps`. But on this branch, `preload` writes to the file and then reads it. As a result, it takes over 3 seconds to begin displaying `All Apps`. (without any preexisting preload folder)

all testing I have done found that it improved UI responsiveness in ALL cases. generation of the all apps list now takes less than 1/2 a second ON ITS OWN (with nothing existing before). the first run, all apps lists are generated at once, this increases the individual time for each app to generate, but reduces the time for all the lists since they happen at once, about 4 seconds on my system. but that is 4 seconds from first launch (click of the pi-apps icon). its already done by the time the user clicks on all apps in the gui. when only one list needs regenerating, these changes have improved the generation time from 1.5 seconds to 0.5 seconds for the all apps category.

> I understand the reason for this change, but could another solution be found for the preload pipe be made synchronous, without risking yad corruption? I would think a lock file could solve this, though there may be unforeseen complications like with `categoryedit`. rofl

no, there is no way and there is no need.

theofficialgman:
keep in mind if you have been running an old pi-apps instance before this and its 10 minute timer isn't up, you should kill that preload-daemon before testing this new UI.

Botspot:
Maybe a screen recording will help. I've ran both of these tests multiple times, so icon-caching shouldn't be a factor here.

https://user-images.githubusercontent.com/54716352/173462632-3eb9586f-1de5-4032-905c-af1e2f73adf3.mp4

https://user-images.githubusercontent.com/54716352/173462640-d327e855-b7a1-408c-8ca4-52582bcf082c.mp4

theofficialgman:
> Maybe a screen recording will help. I've ran both of these tests multiple times, so icon-caching shouldn't be a factor here.

refresh the page, I posted one

theofficialgman:
> Maybe a screen recording will help. I've ran both of these tests multiple times, so icon-caching shouldn't be a factor here.

you test removes all folders within the preload folder. this is the worst case scenario where all lists are generated at once simultaneously. this is hard on the cpu but is only performed once, at the first run of pi-apps.
for a more representative test, you should be removing only the one folder which you want to check.

I already explained this here:
https://github.com/Botspot/pi-apps/pull/1932#issuecomment-1145478249
https://github.com/Botspot/pi-apps/pull/1932#issuecomment-1154531794

theofficialgman:
> if the background cpu load is too much for, we can easily revert the [9b07a92](https://github.com/Botspot/pi-apps/commit/9b07a92e6fb3b083c6b741e6079b9d7c6e412914) commit which will change it so that the daemon only loads one category at once. this means that the likelihood that a user clicks on a category that is not loaded on first run increases, but the time it takes to preload that selected category while the background process is occurring is greatly reduced

you can see the result of this change as I make it on the fly.

https://user-images.githubusercontent.com/28281419/173464239-64e91e00-205e-466d-a60e-867e568ae446.mp4


again though, this ONLY affects when the preload-daemon is executing generating the lists at the same time as a user is selecting a category. which in 99% of cases is not the case.

Botspot:
> you test removes all folders within the preload folder. this is the worst case scenario where all lists are generated at once simultaneously. this is hard on the cpu but is only performed once, at the first run of pi-apps. for a more representative test, you should be removing only the one folder which you want to check.

You are correct - for cases with only one preload process, it completes so fast now that there is no reason to make yad display the list as it's being generated.
But that's only for cases when there is one preload process!

> all testing I have done found that it improved UI responsiveness in ALL cases.

While your test is more true to the majority of cases, it is not **all**.

> this is hard on the cpu but is only performed once, at the first run of pi-apps.

Actually, whenever an app or file is modified, all categories will need to be preloaded again due to how `mktimestamps` works.
Combine this with how `preload-daemon` now preloads all categories simultaneously, this will be hard on the cpu during normal operation, not just on the first run.
If the user is switching categories during this high cpu usage, they WILL notice that Pi-Apps seems sluggish.

Botspot:
Maybe the best solution would be to limit the number of simultaneous threads in `preload-daemon` to `$(($(nproc)/2))`

theofficialgman:
> Maybe the best solution would be to limit the number of simultaneous threads in `preload-daemon` to `$(($(nproc)/2))`

I think thats more complexity that necessary. we can't account for all the types of storage medium (usb drives, ssds, emmc, sd cards) or the number of cpus (and their speeds) so we might as well just make it non-multithreaded. on a faster system with faster storage, like my desktop, the whole thing already completes in less than a second anyway, multithreaded or not.

Botspot:
> (performance is also worse in all cases for me, since I am screen recording using xorg)

Setting the frame rate to 5 fps should minimize this effect.

theofficialgman:
> > (performance is also worse in all cases for me, since I am screen recording using xorg)
> 
> Setting the frame rate to 5 fps should minimize this effect.

the sad part is, its not the encoding that is slow or takes up cpu resources (I have nvidia NVENC), it is their stupid slow xorg implementation

PR #1953, opened 2022-06-06
Linux Wi-Fi Hotspot: use precompiled arm64 deb

ryanfortner:
I split the `install` file into separate `install-32` and `install-64`, arm64 script is the only modified one so far since I don't currently have access to a 32-bit buster OS. `uninstall` script remains untouched.

For future reference, here's how to build a deb.
```bash
wget https://github.com/lakinduakash/linux-wifi-hotspot/archive/refs/tags/v4.4.0.tar.gz -O linux-wifi-hotspot-4.4.0.tar.gz
tar -xvf linux-wifi-hotspot-4.4.0.tar.gz
cd linux-wifi-hotspot-4.4.0/
sudo mk-build-deps -i debian/control
dpkg-buildpackage -us -uc -nc
ls ../
```


Botspot:
> so far since I don't currently have access to a 32-bit buster OS.

Vdesktop should work for you when you download this:
https://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2021-05-28/2021-05-07-raspios-buster-armhf-lite.zip

theofficialgman:
I'm going to request that we do not make unnecessary debs. this is a fast compile and its the most compatible with other systems when we leave it as a buildscript

Crilum:
> I'm going to request that we do not make unnecessary debs. this is a fast compile (the compile itself takes less than 10 seconds and I installed the app in less than a minute) and its the most compatible with other systems when we leave it as a buildscript. making something that was previously compatible on bionic, stretch, etc, now not compatible is not a good thing.

I totally agree. If the build takes a long time (over 30 minutes), then a deb is a good idea. But in this case, the install script will build you the latest version (and like @theofficialgman said, it doesn't take long at all), so we don't need a GH Action (or a person) to build debs every time a new version is released.

theofficialgman:
> @theofficialgman How would you make an update script for an app that doesn't download a release file (specifically this app)? Just like this?
> 
> ```
> #!/bin/bash
> 
> webVer=$(get_release lakinduakash/linux-wifi-hotspot)
> 
> source $GITHUB_WORKSPACE/.github/workflows/update_github_script.sh
> ```

that won't do anything because there isn't a version variable in the script to update. right now its building from main, yes, and that is ok, there haven't been anything but readme changes for months.

when another release gets made (if one gets made) we can switch the script to checkout a tag (and add the version= to the script with that tag) and then add the simple auto updater to it

Crilum:
> > @theofficialgman How would you make an update script for an app that doesn't download a release file (specifically this app)? Just like this?
> > ```
> > #!/bin/bash
> > 
> > webVer=$(get_release lakinduakash/linux-wifi-hotspot)
> > 
> > source $GITHUB_WORKSPACE/.github/workflows/update_github_script.sh
> > ```
> 
> that won't do anything because there isn't a version variable in the script to update. right now its building from main, yes, and that is ok, there haven't been anything but readme changes for months.
> 
> when another release gets made (if one gets made) we can switch the script to checkout a tag (and add the version= to the script with that tag) and then add the simple auto updater to it

But that is how you would write the update script? I know that the install script doesn't have the version variable; I was making a PR for adding the update script, but I didn't know exactly how to do it. (Since I usually only work on apps with downloads.)

theofficialgman:
> But that is how you would write the update script? I know that the install script doesn't have the version variable; I was making a PR for adding the update script, but I didn't know exactly how to do it. (Since I usually only work on apps with downloads.)

you don't write an update script. there is nothing to update. there is no version to change in the install

Crilum:
> > But that is how you would write the update script? I know that the install script doesn't have the version variable; I was making a PR for adding the update script, but I didn't know exactly how to do it. (Since I usually only work on apps with downloads.)
> 
> you don't write an update script. there is nothing to update. there is no version to change in the install

I know.. I was changing the install script to clone the code from a Release, and update it when a new version gets released.

theofficialgman:
> I know.. I was changing the install script to clone the code from the latest tag, and update it when a new version gets released.

see:

> when another release gets made (if one gets made) we can switch the script to checkout a tag (and add the version= to the script with that tag) and then add the simple auto updater to it

changing it to the lastest tag at this point would actually downgrade all users installs (since the latest tag is from last year). not idea, its better to leave it as is till a new release is made (you can subscript to the repo to get notified when a new release is made)

Crilum:
> > I know.. I was changing the install script to clone the code from the latest tag, and update it when a new version gets released.
> 
> see:
> 
> > when another release gets made (if one gets made) we can switch the script to checkout a tag (and add the version= to the script with that tag) and then add the simple auto updater to it
> 
> changing it to the lastest tag at this point would actually downgrade all users installs (since the latest tag is from last year). not idea, its better to leave it as is till a new release is made (you can subscript to the repo to get notified when a new release is made)

Nevermind.. I just wanted to know if that's the way you would make the update script.. (and now we have this long thread, lol)

theofficialgman:
> Nevermind.. I just wanted to know if that's the way you would make the update script.. (and now we have this long thread, lol)

if you want to know how to git checkout a version, refer to the DDNet install script and auto updater
https://github.com/Botspot/pi-apps/blob/bd464281c2d20d859d9228b64a3fec19de4a1435/apps/DDNet/install
https://github.com/Botspot/pi-apps/blob/master/.github/workflows/updates/DDNet.sh

Crilum:
Thanks.

ryanfortner:
Personally, I prefer debs because they are easier to install, update, and uninstall on users systems with minimal dependencies. But if you all don’t agree with me, feel free to close this PR.

PR #487, opened 2021-03-20
rpi-play app request

CleanMachine1:
Hi, 
this is my first real commit for your project other than changing txt!
I may have a few errors you can iron out!
I haven't tested my work as I couldn't get pi-apps settings to work very well on LM x86, sorry in advance!

I have only followed the template, so you may need to add some edits to separate files to make it appear in the menu 

Also I didn't understand how the "${DIRECTORY}/pkg-install" bit works so I have installed the dependencies just through apt.

Sorry/You're welcome

PR #1975, opened 2022-06-15
add RPiPlay app

ryanfortner:
Closes https://github.com/Botspot/pi-apps/pull/487

I decided against using debs, because like https://github.com/Botspot/pi-apps/pull/1953, rpiplay is a relatively fast compile and doesn't take up much space.

Also I added a run script (modified from @CleanMachine1 's [here](https://github.com/Botspot/pi-apps/pull/487#issuecomment-828847512)) which asks the user which audio output device (hdmi or analog) they would like to use each time they open the desktop entry.


theofficialgman:
@ryanfortner did you not see the comments in that issue? I don't think we care for RPIPlay anymore since it seems worse in all cases than UXPlay https://github.com/FDH2/UxPlay

https://github.com/Botspot/pi-apps/pull/487#issuecomment-1155306313

ryanfortner:
I saw your comments. I still think rpiplay is worth adding because it aims to run well even on low spec models like Raspberry Pi Zero. Since the GPU on all Raspberry Pi models are not very powerful at all, it's important to at least give a reliable option. 

In my testing, rpiplay works flawlessly on all my pies (I hook them up to a tv frequently and it's quite useful)

ryanfortner:
I'm not saying that UxPlay shouldn't be added though. Both seem like good applications to me 😀

ryanfortner:
I will leave it up to @Botspot if he wants both applications or just one.

Botspot:
Would @ryanfortner be willing to test out uxplay to see if the performance is really worse or not?

PR #1983, opened 2022-06-16
Fix version number with ~ displaying as strikethrough in auto app updates

Jai-JAP:


theofficialgman:
see: https://github.com/Botspot/pi-apps/pull/1646

closing

PR #1986, opened 2022-06-17
add pin priority to local pi-apps repo

theofficialgman:
example affect of pin priority
```
obs-studio:
  Installed: 27.2.0-1
  Candidate: 26.1.2+dfsg1-2
  Version table:
 *** 27.2.0-1 100
        100 /var/lib/dpkg/status
     26.1.2+dfsg1-2 1001
       1001 file:/tmp/pi-apps-local-packages ./ Packages

```
notice how even though the 26.1.2 package (which I placed in my local repo) is of lower version, it is still the candidate for install due to the high priority


Botspot:
One alternative to this would be to have the deb require a particular version of a package. The only edge-case for that would be if a local package and available package were different, but had the same version number.

Botspot:
> we might consider using 990 instead of 1001, see here: https://gist.github.com/JPvRiel/8ae81e21ce6397a0502fedddca068507

If I'm reading that right, it means that in the case of #1981, the affected user would not receive the local package if the one from debian repos is already installed. (Assuming it's a newer version)

theofficialgman:
> > we might consider using 990 instead of 1001, see here: https://gist.github.com/JPvRiel/8ae81e21ce6397a0502fedddca068507
> 
> If I'm reading that right, it means that in the case of #1981, the affected user would not receive the local package if the one from debian repos is already installed. (Assuming it's a newer version)

thats correct. however, the one from the debian repo should not be installed in this case, because the obs script does a apt purge of obs-studio before install_packages

I want install logs from that user before we act on anything here

Botspot:
> I want install logs from that user before we act on anything here

In the case of that user, the core problem was a bug in repo_refresh that has since been fixed in commit https://github.com/Botspot/pi-apps/commit/65c130e73eb90fd879a1272096afd68afd3f3ab4.
This PR would not have helped in that case, though I still think this is useful.

Botspot:
The only time this PR is useful is if the pi-apps local repository has a package that is older than the same package available elsewhere. I'm not aware of this ever happening, but if it does, it may be better to allow the more recent version to be installed.

@theofficialgman, do you see a use for this PR in light of the fix in commit https://github.com/Botspot/pi-apps/commit/65c130e73eb90fd879a1272096afd68afd3f3ab4?

theofficialgman:
I don't think we need the PR. I'll close it and delete the branch. the commit will always be available in the closed PR if we ever determine it is necessary in the future.

Botspot:
I have added these changes in commit https://github.com/Botspot/pi-apps/commit/26ae29d1662d5b56ac2276cc48a29a075e04493f.
This fixes the scratch2 app's new installation method. https://github.com/Botspot/pi-apps/blob/6310c144f909f715ed96611f28fe8dedf91afc4c/apps/Scratch%202/install-32

PR #1989, opened 2022-06-19
compress non-installed apps in the GUI of the updater

theofficialgman:
apps which are not installed and not new apps will be placed into their own special list entry "Pi-Apps Local App Install/Uninstall Scripts" which the user should leave checked if they wish to update all locally present (but non installed) apps

developers may wish to leave this unchecked if they have local uncommitted changes that they do not want to lose

we have seen a few instances where users are confused as to why pi-apps wants to "update" apps that aren't even installed. this clears up that confusion by placing all apps that haven't been installed and are not new into their own toggleable selection in the gui updater


theofficialgman:
also added tooltips for all list entries (for apps this is based on the first line of the description just like the regular yad GUI)

updater
```diff
@@ -324,8 +283,69 @@ list_updates_cli () {  #from https://unix.stackexchange.com/a/673436
...
+      #otherwise display txt icon.
+      mimeicon="${DIRECTORY}/icons/txt.png"
+      mimetype='file'
+    fi
+    
+    LIST+="TRUE
+${mimeicon}
+$file ($mimetype)
+file:$file
+No description"$'\n'
```
Botspot, line 334: I think it would be better to not display any tooltip if no useful information is contained in it.
I just tested yad with an empty tooltip column, and it correctly avoided displaying an empty tooltip box.

updater
```diff
@@ -324,8 +283,69 @@ list_updates_cli () {  #from https://unix.stackexchange.com/a/673436
...
+    
+  done
+
+  # add special entry to list for compressed app updates
+  if [ ! -z "$compressed_update" ];then
+    LIST+="TRUE
+${DIRECTORY}/icons/none-24.png
```
Botspot, line 341: This would probably look better with the `${DIRECTORY}/icons/categories/All Apps.png` icon.

updater
```diff
@@ -324,8 +283,69 @@ list_updates_cli () {  #from https://unix.stackexchange.com/a/673436
...
+Pi-Apps Local App Install/Uninstall Scripts
+compressed_apps
+Applications in your local installation of Pi-Apps that you have not installed yet. This should always remain selected for pi-apps users."$'\n'
```
Botspot, line 344: It might be a good idea to list the apps in the tooltip.

theofficialgman: yeah I thought of that... but the list could get long
should it be one column or space/:/; delimited?

updater
theofficialgman: ah I was looking for that one but couldn't remember the name or find it

updater
Botspot: > yeah I thought of that... but the list could get long should it be one column or space/:/; delimited?

Good point. But curious users would still want to know, as long as the list was less than 10 or so.
So why not do that? :) Show the list of inactive apps if it's less than 10. (and use this delimiter: `, `)

Botspot:
> apps which are not installed and not new apps will be placed into their own special list entry "Pi-Apps Local App Install/Uninstall Scripts"

That title seems quite long, and potentially confusing, as all users will see it and many will wonder what it is. Explaining its purpose is best done in the tooltip anyway, so I think it would be better to use a name like "Inactive apps" or "Unused apps".

theofficialgman:
> Another option would be to hide this entry altogether. I cannot think of any major use-cases where doing that would be a problem.

I originally had it as not an option but then thought developers may wish to not lose their local script changes (see the first commit notes https://github.com/Botspot/pi-apps/pull/1989/commits/26ffa9653bddc8425b115e8e69c4e90a1b4f4e2a)

theofficialgman:
also I figured we need to show an option because otherwise the update list will be empty if only apps that are not installed have changes.

Botspot:
> also I figured we need to show an option because otherwise the update list will be empty if only apps that are not installed have changes.

Speaking of that, it may be good for the update notification and "Updates" list entry to not be shown if the only updates are for inactive apps. Users with fewer apps should have fewer update notifications.

theofficialgman:
> Speaking of that, it may be good for the update notification and "Updates" list entry to not be shown if the only updates are for inactive apps. Users with fewer apps should have fewer update notifications.

we can do that in the background before the regular update check runs if you want. I am not opposed to that

updater
theofficialgman: complete with https://github.com/Botspot/pi-apps/pull/1989/commits/71bd730f0ee6061069d3c4dc6ae8a0d384760776

theofficialgman:
your commit looks good. I was lazy and didn't care too much about the formatting

theofficialgman:
I think as a step 1, this PR is ready.

For the other functionality desired/mentioned (background updating, skipping updating, etc), that can come later

Issue #1988, opened 2022-06-18
Teamviewer Client and Host


GYKgamer:
### What is the name of the app?

Teamviewer Client and Host

### (Optional) Where is the app hosted?

https://www.teamviewer.com/en/download/raspberry-pi/

### About the app

Teamviewer Host
![image](https://user-images.githubusercontent.com/41442793/174435218-e10c93c1-94af-43b3-880e-31cd782308ca.png)

Teamviewer (which is a hidden deb with both the host and client)
![image](https://user-images.githubusercontent.com/41442793/174435278-7f806b3c-ccb4-4840-9488-207ebcfd646d.png)

### Upload file or Add PR Link

https://download.teamviewer.com/download/linux/teamviewer_armhf.deb
https://download.teamviewer.com/download/linux/teamviewer_arm64.deb

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.
- [X] I have confirmed that this app can run on Raspberry Pi.

theofficialgman:
@GYKgamer where is this deb from? the website is teamviewer host which is what our current script is

GYKgamer:
> @GYKgamer where is this deb from? the website is teamviewer host version 14, the latest released version, which is what our current script is
The website, the current version is 14

theofficialgman:
closing as the current deb for teamviewer host in pi-apps is the 15.30.3 version

```
Package: teamviewer-host
Maintainer: TeamViewer Germany GmbH <service@teamviewer.com>
Homepage: http://www.teamviewer.com
Version: 15.30.3
```

GYKgamer:
Wait but Teamviewer Preview is different

GYKgamer:
Host only lets you connect to the raspberry pi

GYKgamer:
Preview lets you connect to and from the raspberry pi

theofficialgman:
preview is the name given by the next version. when 14 was stable, 15 was called next. 15.30.3 has been released and is what the pi-apps script links to. it is not different.

theofficialgman:
nevermind. I see what you are pointing to. there is a hidden link to teamviewer host and client for the pi.

we should migrate all users to this instead of teamviewer host @Botspot 
old link, host only:
```
[...]

PR #1994, opened 2022-06-21
properly and fully rename TeamViewer Host to TeamViewer

theofficialgman:
closes https://github.com/Botspot/pi-apps/issues/1988

before the question gets asked, installing `teamviewer` automatically removes `teamviewer-host` along with its old dummy deb

the methods used in this PR result in TeamViewer Host being renamed to TeamViewer on the users local pi-apps install before the GUI opens, along with renaming it in the categories file. Then, when the user goes to update, there will be a difference in the local TeamViewer (actually the old teamviewer host) and update folder TeamViewer (the new teamviewer host and client). this keeps a functional updater without the user even being aware, while other methods do not keep a functional updater.


gui
```diff
@@ -435,7 +435,14 @@ runonce <<"EOF"
...
+  true
+EOF
+
+#rename TeamViewer Host to TeamViewer
+runonce <<"EOF"
+  #rename current app 'TeamViewer Host' to 'TeamViewer' if not already present
+  [ ! -d "${DIRECTORY}/apps/TeamViewer" ] && mv -f "${DIRECTORY}/apps/TeamViewer Host" "${DIRECTORY}/apps/TeamViewer" && sed -i 's;TeamViewer Host|Internet;TeamViewer|Internet;g' "${DIRECTORY}/etc/categories"
+  mv -f "${DIRECTORY}/data/status/TeamViewer Host" "${DIRECTORY}/data/status/TeamViewer" 2>/dev/null
   true
 EOF
```
Botspot, line 447: Three problems with this line:

1. If the user updates everything, the `${DIRECTORY}/apps/TeamViewer` directory will be created at the same time the `gui` script is updated with this code. As a result, this code will never be run, as it's checking for the `TeamViewer` app to *not* exist.
2. The `etc/categories` file should not be edited for any reason.
3. This whole runonce does not need to be in its own separate location. Please put this app-renaming stuff in the pre-existing runonce and follow that format.

theofficialgman: I was not aware that the app would be re-installed if the update folder contained an app folder that does not exist on the local install and the status file for that app existed.

I tested locally and that does seem to be the case, that explains your method which I thought did not work.

theofficialgman: actually I guess yad now calls the GUI any time the user clicks a button. which is good in this case which means that new runonce files get run without the user re-loading the GUI itself

theofficialgman: @Botspot seehttps://github.com/Botspot/pi-apps/pull/1994/commits/92ae9bd9e229a03521485685d92c66d622f5800c for the proper method which triggers the updater after doing the initial update

theofficialgman: > actually I guess yad now calls the GUI any time the user clicks a button. which is good in this case which means that new runonce files get run without the user re-loading the GUI itself

actually unfortunately the yad called processes never reach the runonce (they exit before any runonce is called) so this does still require a re-launch of the gui

theofficialgman: alright... the question now is. do we hack the runonce into `set-status` like we did with the new-gui update and make this official practice when renaming new apps? because that is the only part of any script that is called after running an update

theofficialgman:
@Botspot I have tested this PR and it works (with the issues that renaming all apps face as mentioned here: https://github.com/Botspot/pi-apps/pull/1994#discussion_r902122148 https://github.com/Botspot/pi-apps/pull/1994#discussion_r902128054

Botspot:
> the question now is. do we hack the runonce into set-status like we did with the new-gui update and make this official practice when renaming new apps? because that is the only part of any script that is called after running an update

We haven't had this problem before because no app has been both renamed *and* updated simultaneously.
I think the solution to delete both app folders and rename the status file, in order to force a reinstall is clunky, but simple.

Is this PR ready to merge now that bitly links are in place?

theofficialgman:
this will not help us now, but maybe we should consider putting our runonce entries into a separate file, so that more scripts other than the GUI can call them (like the updater and cli installation)

> We haven't had this problem before because no app has been both renamed _and_ updated simultaneously.

ah, yes that might be. I guess users that waited long enough on those other apps might have run into this issue. but there isn't anything we can really do about it now.

> I think the solution to delete both app folders and rename the status file, in order to force a reinstall is clunky, but simple.
> Is this PR ready to merge now that bitly links are in place?

yeah it is ready

Botspot:
@theofficialgman, I just realized a problem: updating teamviewer in this fashion means that the new "TeamViewer" app will first be uninstalled for the first time, then installed.

This, in effect, will cancel out the bitly link numbers.
I plan to change the runonce to avoid deleting the "TeamViewer" app-folder. Instead, if the "TeamViewer Host" app is marked as "installed", then the runonce will install TeamViewer.
Let me know if you have any better ideas.

theofficialgman:
> @theofficialgman, I just realized a problem: updating teamviewer in this fashion means that the new "TeamViewer" app will first be uninstalled for the first time, then installed.
> 
> This, in effect, will cancel out the bitly link numbers. I plan to change the runonce to avoid deleting the "TeamViewer" app-folder. Instead, if the "TeamViewer Host" app is marked as "installed", then the runonce will install TeamViewer. Let me know if you have any better ideas.

I'll just change the runonce to remove the TeamViewer app and then move the TeamViewer Host app to Teamviewer. that will fix it
I would prefer the runonce not try to install anything

Botspot:
> I'll just change the runonce to remove the TeamViewer app and then move the TeamViewer Host app to Teamviewer. that will fix it I would prefer the runonce not try to install anything

Hold on. Don't try anything yet.

Botspot:
> wait actually no, what is the concern? canceling out the bitly link numbers is what we want. I copied over the previous data from Teamview Host to Teamviewer in the analytics repo

Wait, we can do that?!
If so, I want to rename `Minecraft Java` to `Minecraft Java Lunar`.

theofficialgman:
yeah your script skips checking for data that already exists in the repo. so you can just copy the folder

Botspot:
> yeah your script skips checking for data that already exists in the repo. so you can just copy the folder
> 
> https://github.com/Botspot/pi-apps-analytics/blob/9893735107ac8afb8e2674dce2655ade41a90cd2/.github/workflows/getclicks.sh#L102

Gotta love those unintentional features. ;)

theofficialgman:
> > wait actually no, what is the concern? canceling out the bitly link numbers is what we want. I copied over the previous data from Teamview Host to Teamviewer in the analytics repo
> 
> Wait, we can do that?! If so, I want to rename `Minecraft Java` to `Minecraft Java Lunar`.

yeah we should be able to do that as well. but lets test the theory with TeamViewer first to see if there are any unexpected results

Botspot:
In that case, I guess the runonce for renaming teamviewer can stay as-is.

theofficialgman:
actually I wonder if this will be a problem in a few months... https://github.com/Botspot/pi-apps-analytics/blob/9893735107ac8afb8e2674dce2655ade41a90cd2/.github/workflows/getclicks.sh#L89

how does the "is a total" work? does that only happen when collecting really old (months old) click data?

Botspot:
> actually I wonder if this will be a problem in a few months... https://github.com/Botspot/pi-apps-analytics/blob/9893735107ac8afb8e2674dce2655ade41a90cd2/.github/workflows/getclicks.sh#L89
> 
> how does the "is a total" work? does that only happen when collecting really old (months old) click data?

I had to review the code to refresh myself on how this works, but the answer to your question is yes. After a certain point, Bitly doesn't share individual days' worth of clicks anymore, it just returns a total number of clicks, from the link-creation time to the date specified.
Hilariously, this limitation is easily circumvented because the *difference* in the totals varies for each day. For example, 12/30/2019 may have 12 clicks total, but 12/31/2019 has 14 clicks. So from that it's easy to derive the day's clicks.

This is of no concern in TeamViewer's case, as the Bitly API is not used for days with existing values.

theofficialgman:
yeah I refreshed myself on it as well. so as long as the new TeamViewer script never needs to collect data from bitly over a month old, the total (limited) data will never be collected.

if we were to collect the total data from bitly, it would be wrong.

I guess this gives us a 1 month deadline to fix the analytics repo in the case it stops updating 😆

theofficialgman:
worst case we can always spam the bitly link X number of times in one day to bring the numbers back up to match the old value

theofficialgman:
actually @Botspot I think we should not copy over the data.
```
elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
  #for this operation, a program name must be specified.
  app="$2"
  if [ -z "$app" ];then
    error "For this operation, you must specify which app to operate on."
  elif [ ! -d "${DIRECTORY}/apps/$app" ];then
    error "${DIRECTORY}/apps/$app does not exist!"
  fi
  
 ........
  
  #analytics
  bitly_link "$app" "$action" &
```

I checked and when the app is removed, the error is triggered so the bitly uninstall link is never clicked.
```
TeamViewer's install script has been updated. Reinstalling TeamViewer...
/home/garrett/pi-apps/apps/TeamViewer does not exist!
Installing TeamViewer...
```
updating doesn't exit though since the manage script doesn't check for errors on update
```
    installback=yes
    status "$app's install script has been updated. Reinstalling $app..."
    #uninstall it
    "${DIRECTORY}/manage" uninstall "$app" update #report to the app uninstall script that this is an uninstall for the purpose of updating by passing "update"
    
    #fix edge case: if app is installed but uninstall script doesn't exist somehow, then pretend app was uninstalled so that the reinstall later will happen noninteractively
    if [ "$(app_status "$app")" == installed ];then
      echo 'uninstalled' > "${DIRECTORY}/data/status/${app}"
    fi
```

we will see in a few days, but I expect very large numbers of install of TeamViewer with basically no uninstall clicks

Botspot:
@theofficialgman, that all seems quite complicated.
Having multiple nested workarounds to improve the accuracy of an inherently unstable system seems like asking for trouble.

For that reason, I've been thinking of moving the invocation of `bitly_link` to only *after* the app successfully installs/uninstalls *and* this is not during an update-reinstall.

theofficialgman:
> @theofficialgman, that all seems quite complicated. Having multiple nested workarounds to improve the accuracy of an inherently unstable system seems unwise.
> 
> For that reason, I've been thinking of moving the invocation of `bitly_link` to only _after_ the app successfully installs/uninstalls _and_ this is not during an update-reinstall.

I didn't propose any changes. This is code from pi-apps as it is right now

Botspot:
See PR https://github.com/Botspot/pi-apps/pull/1999

Botspot:
BTW, I just checked bitly.com and right now `pi-apps-install-TeamViewer` has 18 clicks, while `pi-apps-uninstall-TeamViewer` has only 4 clicks.

theofficialgman:
lets leave proper update tracking till we move to [shlink](https://shlink.io/) (as Jai-JAP said you were looking into)

I'm ok with merging that PR for now, if we plan on making this move to shlink in the future

theofficialgman:
> BTW, I just checked bitly.com and right now `pi-apps-install-TeamViewer` has 18 clicks, while `pi-apps-uninstall-TeamViewer` has only 4 clicks.

those 4 uninstall clicks are all from me. I tested out multiple times and ran the "uninstall" of the new teamviewer app after you added the links but before the PR was merged.

PR #1999, opened 2022-06-21
Eliminate unnecessary bitly calls in pi-apps client to improve accurate user-counts

Botspot:
With this PR, the bitly link is clicked ONLY if the app is successfully installed/uninstalled, *and* if it's not being reinstalled during an update.

### DIFF
--- a/manage
+++ b/manage
@@ -624,7 +624,7 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
   #if app succeeded
   if [ $exitcode == 0 ];then
     
-    #Contribute to app install count
+    #Contribute to app install/uninstall count
     if [ $updating == 0 ];then
       bitly_link "$app" "$action" &
     fi
--- a/manage
+++ b/manage
@@ -622,8 +622,9 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
   #if app succeeded
   if [ $exitcode == 0 ];then
     
-    #Contribute to app install/uninstall count
-    if [[ "$4" != "no-bitly" ]];then
+    #Contribute to app install/uninstall count as long as parent processes is NOT updater (during an app-reinstall)
+    #See: https://askubuntu.com/a/1012236
+    if [ $(cat /proc/$PPID/comm) != updater ];then
       bitly_link "$app" "$action" &
     fi
     
--- a/manage
+++ b/manage
@@ -624,7 +624,7 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
     
     #Contribute to app install/uninstall count as long as parent processes is NOT updater (during an app-reinstall)
     #See: https://askubuntu.com/a/1012236
-    if [ $(cat /proc/$PPID/comm) != updater ] || [ "$3" == "update" ];then
+    if [ $(cat /proc/$PPID/comm) != updater ];then
       bitly_link "$app" "$action" &
     fi
     
--- a/manage
+++ b/manage
@@ -592,9 +592,6 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
     done
   fi
   
-  #analytics
-  bitly_link "$app" "$action" &
-  
   #if this app has scripts, determine which script to run
   if [ "$(app_type "$app")" == standard ];then
     if [ "$action" == install ];then
@@ -615,8 +612,10 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
   echo
   cd $HOME
   if [[ "$3" == "update" ]]; then
+    updating=1
     script_input="update"
   else
+    updating=0
     script_input=""
   fi
   nice "${appscript[@]}" "$script_input" &> >(tee -a "$logfile")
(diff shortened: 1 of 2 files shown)


theofficialgman:
@Botspot this won't work as is because only the uninstall script gets passed `update` from the updater

theofficialgman:
actually I have a solution... ugly but it works

Botspot:
I found a much cleaner solution.
The `manage` process can directly determine if it is being run by `updater`, without relying on any flags or environment variables.

All `manage` needs to do is to check if `$(cat /proc/$PPID/comm)` == `updater`.

Source: https://askubuntu.com/a/1012236

theofficialgman:
> All `manage` needs to do is to check if `$(cat /proc/$PPID/comm)` == `updater`.

I tested this out locally. it works. HOWEVER, it only works if we make sure to NEVER remove `#!/bin/bash` from the top of the `updater` script. otherwise it just returns `bash` and not `updater`

I know this because I tested locally with a script I wrote and only got `bash` to return, until I added the shebang at the top

Botspot:
> I tested this out locally. it works. HOWEVER, it only works if we make sure to NEVER remove `#!/bin/bash` from the top of the `updater` script. otherwise it just returns `bash` and not `updater`

Interesting caveat. Just in case PPID somehow breaks on a marginal setup, I've also made it skip bitly_link if $3 is 'update'. Yes, I understand this won't have any immediate effect as `updater` doesn't pass the update flag, but it might as well be added now, just in case anything ever breaks PPID.

@theofficialgman, do you think anything else should be changed before merge?

theofficialgman:
> @theofficialgman, do you think anything else should be changed before merge?

yes, I do. your commit is incorrect. it should be not equals

manage
```diff
@@ -611,7 +611,7 @@ elif [ "$1" == 'install' ] || [ "$1" == 'uninstall' ];then
   status "${action^}ing \e[1m${app}\e[0m\e[96m..." | tee -a "$logfile"
   echo
   cd $HOME
-  if [[ "$3" == "update" ]]; then
+  if [ "$3" == "update" ]; then
```
theofficialgman, line 614: enlighten me. what is the difference between the [[ ]] and [ ]. I know the former is more versatile but thats about it. if you have some resource that explains (well) that would be nice

Botspot:
> actually @Botspot you can't add that check at all (even though your check is backwards currently).... in the case before the new updater is running. there will be no `update` for uninstall, and there will be `update` for install. this means on update you will get a bitly link click on install and not on uninstall due to the ||

That would only be true if the PPID conditional was not there.

theofficialgman:
you must add a 4th arg as I did here if you want to have an actual check https://github.com/Botspot/pi-apps/pull/1999/commits/c2d11b5640ad06d1966e4bc430068050dcfe50ab . otherwise you break all bitly link clicks which occur during this update (which will likely be a lot)

the 4th arg can not fail

manage
Botspot: I don't have a particularly good source on-hand, but IIRC, the only time double brackets are necessary is for advanced operations like partial matches, regex, `=~`, things like that. Simple `==` or `-gt` don't need double brackets.

Jai-JAP: https://unix.stackexchange.com/questions/306111/what-is-the-difference-between-the-bash-operators-vs-vs-vs

Jai-JAP: @theofficialgman 
> https://unix.stackexchange.com/questions/306111/what-is-the-difference-between-the-bash-operators-vs-vs-vs

theofficialgman:
> That would only be true if the PPID conditional was not there.

incorrect... your check was still backwards. it will click on update uninstall. and even if you reversed it to be correct, it would then click on update install

Botspot:
> there will be no `update` for uninstall, and there will be `update` for install. this means on update you will get a bitly link click on install and not on uninstall due to the ||

At least we're both getting things backwards! (Actually there will be no `update` for _install_, and there will be `update` for uninstall)
You're right about the || - it should have been `&&` with `[ $3 != update ]`

theofficialgman:
no I'm right.. read again your commit. 
I will explain it to you

the updater (before the user restarts after this PR merges) will still call apps like so
for uninstall it passes the update as $3:
```
"${DIRECTORY}/manage" uninstall "$app" update
```
for install, it does NOT pass update as $3 since this didn't exist in the loaded script
```
"${DIRECTORY}/manage" install "$app"
```

here is your commit:
```
if [ $(cat /proc/$PPID/comm) != updater ] || [ "$3" == "update" ];then
  bitly_link "$app" "$action" &
fi
```

assuming $$PID does come out to be updater:

your check for update uninstall will pass, since $3 == update. meaning the bitly link gets clicked, BAD

your check for update install will NOT pass, since $3 does not contain a value, meaning the bitly link does NOT get clicked

notice how in my comment, I said
>  it will click on update uninstall. and even if you reversed it to be correct, it would then click on update install

this is correct with my explanation above

theofficialgman:
> You're right about the || - it should have been && with [ $3 != update ]

anyway, this is fine. with how the script is now, the PPID should function correctly and keep the bitly link from getting clicked. In the future if the PPID check gets messed up, the lack of "update" will allow the bitly link to get clicked

theofficialgman:
> This picture, particularly line 2, seems especially relevant for this PR: 

well said. ignore my argumentative attitude right now. forgive me after watching a that marvel movie which literally had demon summoning in it 😆

PR #2000, opened 2022-06-21
highly optimized set-status function depends

theofficialgman:
this is a draft because this introduces a new dependency to pi-apps, `rsync`, which is an extremely helpful tool for moving (and diffing) files.

the purpose of this PR was to reduce runtime of the set-status function so that it beats out the preload call, and hence would show updates in the GUI on launch without switching tabs
```
    (echo -e '\f' > $pipe
    sleep 0.5
    "${DIRECTORY}/preload" yad '' >> $pipe) &
```

unfortunately, set-status calls `git pull` which on my system takes 0.5 seconds regardless of if the local repo matches master or if it has to pull new changes.

the changes to `get_updatable_files` and `get_updatable_apps` greatly optimize running these functions. each one would take approximately 0.5 seconds to run in their entirety since they called diff on each and every file. the new function calls rsync on the entire folder contents which speeds up runtime by 10X and each function now takes 0.05 seconds

### DIFF
--- a/gui
+++ b/gui
@@ -317,7 +317,7 @@ generate_logo &
 
 #install dependencies
 runonce <<"EOF"
-  dependencies='yad curl wget aria2 lsb-release apt-utils imagemagick bc librsvg2-bin locales shellcheck git wmctrl xdotool x11-utils'
+  dependencies='yad curl wget aria2 lsb-release apt-utils imagemagick bc librsvg2-bin locales shellcheck git wmctrl xdotool x11-utils rsync'
   # Install dependencies if necessary
   if ! dpkg -s $dependencies >/dev/null 2>&1; then
     sudo_popup apt install $dependencies -y -f --no-install-recommends
--- a/install
+++ b/install
@@ -23,7 +23,7 @@ fi
 sudo apt update || error "The command 'sudo apt update' failed. Before Pi-Apps will work, you must fix your apt package-management system."
 
 #install dependencies
-dependencies='yad curl wget aria2 lsb-release apt-utils imagemagick bc librsvg2-bin locales shellcheck git wmctrl xdotool x11-utils'
+dependencies='yad curl wget aria2 lsb-release apt-utils imagemagick bc librsvg2-bin locales shellcheck git wmctrl xdotool x11-utils rsync'
 
 if ! dpkg -s $dependencies &>/dev/null ;then
   sudo apt install $dependencies -y -f --no-install-recommends
--- a/updater
+++ b/updater
@@ -489,6 +489,16 @@ if [ "$runmode" == autostarted ];then #if update-interval allows, and one app in
   updatable_apps="$(get_updatable_apps)"
   updatable_files="$(get_updatable_files)"
 
+  #write to updatable-apps file only if something was changed
+  if [ "$updatable_apps" != "$(cat "${DIRECTORY}/data/update-status/updatable-apps")" ];then
+    echo "$updatable_apps" | grep . > "${DIRECTORY}/data/update-status/updatable-apps"
+  fi
+  
+  #write to updatable-files file only if something was changed
+  if [ "$updatable_files" != "$(cat "${DIRECTORY}/data/update-status/updatable-files")" ];then
+    echo "$updatable_files" | grep . > "${DIRECTORY}/data/update-status/updatable-files"
+  fi
+
   if [ -z "$updatable_files" ] && [ -z "$updatable_apps" ];then
     status "Nothing is updatable."
     exit 0
@@ -541,7 +551,6 @@ if [ "$runmode" == autostarted ];then #if update-interval allows, and one app in
   list_updates_gui
   if [ -z "$updatable_files" ] && [ -z "$updatable_apps" ];then
     status "User did not allow anything to be updated."
-    "${DIRECTORY}/updater" set-status
     exit 0
   fi
   
--- a/updater
+++ b/updater
@@ -541,6 +541,7 @@ if [ "$runmode" == autostarted ];then #if update-interval allows, and one app in
   list_updates_gui
   if [ -z "$updatable_files" ] && [ -z "$updatable_apps" ];then
     status "User did not allow anything to be updated."
+    "${DIRECTORY}/updater" set-status
     exit 0
   fi
   
@@ -553,6 +554,7 @@ if [ "$runmode" == autostarted ];then #if update-interval allows, and one app in
     --image="${DIRECTORY}/icons/logo-64.png" \
     --button="Close!${DIRECTORY}/icons/exit.png":1 \
     --timeout=30
+  "${DIRECTORY}/updater" set-status
   
 elif [ "$runmode" == 'get-status' ];then #Check if anything was deemed updatable the last time updates were checked for.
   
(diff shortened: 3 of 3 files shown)


Botspot:
> unfortunately, set-status calls `git pull` which on my system takes 0.5 seconds regardless of if the local repo matches master or if it has to pull new changes.

It's possible that `git pull` could be avoided if the local repo's most recent commit ID is identical to the online one. This is still reliant on Internet speed, but assuming there were no new commits, this might shave off some time.

theofficialgman:
> It's possible that `git pull` could be avoided if the local repo's most recent commit ID is identical to the online one. This is still reliant on Internet speed, but assuming there were no new commits, this might shave off some time.

yeah but at that point, if the local repo already matches the online repo, then you are guaranteed to already have populated `data/update-status/updatable-apps` and `data/update-status/updatable-files`. since these files will already exist from whenever the repo was last pulled (since they were generated at the same time)

theofficialgman:
the only thing I thought about which could go alongside this is the addition of the `update` selection to all categories like it used to be. with the new gui, it was changed to only appear on the main category selection.

theofficialgman:
no matter what, these are still good changes though, since they shave off at least a second from clicking on the `Updates` button till the update window shows (since it calls these functions before starting)

Botspot:
> https://github.com/Botspot/pi-apps/pull/2000/commits/f016f37ad4615635108d06873603843739925a50

I'm a bit surprised `set-status` needs to be run again after the user dismisses the update dialog. Wouldn't updatable-apps and updatable-files remain intact from the scan that finished a few seconds before?

Botspot:
> Wouldn't updatable-apps and updatable-files remain intact from the scan that finished a few seconds before?

It does not appear to do so. I'd rather fix it like this, rather than to perform a second whole scan unnecessarily.

theofficialgman:
> > Wouldn't updatable-apps and updatable-files remain intact from the scan that finished a few seconds before?
> 
> , rather than to perform a second whole scan unnecessarily.

It isn't unnecessary. Even if it wrote the previous list to the file, it is still necessary to perform the check again. The user could have updated some of the apps, making the saved list from before incorrect.
The check takes less than a second (the pull is the longest part, the scans takes less that .1 seconds total)

theofficialgman:
> I was looking at the place where you added `set-status` to the section that runs if the user closes the update notification.

right, you could change here so that it actually sets the local files instead of just gets the variables. then you wouldn't need to set it if the user closes the update notification

https://github.com/Botspot/pi-apps/blob/9ef9d2e815c600a66e74349280b1c9899694eef5/updater#L524-L527

but you still need to run set-status after if the user chooses to view details and update a few (or all) apps

Botspot:
Changes to the updater script need to be made cautiously. We really don't want to wait for three months!

I found one bug. The `updatable_files` variable was being set to the output of `rsync`, but this subshell was not within double-quotes. As a result, if there were multiple updatable files, this variable would not have been newline-delimited. I have fixed this.

Additionally, I've added a fallback for if the `rsync` command does not exist.

theofficialgman:
> I found one bug. The `updatable_files` variable was being set to the output of `rsync`, but this subshell was not within double-quotes. As a result, if there were multiple updatable files, this variable would not have been newline-delimited. I have fixed this.

actually it worked already with multiple files without double quotes... what are you talking about? did you actually test it like I did? you didn't fix anything that wasn't already working....

theofficialgman:
adding the double quotes is harmless though, so they can stay

theofficialgman:
> Additionally, I've added a fallback for if the `rsync` command does not exist.

I was planning on adding `rsync` as a dependency to pi-apps. in which case, it will always be present. But I'm fine with the fallback existing, in the case that the user runs the updater multiple times before launching the GUI.

Botspot:
> actually it worked already with multiple files without double quotes... what are you talking about? did you actually test it like I did? you didn't fix anything that wasn't already working....

To be honest, no I did not test it this time. Does the lack of the double-quotes always allow for word-splitting without exception? I thought so, but you appear to be right in this case.
I'm starting to wonder if I'm in a parallel universe. :) Has it always been this way? If so, why are quotes used at all when setting variables to the output of commands?

theofficialgman:
> To be honest, no I did not test it this time. Does the lack of the double-quotes always allow for word-splitting without exception? I thought so, but you appear to be right in this case. I'm starting to wonder if I'm in a parallel universe. :) Has it always been this way? If so, why are quotes used at all when setting variables to the output of commands?

rsync outputs newlines between each match. you never needed double-quotes for setting a variable in this case
```
test="$(echo -e "hi\nhello\nhelp")"
garrett@garrett-usb:~$ echo "$test"
hi
hello
help
garrett@garrett-usb:~$ test=$(echo -e "hi\nhello\nhelp")
garrett@garrett-usb:~$ echo "$test"
hi
hello
help
```
there isn't any word splitting here.

> I'm starting to wonder if I'm in a parallel universe.
I just watched Dr Strange and the Multiverse of Madness... me too 😆

theofficialgman:
there might be legitimate uses for "" when setting variables with `"$()"`, but none come to mind right now.

theofficialgman:
I did find this on the internet
`another=$var            # also works, assignment is implicitly double-quoted`

maybe this applies for $() as well? its implicitly double quoted?

theofficialgman:
@Botspot I think including the fallback might be overkill. I checked every piOS image (buster/bullseye lite and full) and rsync is in all of their default package lists

Botspot:
> @Botspot I think including the fallback might be overkill.

It's mostly peace of mind for us developers to know that optimizations have no negative consequences under any circumstances.
Back when Itai wrote a C-based preload program, I made an extensive fallback system to make sure the app list did not break, no matter what crazy system Pi-Apps found itself on.

> I checked every piOS image (buster/bullseye lite and full) and rsync is in all of their default package lists

That's great, but the entire nature of regressions is that they were not anticipated by developers. Especially for something as crucial as the updater, there is no such thing as "too much" safety.

theofficialgman:
rebase on master (since we also changed the updater since this PR was made). I wanted to do final tests and this was the easiest way

theofficialgman:
@Botspot I didn't get to do my checks... and of course there is a bug
not a fatal one, but one you introduced at the last second

theofficialgman:
```
    if command -v rsync >/dev/null ;then
      updatable_files="$(rsync --exclude="/apps/" --exclude="/.git/" --exclude="/.github/" --exclude="/data/" -ric --dry-run --out-format="%n" "${DIRECTORY}/update/pi-apps/" "${DIRECTORY}/")"
      false
    fi

    if [ $? != 0 ] || ! command -v rsync >/dev/null ;then
```
calling `false` always makes the output return an exit code of `1`. making it run the fallback in addition to the rsync (which completed successfully)

why did you put false there?

theofficialgman:
I have reverted the commit... there were other issues with your final change that broke stuff

theofficialgman:
new PR with all issues corrected: https://github.com/Botspot/pi-apps/pull/2005

same spirit as this PR, but without all the bugs from https://github.com/Botspot/pi-apps/pull/2000/commits/c34cd9d44d4432be3be3f7c4e00afecc24254d31 https://github.com/Botspot/pi-apps/pull/2000/commits/20eca59aa7077d305895f481ea34f7d1bf6bbd55

PR #2005, opened 2022-06-23
Set status speedup again

theofficialgman:
lets start back from what worked without any issues...

### DIFF
--- a/updater
+++ b/updater
@@ -489,6 +489,16 @@ if [ "$runmode" == autostarted ];then #if update-interval allows, and one app in
   updatable_apps="$(get_updatable_apps)"
   updatable_files="$(get_updatable_files)"
 
+  #write to updatable-apps file only if something was changed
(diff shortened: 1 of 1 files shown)


theofficialgman:
the cause of the bug has been determined:
https://github.com/Botspot/pi-apps/blob/20eca59aa7077d305895f481ea34f7d1bf6bbd55/updater#L153

botspot added back in the old functions as a fallback but broke the fallback. what happened was the rsync ran, generated the list correctly, and then the fallback ran (due to the additional false) and removed the first letter of the correctly generated rsync.

theofficialgman:
> Moral of the story is to never make changes past 10pm, even if testing seems to complete correctly. What timezone are you in, @theofficialgman?

Oh I'm worse than you in that regard, I was up late with this change. Eastern Standard Time (EST)

PR #2007, opened 2022-06-23
rename Minecraft Java to Minecraft Java Lunar

theofficialgman:
uninstall script has been edited to remove the (now old) minecraft java dummy deb. this will not trigger an update, so until lunar client is next updated in pi-apps, there will be no "Minecraft Java Lunar" dummy deb, but there will still be a "Minecraft Java" dummy deb.


apps/Minecraft Java Lunar/uninstall
```diff
@@ -7,4 +7,8 @@ sudo rm -rf ~/lwjgl2arm32 ~/lwjgl3arm32 ~/lwjgl3arm64 ~/lwjgl2arm64 /opt/jdk/jdk
 echo "Removing Lunar Appimage"
 rm -rf ~/.local/share/applications/Lunar-Client.desktop ~/.local/bin/lunarclient.AppImage
 
+# purge old Minecraft Java dummy deb
+apt_lock_wait
+sudo apt purge -y pi-apps-c6dfb753
```
Botspot, line 12: While this solution is fine and I'm not requesting any changes, I just wanted to point out that purge_packages could technically be used here.
```
app='Minecraft Java' purge_packages || exit 1
```
This same trick can be used for `install_packages`.

Botspot:
This looks good to me.

theofficialgman:
this needs a bitly link created before it can be merged

theofficialgman:
@Botspot I think it is best that we make the Minecraft Java entry hidden. otherwise, if the user installs via the updater in the pi-apps GUI, they will see the old Minecraft Java on the main screen below all categories. once they restart of course, the runonce will run and the folder will be removed and the install status will be changed over to Minecraft Java Lunar

theofficialgman:
I would also like to discuss moving our runonce entries out of the `gui` and into their own file as part of pi-apps that both the `gui` and `updater` can call. this would allow for running the runonce entries without restarting the gui, meaning we could change the status entry without reload

theofficialgman:
see PR: https://github.com/Botspot/pi-apps/pull/2021

PR #2008, opened 2022-06-24
Make VSCode's deb URL hardcoded

cycool29:
For pi-apps to handle the job of updating VSCode when a new version comes, instead of manually downloading the new version by users.


theofficialgman:
@cycool29 @Botspot see my comment above and why it has been reverted

theofficialgman:
as for why the current version of code in the [archive.raspberrypi.](https://archive.raspberrypi.org/debian/pool/main/c/code/) is only 1.66, keep following this for updates: https://github.com/microsoft/vscode/issues/154038

PR #2009, opened 2022-06-24
Update credits of tldr

CleanMachine1:
TLDR is just lots of pages, the software here is unrelated to tldr-pages, it just uses the pages


Botspot:
The original purpose of the Pi-Apps credits file was to track which Pi-Apps developer was responsible for maintaining the app. We didn't create this file to document whoever was behind the app - that's a job better done in the `description`.

PR #2021, opened 2022-07-03
move runonce entries to their own file

theofficialgman:
also call runonce entries as part of the updater set-status function. this effectively moves the runonce entries up in the GUI to occur during the set-status function call

I have tested this, and it does not have any obvious drawbacks to me


Botspot:
I'm trying to make a change to this branch, but there was an error.
```
pi@raspberrypi:/tmp/pi-apps $ git switch 'runonce-file'
fatal: invalid reference: runonce-file
```
Any ideas?

theofficialgman:
git switch? I was not aware that was a thing. I always use `git checkout branchname`

ofc, you need to do a `git pull` first before the branch is available

PR #2052, opened 2022-07-13
Fix LineRider install for Buster

Sussy-OS:
This is my fix for Buster textures dissapearing glitch in LineRider. Please merge this.

### DIFF
--- a/apps/LineRider/install
+++ b/apps/LineRider/install
@@ -1,19 +1,35 @@
 #!/bin/bash
 
-version=2022.5.3
 #Bot automatically updates this when I make a new release.
+version=2022.5.3
+#Buster version
+version2=2022.5.1
 
+status "Installing Depends... "
 install_packages apt-transport-https dirmngr gnupg ffmpeg mono-devel unzip zip || exit 1
+status_green "Done."
 
 mkdir -p ~/LineRider
 cd ~/LineRider || error "Could not move to folder"
 
+if [[ $(get_codename) == "buster" ]]; then
+  use_version="$version2"
+elif [[ $(get_codename) == "bullseye" ]]; then
+  use_version="$version"
+else
+  warning "OS unknown, defaulting to Raspberry PI OS Bullseye install (If you do not have Raspberry Pi OS, ignore this.)"
+  use_version="$version"
+fi
+
 #Download LRA-Community-Edition build from repo
-wget -O linerider.zip https://github.com/Sussy-OS/LRA-Community-Edition/releases/download/${version}/LineRider.game.auto-release.${version}-LR.zip || error "Could not download linerider"
+status "Downloading LineRider... "
+wget -O linerider.zip https://github.com/Sussy-OS/LRA-Community-Edition/releases/download/${use_version}/LineRider.game.auto-release.${use_version}-LR.zip || error "Could not wget LineRider from GitHub!"
 
 #Install
+status "Unzipping... "
 unzip linerider.zip || error "Could not unzip the files"
 rm linerider.zip
+status_green "Done."
 
 #Add system ffmpeg to expected location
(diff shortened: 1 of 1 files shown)


apps/LineRider/install
```diff
@@ -1,19 +1,35 @@
...
 
+if [[ $(get_codename) == "buster" ]]; then
+  use_version="$version2"
+elif [[ $(get_codename) == "bullseye" ]]; then
+  use_version="$version"
+else
+  warning "OS unknown, defaulting to Raspberry PI OS Bullseye install (If you do not have Raspberry Pi OS, ignore this.)"
+  use_version="$version"
+fi
+
```
Botspot, line 23: I think it would be better to not display a warning if the system is neither buster nor bullseye.
Do `if buster;then, else`

Sussy-OS: Like that?

Sussy-OS:
I have tested this on R4k1ta's Buster OS card, (using AnyDesk) so I can verify that this works.

apps/LineRider/install
```diff
@@ -1,19 +1,32 @@
...
 #Download LRA-Community-Edition build from repo
-wget -O linerider.zip https://github.com/Sussy-OS/LRA-Community-Edition/releases/download/${version}/LineRider.game.auto-release.${version}-LR.zip || error "Could not download linerider"
+status "Downloading LineRider... "
+wget -O linerider.zip https://github.com/Sussy-OS/LRA-Community-Edition/releases/download/${use_version}/LineRider.game.auto-release.${use_version}-LR.zip || error "Could not wget LineRider from GitHub!"
 
 #Install
+status "Unzipping... "
 unzip linerider.zip || error "Could not unzip the files"
 rm linerider.zip
+status_green "Done."
```
Botspot, line 29: These "Unzipping... " and "Done." messages are unnecessary for the `unzip` command, as Pi-Apps already has a function in the `api` script called `unzip`. This `unzip` function makes a line of output to say what is being extracted.

See the api line 2353.

apps/LineRider/install
```diff
@@ -1,19 +1,32 @@
...
 
-version=2022.5.3
 #Bot automatically updates this when I make a new release.
+version=2022.5.3
+#Buster version
+version2=2022.5.1
 
+status "Installing Depends... "
 install_packages apt-transport-https dirmngr gnupg ffmpeg mono-devel unzip zip || exit 1
+status_green "Done."
```
Botspot, line 10: These "Installing Depends... " and "Done." messages are unnecessary for the `install_packages` command, as it will already say what is being done. When `install_packages` finishes, it says "Package installation complete."

See the api line 518.

apps/LineRider/install
Sussy-OS: Resolved: [e31ca13](https://github.com/Botspot/pi-apps/pull/2052/commits/e31ca13b030484b318ec881494da2eb9d2a52ae9)

apps/LineRider/install
Sussy-OS: Resolved: [e31ca13](https://github.com/Botspot/pi-apps/pull/2052/commits/e31ca13b030484b318ec881494da2eb9d2a52ae9)

Sussy-OS:
Thanks!

theofficialgman:
@Sussy-OS what is your intention here? keep buster always at version `2022.5.1`. because that is what will happen

Sussy-OS:
> @Sussy-OS what is your intention here? keep buster always at version `2022.5.1`. because that is what will happen

Right now, the only changes from `2022.5.1` to `2022.5.3` are image optimizations (those do not work on Buster) and nothing else. When any changes are made, I will make two builds of LineRider, one without image optimiztation, and one with, for Buster and one for other OS.

PR #2065, opened 2022-07-20
Update Sonic Pi to bullseye debs

Botspot:
I'd like to know if the buster deb correctly installs on @theofficialgman's system that has an older glibc version.


Botspot:
> there is no buster arm64 deb. so there isn't anything to try even

I forgot you were on arm64. Never mind.

theofficialgman:
the funny thing is their debian control file for the bullseye debs it just straight up wrong. it has this in it
```
libc6 (>= 2.15)
```

for me, I don't get to any libc errors on the arm64 package, I believe I am failing earlier than that

```
/opt/sonic-pi/bin/../app/build/gui/qt/sonic-pi: relocation error: /opt/sonic-pi/bin/../app/build/gui/qt/sonic-pi: symbol _ZN17QAbstractItemView11eventFilterEP7QObjectP6QEvent version Qt_5 not defined in file libQt5Widgets.so.5 with link time reference
```
which probably indicates that a newer version of qt5widgets was linked against than the one in bionic. but again, they messed up their control file and allow it to install on older versions:
```
libqt5widgets5 (>= 5.1.0)
```

PR #2155, opened 2022-10-19
add retroarch

Ninoh-FOX:
I am compiler retroarch program with access to online cores and this work great in raspbian OS in rpi 400


Rak1ta:
Hello. Does your version differ from what is in pi-kiss? The script from jmcerrejon, in addition to the binary, installs all other retroarch files.

Ninoh-FOX:
Is a binary, the last version off retroarch (1.12.0) and you can download all files from menu of retroarch in online updater seccion.

Ninoh-FOX:
Anyway I don't see another retroarch in the search section in pi-apps ...

theofficialgman:
I am going to be honest. I don't see this as fitting into pi-apps. Not because retroarch isn't helpful, but because retroarch alone isn't useful.

There are already many projects out there that are much better setup to distribute retroarch in a useful way with automatic retroarch updates (retroarch itself, in addition to core updates) like RetroPie and Lakka. Given that this is also only useful for armhf (a dying platform as its no longer the default distro), this also doesn't have any long term liability until retroarch's buildbot starts making arm64 core builds.

Retroarch also really should not be distributed in binary form at all across the multiple distros that pi-apps supports. Users could be running wayland (in different versions), xorg, or even no display server with legacy direct rendering (that is supported by the legacy gpu drivers and some custom tweaks for retroarch support). You can not produce one binary that supports each of those configurations, they are incompatible compile time options.

PR #2157, opened 2022-10-20
Use `HOME` variable instead of `/home/pi` in CommanderPi description

cycool29:
Users with different usernames can directly copy the command from the description and paste it into the terminal.

### DIFF
--- a/apps/CommanderPi/description
+++ b/apps/CommanderPi/description
@@ -1,4 +1,4 @@
 Easy RaspberryPi4 GUI system management
 Using CommanderPi, you can change overclock settings, bootloader settings, switch kernels, and view performance diagnostics.
 To run: Menu -> Accessories -> CommanderPi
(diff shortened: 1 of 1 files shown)


apps/CommanderPi/description
```diff
@@ -1,4 +1,4 @@
-Easy RaspberryPi4 GUI system managment
+Easy RaspberryPi4 GUI system management
 Using CommanderPi, you can change overclock settings, bootloader settings, switch kernels, and view performance diagnostics.
 To run: Menu -> Accessories -> CommanderPi
-To run in a terminal: /home/pi/CommanderPi/src/start.sh
+To run in a terminal: $HOME/CommanderPi/src/start.sh
```
Itai-Nelken, line 4:
```suggestion
To run in a terminal: ~/CommanderPi/src/start.sh
```
Exact same thing, but it looks better in my opinion.

Itai-Nelken (approved):

PR #2158, opened 2022-10-22
Bump BlockBench to 4.4.3-391

NoozSBC:
It's been a long time since I've touched the codebase and i'm not sure how the automatic updating bot works, so I'm opening this PR just in case. Please close if this is to be handled by the bot!


ryanfortner:
Blockbench updates are handled by the GitHub actions bot, however updates are weekly instead of daily now. Thanks

PR #2162, opened 2022-10-23
Add Golang

slashtechno:
Utilizes [canha/golang-tools-install-script](https://github.com/canha/golang-tools-install-script)


theofficialgman:
I'm personally not a fan of random scripts uploaded to github for multiple reasons
for one, this script you linked still uses go 1.18.1, a version from March of this year. stable 1.19.2 is out now.

also, I don't think this meets our app elegibility rubric as this is a programming language/compiler and doesn't really have any GUI functionality https://github.com/Botspot/pi-apps/issues/185

I'll keep this open for feedback but if it were to be included, we would need to make an exception to the elegibility rubric or change it.

slashtechno:
> I'm personally not a fan of random scripts uploaded to github for multiple reasons
In regards to security. I linked the script to the current commit to prevent changes to the script having an effect until the install script is manually updated


> also, I don't think this meets our app elegibility rubric as this is a programming language/compiler and doesn't really have any GUI functionality #185

I considered this. However, the rubric specifies the following:
> Apps that are not designed for everyday GUI users won't be accepted. (command-line file servers, webservers, etc)

Since Pi-Apps has a category for programming, and many individuals utilize Raspberry Pis for programming, I think it could make life easier for individuals who wish to program in Go, or use `go build` to compile Go programs

ryanfortner:
We do also have other command-line utilities like GitHub CLI, which many mainly GUI-oriented users aren’t likely to use.

slashtechno:
In addition, there are other development focused apps available on Pi-Apps

theofficialgman:
The precedent has not been set for programming languages and compiler to be included in pi-apps. Everything in the programming category is an IDE and related tooling.

We would need a precident set that we would like CLI only code compilers to be included in the programming category. `go` is the same in this sense as `rustc`, `clang`, `gcc`, etc and these are not part of pi-apps as "apps"

slashtechno:
Okay, makes sense

CleanMachine1:
Of course the version in the debian repos is outdated, why not just skip the middle man here?
The installation according to the docs is just unzipping the binary and deleting the old. 

https://go.dev/dl/ https://go.dev/doc/install

This way it can be updated easier and doesn't depend on any script + checksums can be used since they are supplied.

slashtechno:
That's a fair point. However, to install Go, you also need to set up your PATH manually.

ryanfortner:
@slashtechno I do not like the idea of a script running within another script, which is what is going on here. If I were you, I would adapt canha's script into a simpler and shorter one.

slashtechno:
@ryanfortner How do you suggest shortening it?

ryanfortner:
Maybe just simply splitting `install` into separate `install-32` and `install-64` files, then creating a script based on the official Go installation instructions. https://go.dev/doc/install.

slashtechno:
@ryanfortner Okay, will do
I'll try to get the script done this weekend. It'll be hard for me to test it however, since I don't currently have access to my Pi 4.

ryanfortner:
I am willing to help out if you need assistance. Just ping me if so.

slashtechno:
@ryanfortner Is this better? I attempted to get rid of functions which won't be used in the script

apps/golang/install
```diff
@@ -0,0 +1,99 @@
...
+            ARCH=armv6l
+            ;;
+        "armv8")
+            ARCH=arm64
+            ;;
+        esac
+        PLATFORM="linux-$ARCH"
+    ;;
+
+esac
```
theofficialgman, line 26: this won't work. (armhf systems can run arm64 kernels)

if you want to do architecture detection, pi-apps already sets a variable `arch` for you to utilize.
refer to other scripts like the multimc script
https://github.com/Botspot/pi-apps/blob/81ea2d45221300fd8e7a1a827ccad4bb74eabd47/apps/Minecraft%20Java%20MultiMC5/install#L262-L266

slashtechno: Okay, I'll try to work on this later today.

slashtechno: @theofficialgman I think I fixed it in commit d4ec0bfe615c0ea91c23260caab58e3cd1c792c2

apps/golang/install
```diff
@@ -4,26 +4,18 @@ VERSION="1.19.2"
...
+    ARCH="arm64"
     ;;
-
-esac
+  "32")
+    ARCH="armv6l"
+    ;;
+  *)
+    error "arch variable is not set, can not continue"
+    PLATFORM="linux-$ARCH"
```
theofficialgman, line 16: incorrect location for setting the `PLATFORM` variable. look at the previous code again

slashtechno: @theofficialgman Better in b8d7f978380379edf56856517411790d3bbd1cf7?

apps/golang/install
```diff
@@ -0,0 +1,90 @@
...
+  "64")
+    PLATFORM="linux-arm64"
+    ;;
+  "32")
+    PLATFORM="linux-armv6l"
+    ;;
+  *)
+    error "arch variable is not set, can not continue"
+    ;;
+    esac
```
theofficialgman, line 17: indentation issue with `esac`

apps/golang/uninstall
```diff
@@ -0,0 +1,51 @@
...
+            ARCH=armv6l
+            ;;
+        "armv8")
+            ARCH=arm64
+            ;;
+        esac
+        PLATFORM="linux-$ARCH"
+    ;;
+
+esac
```
theofficialgman, line 20: none of this is used in the uninstall script

apps/golang/install
```diff
@@ -0,0 +1,90 @@
+
```
theofficialgman, line 1: both this and the uninstall script are missing bash shebangs
`#!/bin/bash`

apps/golang/credits
```diff
@@ -0,0 +1,4 @@
+(github.com/.../...)
```
theofficialgman, line 1: what is this line?

slashtechno: To signify that the following lines are Github links in the form of `username/repo`

slashtechno:
@theofficialgman Attempted to resolve the issues in commit cf6e56e4d2b4af207691346560b9fcdc6a854035

slashtechno:
> pi-apps scripts are always exclusivly run in bash, no matter what the users default shell program is (this is for compatibility reasons). the script checks the currently running shell, which will always return bash in pi-apps

How can this be alleviated?

theofficialgman:
actually nevermind, it looks like SHELL is a special varaible containing the default login shell. so if the user changes their default shell, it will be correctly interpreted even if the script is currently running in bash.

uninstall script still needs those macos removal edits though

theofficialgman:
I think you need to actually go through and read and understand the scripts. There are quite a few things broken in the uninstall due to missing variables being set from you removing too much.

slashtechno:
> I think you need to actually go through and read and understand the scripts. There are quite a few things broken in the uninstall due to missing variables being set from you removing too much.

Apologies, should be fixed now. I should have noticed the missing variable declaration earlier. 
I tested the uninstall script, and it seems to be working. I don't have my Raspberry Pi currently, so I am unable to test the install script fully.

apps/golang/credits
theofficialgman: why not just add them as links directly then?

apps/golang/install
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+VERSION="1.19.2"
```
theofficialgman, line 2: if you change this and the references to lowercase `version` then our updater will be able to update it at some point as well as our metadata parser will be able to find the app version

theofficialgman: also stable version is now 1.19.3, which is why an updater within pi-apps would be necessary

for personal reference, this will get the current version to add to a pi-apps auto updater script
`curl -s https://go.dev/dl/?mode=json | jq -r '.[0].version'`

slashtechno: I made `VERSION` lowercase, and removed `VERSION="1.19.2"`
Is that all that was needed?

apps/golang/install
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-VERSION="1.19.2"
```
theofficialgman: uhh.. you removed it outright

apps/golang/install
theofficialgman: no, `version=1.19.3` need to exist there... you weren't supposed to remove it, just change from uppercase to lowercase

apps/golang/install
```diff
@@ -0,0 +1,89 @@
...
+if [ -d "$GOROOT" ]; then
+    echo "The Go install directory ($GOROOT) already exists. Exiting."
+    exit 1
+fi
+
+PACKAGE_NAME="go$version.$PLATFORM.tar.gz"
+TEMP_DIRECTORY=$(mktemp -d)
+
+echo "Downloading $PACKAGE_NAME ..."
+if hash wget 2>/dev/null; then
```
theofficialgman, line 50: just use wget directly. pi-apps already has a proper wget function

I also suggest adding error checking which is a pi-apps feature

add something like this on the same line as the wget command
`|| error "Failed to download"`

apps/golang/install
```diff
@@ -0,0 +1,89 @@
...
+
+if [ $? -ne 0 ]; then
+    echo "Download failed! Exiting."
+    exit 1
+fi
+
+echo "Extracting File..."
+mkdir -p "$GOROOT"
+
+tar -C "$GOROOT" --strip-components=1 -xzf "$TEMP_DIRECTORY/go.tar.gz"
```
theofficialgman, line 64: you should also add error testing here (incase the extract command fails then pi-apps needs to report as an installation error)

theofficialgman:
I suggest reading the create an app wiki page: https://github.com/Botspot/pi-apps/wiki/Creating-an-app

apps/golang/uninstall
```diff
@@ -0,0 +1,28 @@
...
+    shell_profile="$HOME/.bashrc"
+elif [ -n "$($SHELL -c 'echo $FISH_VERSION')" ]; then
+    shell="fish"
+    if [ -d "$XDG_CONFIG_HOME" ]; then
+        shell_profile="$XDG_CONFIG_HOME/fish/config.fish"
+    else
+        shell_profile="$HOME/.config/fish/config.fish"
+    fi
+fi
+rm -rf "$GOROOT"
```
theofficialgman, line 14: GOROOT variable not defined

theofficialgman:
I would rather have go installed systemwide instead of only for the current user

follow official go instructions https://go.dev/doc/install

create a custom profile.d script to set the path variable to include the system go installation. I suggest naming it `/etc/profile.d/go.sh`


to be honest there is so much wrong with this installation method that I will probably just rewrite it myself and add you as a co-contributor to the commit message.

@Botspot can you confirm if you would even like this (go/golang) in pi-apps (see here and the following messages https://github.com/Botspot/pi-apps/pull/2162#issuecomment-1288150289 )

PR #2170, opened 2022-11-02
Use Freetube upstream debs

Jai-JAP:


theofficialgman:
I did a quick check to make sure it would be compatible with buster based on glibc (it only requires 2.17) so should be all good assuming it actually works. Have you tested these official builds @Jai-JAP ? the auto updater would have picked up on the new beta for arm64 next time it runs already.

theofficialgman:
edit: well according to at least one user it is working https://github.com/FreeTubeApp/FreeTube/pull/2113#issuecomment-1250191631

I will go ahead and merge since it looks to be well tested

theofficialgman:
good work at getting that merged upstream btw

PR #2178, opened 2022-11-09
Icons update

Rak1ta:
Improving the image quality (it's actually almost imperceptible) of some icons and updating them to match current program logos.


theofficialgman:
If I might ask, what have you done differently when creating the icons that are the same?
I recently just used inkscape for scaling the BTOP logo and correcting the width/height of it

theofficialgman:
transmission icons are correct

transmission is a package app in pi-apps and is version 3.0.0 and 2.9.4 in bullseye and buster respectivly. the current icon is the icon used in those versions. only the main brain (which they have named 4.0.0) has changed the icon.

Botspot:
While we're at it, we might as well fix the two remaining icons I'm aware of that do not have a square aspect ratio: GIMP and Amiberry.
These appear stretched in pi-apps xlunch mode.

As I do not have permission to edit @Rak1ta's branch, I have uploaded these corrected icons to a zip file:
[squared-icons.zip](https://github.com/Botspot/pi-apps/files/9976374/squared-icons.zip)

theofficialgman:
> As I do not have permission to edit @Rak1ta's branch, I have uploaded these corrected icons to a zip file: [squared-icons.zip](https://github.com/Botspot/pi-apps/files/9976374/squared-icons.zip)

unless they changed it since you commented, I just pushed to the branch

PR #2181, opened 2022-11-14
Update webapp-manager to v1.2.5

Rak1ta:
The original [repository](https://github.com/linuxmint/webapp-manager) has had a lot of app updates lately. Notable improvements include the Vivaldi Browser/Chromium Snap and the ability to launch a window in incognito mode. The repo from [phoenixbyrd](https://github.com/phoenixbyrd/Linux_Mint_Webapps) hasn't changed in two years and I made a fork so I can upgrade quickly. Followed this [instruction](https://github.com/hsbasu/webapp-manager/blob/80fc68a13d5c9ce0f224a80788673b1cffe811a4/README.md).

### DIFF
--- a/apps/Web Apps/install-32
+++ b/apps/Web Apps/install-32
@@ -1,10 +1,10 @@
 #!/bin/bash
 rm -rf ~/Linux_Mint_Webapps || error "Failed to delete Linux_Mint_Webapps folder for some reason."
-git_clone https://github.com/phoenixbyrd/Linux_Mint_Webapps || error 'Failed to clone phoenixbyrd web apps repository!'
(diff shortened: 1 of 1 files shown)


theofficialgman:
I will add this to the todo list of looking into myself...
I don't realy like the idea of updating system packages that other normal packages depend on. I will look into building this in such a way so as to not touch system packages.

this will also be hosted at the pi-apps files repo. thanks for the heads up on the updates

theofficialgman:
I determined that the packages are fully backwards compable with older versions so it is safe to upgrade to a point. later than 1.6.0 of `gir1.2-xapp-1.0` and the family of packages can not be used later versions required higher `libglib2.0-0` version than is available in bionic.

PR #2182, opened 2022-11-14
Update Godot to 3.5.1

NoozSBC:
I'm writing this PR from a school iMac since I need godot for a school project 💀

Someone test this new version to make sure it runs, nothing major was changed afaik but just in case...

### DIFF
--- a/apps/Godot/credits
+++ b/apps/Godot/credits
@@ -1,3 +1,3 @@
 Credits to Irtexo for putting it on Pi-Apps.
-Credits to Hiulit for porting it to the Rpi.
-Credits to Juan Linietsky and Ariel Manzur for creating Godot.
+Credits to Hiulit for porting it to the Raspberry Pi.
+Credits to Juan Linietsky and Ariel Manzur for creating Godot Engine.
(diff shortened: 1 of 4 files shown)


ryanfortner:
Your school has iMac's? Mine has crappy Celeron machines 😂

I would consider writing an auto-update script for GitHub actions if I were you.

theofficialgman:
the desktop entry has the version hardcoded. this should be changed to a variable that is set at the top of the script and used for both creating the desktop entry as well as downloading the file

NoozSBC:
The automatic script thing sounds good, I'll try my best

NoozSBC:
@theofficialgman Since your the GH actions expert, I'll have to delegate this to you. The repository pi-apps downloads Godot from has a weird versioning scheme, where the version listed in the release title and tag is different from the one listed in the download URL of the file (https://github.com/hiulit/Unofficial-Godot-Engine-Raspberry-Pi/releases), which means you might have to add a extra function to https://github.com/Botspot/pi-apps/blob/master/.github/workflows/update_apps.yml that specifically grabs the version from the "browser_download_url" of the Github API JSON response instead of from the release tag.

theofficialgman:
ugh... I see that now
I'll write a custom updater for it then.

one thing I noticed is that godot 64bit depends on glibc 2.29+, meaning its being built on bullseye 64bit and won't run on buster 64bit

theofficialgman:
after looking into this further, I don't think this is a suitable candadate for an automatic updater. the package format is non-stable and not the same between 32bit and 64bit which makes me belive that its up to the developers will how it might be packaged each version.

we can still updated though

PR #2118, opened 2022-08-25
switch apt.raspbian-addons.org packages to another mirror

ryanfortner:
As explained [here](https://github.com/raspbian-addons/raspbian-addons/issues/245), the hosting provider for raspbian-addons is no longer renewing the raspbian-addons.org domain. I've switched to a different mirror for the affected packages temporarily until I can sort the issue out.

I also disabled the auto update scripts for the time being since those packages also won't be updated for now. I may have forgotten to switch over a few scripts, but feel free to merge if everything looks good.

PR #2193, opened 2022-12-07
Migrate all raspbian-addons packages

ryanfortner:
Continuing from the original PR [here](https://github.com/Botspot/pi-apps/pull/2118). 

Raspbian-addons is now EOL as per this announcement: https://github.com/raspbian-addons/raspbian-addons/issues/251. This PR switches the locations of the affected packages to either the Pi-Apps-Coders/files repo, or another safe location. I'd like to get this merged ASAP as I don't know how long apt.raspbian-addons.org will still exist.


theofficialgman:
still needs the following apps moved:

KeePassXC
Stunt Rally
SysMonTask

ryanfortner:
> still needs the following apps moved:
> 
> KeePassXC Stunt Rally SysMonTask

Done, thanks.

ryanfortner:
> do you think we should have a runonce to remove the raspbian addons repo (the main one) from a users sources.list.d folder if it is found? Incase a user has added it themselves

I don't think that's necessary since pi-apps didn't actually require the repo to be installed in order for those applications to be installed... so it's on the user to uninstall it if they want.

Botspot:
> I'd like to get this merged ASAP as I don't know how long apt.raspbian-addons.org will still exist.

When a repo site no longer exists, all systems with that repo installed will receive an error during `sudo apt update`. Pi-Apps does not proceed past any `E:` or `Err:` type errors thrown by `apt_update()`.

> do you think we should have a runonce to remove the raspbian addons repo (the main one) from a users sources.list.d folder if it is found? Incase a user has added it themselves

IIRC, a few app scripts used to set up the raspbian-addons repo as part of installtion. But regardless, Pi-Apps will soon fail to work on all systems that have this repo enabled, so I think it is best to make a runonce to minimize the damage from this.

PR #2219, opened 2023-01-02
use theofficialgman install based neofetch fork

theofficialgman:
adds generic (ARM) gpu support and ARM cpu offiicial name support

before and after pictures:
Before:
InkedScreenshot_from_2021-10-08_12-31-31
After:
InkedScreenshot_from_2021-10-08_14-24-32

On a pi4, gpu info for example will show `V3D 4.2` if using the mesa drivers, and `llvmpipe` if using software rendering. Which is helpful debugging information

### DIFF
--- a/apps/Neofetch/credits
+++ b/apps/Neofetch/credits
@@ -1,3 +1,2 @@
 App Credit: dylanaraps - (GitHub: https://github.com/dylanaraps)
-
-Added to Pi-Apps by nikoloiz - (GitHub: https://github.com/nikoloiz)
+Patches: theofficialgman
--- a/apps/Neofetch/description
+++ b/apps/Neofetch/description
@@ -1,3 +1,5 @@
 An aesthetically pleasing bash script to show system information.
 
 To run in terminal: 'neofetch'
+
+This version of Neofetch is better than what can be found by default from APT. It correctly detects CPU and GPU info on ARM systems - details which are hidden in the official version.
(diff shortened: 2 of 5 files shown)


Botspot:
Has a transition from package-app to script-app been tested before?
The first thing I notice is that the updater says this is a "New app".

theofficialgman:
> Has a transition from package-app to script-app been tested before? The first thing I notice is that the updater says this is a "New app".

You transitioned kodi from a package app to a script type app. I think that worked ok but I can test locally for any issues in a few days.

theofficialgman:
> The first thing I notice is that the updater says this is a "New app".

That's because of this test case I added a few months ago. Looks like I forgot to check for the package -> install condition 
https://github.com/Botspot/pi-apps/blob/f774fa019daf2d7e3f5e729de00722add99749d4/updater#L330

Botspot:
I already have a fix for it coming

theofficialgman:
> I already have a fix for it coming

Should just be a matter of adding this to the test case
```
&& [[ ! -f "${DIRECTORY}/apps/${app}/package" ]]
```
I'm away from a computer rn

Botspot:
https://github.com/Botspot/pi-apps/commit/9748579fac0160c46b3d10e82ca2325b88cbf687

theofficialgman:
@Botspot see my edit. It needs to be in both that case and the one above it which is for when an install-${arch} script is added

Botspot:
It passed case 2 in my brief tests.
Oh I see why - you are saying that another transition to script-app could break the scripts into install-32/64 scripts, in which case it would trigger case 2.

PR #2247, opened 2023-01-21
Remove ffmpeg patch

Rak1ta:
Remove the patch because it has already been applied in update 20230119.


theofficialgman:
yup I noticed that as welll
it doesn't cause the install to error but you will get another update notification if I do that

theofficialgman:
closed with https://github.com/Botspot/pi-apps/commit/213bb0a3258d1554d452ef77f52306bf8e18dc63

Issue #685, opened 2021-05-25
GNOME For Pi


TerraGitHuB:
Hi @Botspot !

Looks like a new app just came in...

This is GNOME For Pi (For Pi Apps) based upon this project (mine): https://github.com/TerraGitHuB/gnomeforpi

It removes normal Raspberry Pi OS desktop, and replaces it with GNOME.

GNOME is then configured to work in Raspberry Pi OS.
[GNOMEFORPI-PiAPPS.zip](https://github.com/Botspot/pi-apps/files/6537542/GNOMEFORPI-PiAPPS.zip)

chunky-milk:
> Hi Botspot!
> 
> 
> 
> Looks like a new app just came in...
> 
> 
> 
> This is GNOME For Pi (For Pi Apps) based upon this project (mine): https://github.com/TerraGitHuB/gnomeforpi
> 
> 
> 
> It removes normal Raspberry Pi OS desktop, and replaces it with GNOME.
> 
> 
> 
> GNOME is then configured to work in Raspberry Pi OS.
> 
> [GNOMEFORPI-PiAPPS.zip](https://github.com/Botspot/pi-apps/files/6537542/GNOMEFORPI-PiAPPS.zip)
> 
> 

This looks awesome!

TerraGitHuB:
@chunky-milk Thanks!

Be warned that Pi-Apps implementation will remove Raspberry Pi OS desktop, so be prepared for that...

TerraGitHuB:
It was, but I find it pretty usable, even with animations turned on. Give it a go! I thin I have a pre-built image somewhere

Itai-Nelken:
This looks really nice @TerraGitHuB! unfortunately this cant be added to pi-apps unless there is a functional uninstall script...
it should be easy enough to use something like @techcoder20's [ThemeSaver](https://github.com/techcoder20/ThemeSaver) to save the configuration of the rpi desktop and install the gnome packages using pi-apps's `pkg-install` script.
then to uninstall it, install the rpi desktop and use `purge-installed` to remove all the packages installed for gnome, and use ThemeSaver to reload the users theme.
also make backups of the files the script changes so the uninstall script can revert back.

Botspot:
> This looks really nice @TerraGitHuB! unfortunately this cant be added to pi-apps unless there is a functional uninstall script...
[...]

PR #2268, opened 2023-02-20
Added GNOME

slashtechno:
> this should have been made a native pi-apps script (its as simple as `install_packages gnome gnome-core gnome-software gnome-shell`)  

_Originally posted by @theofficialgman in https://github.com/Botspot/pi-apps/issues/685#issuecomment-1135345893_
            

I edited the install and uninstall scripts to use `install_packages` and `purge_packages` respectively.

### DIFF
--- a/apps/GNOME/credits
+++ b/apps/GNOME/credits
@@ -1,1 +1,1 @@
-TerraGitHuB (inital script) - YT channel: https://www.youtube.com/channel/UCjCHOhXCgQRBHCY8IeBZTzQ
+GNOME Developers - https://www.gnome.org/
--- a/apps/GNOME/credits
+++ b/apps/GNOME/credits
@@ -1,0 +1,1 @@
+TerraGitHuB (inital script) - YT channel: https://www.youtube.com/channel/UCjCHOhXCgQRBHCY8IeBZTzQ
--- a/apps/GNOME/description
+++ b/apps/GNOME/description
@@ -1,2 +1,1 @@
 Installs the GNOME Desktop
-Install GNOME desktop and removes Raspberry Pi OS desktop 
--- a/apps/GNOME/description
+++ b/apps/GNOME/description
@@ -1,0 +1,2 @@
+Installs the GNOME Desktop
+Install GNOME desktop and removes Raspberry Pi OS desktop 
--- a/apps/GNOME/install
+++ b/apps/GNOME/install
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-install_packages gnome gnome-core gnome-software gnome-shell || error "Could not install GNOME and its software."
+install_packages gnome gnome-core gnome-software gnome-shell || exit 1
--- a/apps/GNOME/install
+++ b/apps/GNOME/install
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-install_packages gnome gnome-core gnome-software gnome-shell ||
+install_packages gnome gnome-core gnome-software gnome-shell || error "Could not install GNOME and its software."
--- a/apps/GNOME/install
+++ b/apps/GNOME/install
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+install_packages gnome gnome-core gnome-software gnome-shell ||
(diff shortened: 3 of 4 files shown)


apps/GNOME/install
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+install_packages gnome gnome-core gnome-software gnome-shell ||
```
theofficialgman, line 3: missing exit 1

apps/GNOME/uninstall
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+purge_packages gnome gnome-core gnome-software gnome-shell || error 1
```
theofficialgman, line 3: purge_packages is all you call. it does not take any input

also error 1 is incorrect. that will write `1` to the terminal and exit

apps/GNOME/description
```diff
@@ -0,0 +1,2 @@
+Installs the GNOME Desktop
+Install GNOME desktop and removes Raspberry Pi OS desktop
```
theofficialgman, line 2: incorrect. this does not remove the piOS desktop

apps/GNOME/uninstall
slashtechno: Ah, okay, that makes sense. If I'm understanding correctly,  `error` outputs an error message and exits with exit code `1` whilst `exit` exits with the specified error code - correct?

theofficialgman: yes

apps/GNOME/uninstall
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+purge_packages gnome gnome-core gnome-software gnome-shell || error "Could not remove GNOME and its software."
```
theofficialgman, line 3: still wrong
purge_packages does not take any input
you just call purge_packages || exit 1
that is it

apps/GNOME/install
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+install_packages gnome gnome-core gnome-software gnome-shell || error "Could not install GNOME and its software."
```
theofficialgman, line 3: install_packages already prints to the terminal what it failed to install
you don't need to write any more output, just use exit 1

apps/GNOME/credits
```diff
@@ -0,0 +1 @@
+TerraGitHuB (inital script) - YT channel: https://www.youtube.com/channel/UCjCHOhXCgQRBHCY8IeBZTzQ
```
theofficialgman, line 1: no credit shoud be given to anyone but the gnome developers
this isn't special or unique

slashtechno: I was wondering what the credits should be. I didn't add/remove parties in the credits as I wasn't sure.

slashtechno:
Tried this on ARM64 Pi OS and it didn't install.

slashtechno:
Seems some package was broken.

apps/GNOME/uninstall
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+purge_packages gnome gnome-core gnome-software gnome-shell || exit 1
```
theofficialgman, line 3: still wrong... not going to say why as I have already said it multiple times now

theofficialgman:
> Seems some package was broken.

can confirm that debian and raspbian based systems seem broken here
works on all ubuntu versions. see github actions test build: https://github.com/Botspot/pi-apps/actions/runs/4236698055

slashtechno:
> > Seems some package was broken.
> 
> can confirm that debian and raspbian based systems seem broken here
> works on all ubuntu versions. see github actions test build: https://github.com/Botspot/pi-apps/actions/runs/4236698055

Ah, okay.
Is there any solution to this?

theofficialgman:
> > > Seems some package was broken.
> > 
> > 
> > can confirm that debian and raspbian based systems seem broken here
> > works on all ubuntu versions. see github actions test build: https://github.com/Botspot/pi-apps/actions/runs/4236698055
> 
> Ah, okay. Is there any solution to this?

ask about it in raspberry pi forums
for now this will be closed

PR #2275, opened 2023-03-06
Box86 update

theofficialgman:
depends on merge of this first: https://github.com/ryanfortner/box86-debs/pull/4

reduces complexity of the install scripts, removes libraries that are not possible to be used by box86 and should never have been a part of this script in the first place (boost, ffmpeg, libsnappy, etc).

categorizes packages by type (sound, graphics, sdl1, sdl2, GPU libraries, SSL, etc) and only add ones that are supported by box86 https://github.com/ptitSeb/box86/blob/master/src/library_list.h

### DIFF
--- a/api
+++ b/api
@@ -2727,6 +2727,60 @@ fi
 #Make dual-pane yad windows work correctly on wayland
 export GDK_BACKEND=x11
 
+# add upstream and original lsb-release info for all scripts to use
+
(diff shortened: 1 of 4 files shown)


theofficialgman:
Ubuntu Jammy actually works now without removing the entire 64bit OS so that is good and it can be enabled again.

PopOS jammy is still broken as is PiOS 64bit buster

Issue #1481, opened 2022-02-10
Install steam client (64 bit)


GYKgamer:
### What is the name of the app?

Steam Client

### (Optional) Where is the app hosted?

App is hosted from valve inc

### About the app

You can play steam games (supported with linux / arm64) on the raspberry pi 4 instead of using an existing pc to play.

https://pimylifeup.com/raspberry-pi-steam-client/

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.
- [X] I have confirmed that this app can run on Raspberry Pi.

Botspot:
> ### Confirmations
> * [x]  I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.

Are you sure about that? Steam is already on Pi-Apps and has been for a long time.

GYKgamer:
I've only found steam link, I'm guessing that means it's only in the 32bit version

Botspot:
> I've only found steam link, I'm guessing that means it's only in the 32bit version

You must be running a 64-bit OS then.

Botspot:
> I've only found steam link, I'm guessing that means it's only in the 32bit version

If you would like a challenge, then maybe you would consider trying to get Steam running for 64-bit operating systems? It would help you, and would help many others as well.

GYKgamer:
coding skills = almost 0
wanna try and code this = absolutely

Botspot:
> coding skills = almost 0 wanna try and code this = absolutely

You can start by reviewing the Steam app's install-32 script. I think the hardest part would be to install box86 on a 64-bit OS, but it has been done before. (There are tutorials for it)

GYKgamer:
alright
I will

GYKgamer:
@Botspot I did infact get a script working (it's terrible as hell but does install steam)
But found 2 problems
1. is that it changes your architecture which might give problems installing other apps.
2. I got an error when trying to load into steam "fatal error: Failed to load steamui.so"

Botspot:
Your zip file seems to be corrupted. :(

Could you simply reply with the script in code tags: ```

Botspot:
I've tried downloading your file around 6 times now.
Error output:
```
[...]

PR #653, opened 2021-05-11
Create install-64 for Steam

Jai-JAP:
closes #1481 

Create install-64 for Steam.
And rebase install scripts on https://github.com/ptitSeb/box86/blob/master/install_steam.sh

PR #2276, opened 2023-03-06
Steam: remove install-32 and add install-64

theofficialgman:
closes https://github.com/Botspot/pi-apps/pull/653 closes https://github.com/Botspot/pi-apps/issues/1481

depends on https://github.com/Botspot/pi-apps/pull/2275


Botspot:
What was the rationale behind removing install-32? Was this a result of recent Steam changes that make it incompatible with box86?

theofficialgman:
> What was the rationale behind removing install-32? Was this a result of recent Steam changes that make it incompatible with box86?

@Botspot please refer to the recent commit on the pi-apps repo https://github.com/Botspot/pi-apps/commit/60a8b0c26c80f569840f4793aaf0e62c4f5cb9cd

Botspot:
Test successful on PiOS 64-bit.
At least, I reached the sign-in screen.

To avoid the perception that Pi-Apps updates cause regressions, I want to look into the whole chroot realm again to see how feasible it is here. This would also help bring flatpaks back to armhf.
If others test this PR more thoroughly, I am not opposed to merging it, since existing 32-bit steam installs will remain unchanged.

theofficialgman:
> To avoid the perception that Pi-Apps updates cause regressions, I want to look into the whole chroot realm again to see how feasible it is here.

I don't think it is feasable. since it requires more than 4GB of ram. there is a possibility that you run out of ram that a 32bit kernel can allocate to one process. that wouldn't be an issue with a 64bit kernel on a 32bit userspace though with a 64bit chroot. But I honestly just think that is madness. There really are no legitimate reasons to be using a 32bit OS in 2023. Multiarch works on PiOS bullseye. if there is any software that requires 32bit userspace, it can work in that.

theofficialgman:
> Test successful on PiOS 64-bit. At least, I reached the sign-in screen.

reaching the sign in screen is enough of a test. the sign in screen is CEF (chromium embedded framework) in a react window. if it works, then everything else will.

Botspot:
I was able to sign in today and everything seemed to work okay.

theofficialgman:
intentional by steam (couldn't find the original post, only a screenshot from a discord user): https://cdn.discordapp.com/attachments/700827460454973522/1070478345336270888/Screenshot_2023-02-01-23-58-02-910-edit_com.valvesoftware.android.steam.community.jpg

PR #2286, opened 2023-03-28
flatpak: add theofficialgman flatpak ppa (and switch other apps to use new api functions)

theofficialgman:
currently provides `flatpak` 1.14.4 for bionic/focal/jammy. buster uses bionic repo, bullseye uses focal repo.
also provides `xdg-desktop-portal` and `xdg-desktop-portal-gtk` 1.7.X for bionic/buster

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -334,6 +334,30 @@ This is a special folder (`/tmp/pi-apps-local-packages`) used by the `install_pa
   - This is useful for the `createapp` script to automatically find a suitable icon for package-apps you're trying to add.
   - This uses `dpkg-query` to list all files each package installed. The list is filteres to only show `.png` files in `/icons/` or `/pixmaps/` folders.
   - The list is sorted by filesize to find the picture with the most pixels.
+- `ubuntu_ppa_installer` - Given a ppa name, this function will automatically add the apt repo and key for the current ubuntu dist to the users apt sources.list.d directory.
+  - Example usage:
+    ```bash
+    ubuntu_ppa_installer "ubuntu-mozilla-security/rust-next" || exit 1
+    ```
+  - Removing the repo should be performed as part of an app uninstall script by using the existing pi-apps function `remove_repofile_if_unused`. Repo .list filenames are standardized so the following convention should be used for removal.
+  - Example removal:
+    ```bash
+    ppa_name="ubuntu-mozilla-security/rust-next"
+    ppa_dist="$__os_codename"
+    remove_repofile_if_unused /etc/apt/sources.list.d/${ppa_name%/*}-ubuntu-${ppa_name#*/}-${ppa_dist}.list
+    ```
+- `debian_ppa_installer` - Given a ppa name, dist, and key this function will automatically add the apt repo and key for the specified ubuntu dist to the users apt sources.list.d directory.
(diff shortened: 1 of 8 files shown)


Botspot:
Is the primary purpose for this repo to provide newer versions of flatpak?

Botspot:
Also I would prefer for this to be embedded into the `flatpak_install` function, rather than adding an unnecessary repo for all users who do not use flatpak apps.

theofficialgman:
> We don't want people to try to avoid using pi-apps because it developed a reputation of installing too much unrelated bundleware.

apt go brr `Get:41`

Botspot:
I just noticed that with this runonce it did not actually upgrade flatpak. If some flatpaks require a specific version, then you should probably change this to upgrade the package after running `apt update`.

Botspot:
Is there a reason the ubuntu_ppa_installer and debian_ppa_installer functions have been placed inside this flatpak_install function and are not available to outside functions?

theofficialgman:
> Is there a reason the ubuntu_ppa_installer and debian_ppa_installer functions have been placed inside this flatpak_install function and are not available to outside functions?

no not really

this isn't an urgent PR. let me get back to this tomorrow and I will move them to the api and add some checks to make sure the user/scriptwriter supplies all required input arguments

DOCUMENTATION.md
```diff
@@ -334,6 +334,16 @@ This is a special folder (`/tmp/pi-apps-local-packages`) used by the `install_pa
   - This is useful for the `createapp` script to automatically find a suitable icon for package-apps you're trying to add.
   - This uses `dpkg-query` to list all files each package installed. The list is filteres to only show `.png` files in `/icons/` or `/pixmaps/` folders.
   - The list is sorted by filesize to find the picture with the most pixels.
+- `ubuntu_ppa_installer` - Given a ppa name, this function will automatically add the apt repo and key for the current ubuntu dist to the users apt sources.list.d directory.
+ - Example usage:
+    ```bash
+    ubuntu_ppa_installer "theofficialgman/flatpak-no-bwrap"
+    ```
+- `debian_ppa_installer` - Given a ppa name, dist, and key this function will automatically add the apt repo and key for the specified ubuntu dist to the users apt sources.list.d directory.
```
Botspot, line 342: It may be worth mentioning the naming scheme for the sources.list.d file. Future script-writers should probably know how to remove a repository when an app is uninstalled.

theofficialgman: done

theofficialgman:
@Botspot anything else? looks good to me

api
```diff
@@ -693,17 +693,17 @@ flatpak_install() { #install an app using flatpak
     buster)
       debian_ppa_installer "theofficialgman/flatpak-no-bwrap" "bionic" "0ACACB5D1E74E484"
       apt_lock_wait
-      sudo apt-get --only-upgrade install flatpak -y
+      sudo apt --only-upgrade install flatpak -y
```
theofficialgman, line 696: I generally use apt-get when I can since it avoids the warning in logs that apt is not a stable api

Botspot: Resolved in https://github.com/Botspot/pi-apps/pull/2286/commits/2e0bfe556fcb4513da7e20942f68b6fe8e144b0d

PR #2303, opened 2023-04-14
DDNet: update to 16.9

Kolpixx:


theofficialgman:
Please test on all targets (buster, bullseye 32 and 64bit) to make sure it still builds.

Unable to test in CI since building rust code still crashes GitHub actions qemu.

PR #2340, opened 2023-05-30
use manage daemon for GUI updates

theofficialgman:
updates for apps now occur in the GUI manage daemon just like app installs and uninstalls. it is now safe for the user to start updates and then continue to install/uninstall apps during the update process since they all share the same manage daemon.

the onboot updater now also uses the manage daemon as a result and users can start updates from the onboot updater and then open pi-apps and immediately start installing/uninstalling applications


updater
```diff
@@ -517,43 +517,20 @@ update_now_cli() { #input: updatable_files and updatable_apps variables
...
-      echo
-      echo -e "\e[92mAll updates complete. Closing in 10 seconds.\e[39m"
-      
-    ' "Updating $(echo "$updatable_apps" | wc -l) app$([ $(echo "$updatable_apps" | wc -l) != 1 ] && echo s)..."
+    for app in $updatable_apps ;do
+      terminal_manage update "$app" &
+      # terminal_manage (which calls etc/terminal-run) is sensitive to race conditions. If terminal_manage is called before the first terminal-run terminal opens, multiple terminals will open.
+      # the for loop is very fast and without a sleep would result in the number of terminals equaling the number of updatable_apps which is not the intention.
+      # the intention of terminal_manage is for there to only be one terminal that all manage executions use sequentially
+      sleep 5
```
theofficialgman, line 531: note I would like for this to not be the case but for now its a useable workaround.

Ideally we would only wait for the terminal to exist (max 5 seconds copied from the sleep loop in `etc/terminal-run`) and then start adding updates to the queue as fast as possible. I don't see any way to check for the existance of the terminal thought without knowing `temp_pid_file` which we can't from within this function itself.

theofficialgman: slightly mitigated in https://github.com/Botspot/pi-apps/pull/2340/commits/4c36d7e3630c7341a851402ce5da856975affa6d

manage
```diff
@@ -417,6 +417,9 @@ ${DIRECTORY}/icons/none-1.png
   done
   #all actions have been completed. Daemon has effectively stopped listening, so remove its pid file
   rm -f "${DIRECTORY}/data/manage-daemon/pid"
+
+  #updates could have been run as part of the manage-daemon, so update the updatable-files and updatable-apps status files
+  "${DIRECTORY}/updater" set-status
```
Botspot, line 422: Is there a specific reason you have it check for updates between the current updates finishing and the error-explanation window opening? Could it be moved elsewhere or run as a background process?

manage
```diff
@@ -599,27 +602,64 @@ Or on Discord: \e[94m\e[4mhttps://discord.gg/RXSTvaUvuu\e[0m" | tee -a "$logfile
   #exit the manage script with the same exit-code of the app's script
   exit $exitcode
   
-elif [ "$1" == 'update' ];then #The manage script no longer handles updates. This mode is only for backwards-compatibility and uses the updater script
-  
-  warning "The manage script ONLY updates apps, and this mode has been replaced by the updater script.
-If you want to update Pi-Apps from the command-line, please use:
-~/pi-apps/updater cli-yes"
-  
+elif [ "$1" == 'update' ];then
```
Botspot, line 605: It seems unnecessary to duplicate the app update logic. Do you intend to remove the `update_app()` function from `updater`?

updater
```diff
@@ -692,15 +675,6 @@ if [ "$runmode" == autostarted ];then #if update-interval allows, and one app in
   
   update_now_gui
   
-  #display notification saying that pi-apps updates are complete
```
Botspot, line 695: I am fine with this removal.

manage
theofficialgman: what error explaination window? this location happens after all install/uninstall/updates have fully completed. the error explanation window happens after each individual install/uninstall/update. that happens well before this gets run

Botspot: > what error explaination window? this location happens after all install/uninstall/updates have fully completed. the error explanation window happens after each individual install/uninstall/update. that happens well before this gets run

See the `diagnose_apps` line a few lines beneath this line. Current behavior is to accumulate errors and allow the user to click through all of them at the end of updates.

manage
theofficialgman: the `update_app` function is still used by the CLI updater, so for now it has to stay.

I might change the `update_app` function so that it can be used within the manage update. I had made some very minor changes to the implementation in the manage script that made it so I couldn't use the version from the updater script

manage
theofficialgman: ah my bad I thought from memory that diagnose_apps (the GUI) appeared immediatly after an app failure

I can move `set-status` after diagnose_apps. basically I want it to run after app updates are all complete otherwise the updatable_apps and updatable_files files do not get updated, meaning the "Update" row still appears in the GUI and the update GUI still shows these apps/files as updatable.

manage
theofficialgman: resolved with https://github.com/Botspot/pi-apps/pull/2340/commits/1bb40287c7822ea1136da2e6648d10edda4db0fb

manage
theofficialgman: resolved with https://github.com/Botspot/pi-apps/pull/2340/commits/80f7e4cc741792e9cf9460ccabe6f6f50c103837

icons/refresh.png
Botspot, line 1: This image appears to be identical to the backup.png, and for a refresh icon should be mirrored horizontally.

theofficialgman: yeah it is identical
I figured you would want to make a custom one for it 😄

theofficialgman:
@Botspot I am running into an issue with attempting to implement this:
> Update check for user installing out of date app to not prompt if the app is currently scheduled for a refresh in the manage daemon

The problem is that `validate_apps_gui` is run on the app as the user selects it, not when it actually gets installed by the manage-daemon. The instance of the manage-daemon where the user selects the app to install has no way of knowing what is currently in the queue. The user could already have the app refresh in the queue, in which case they shouldn't be prompted to update.

There isn't really any harm though in not handeling this edge case, there will just be duplicate refresh and update entries in the manage-daemon when the user inevitably selects the app to update again in the update GUI. I could solve the duplicate entries by simply removing them as part of the `reorder_list` function if we wanted to.

What are your thoughts?

Botspot:
I've been monitoring this PR and reading the code bit by bit in my spare time. (I'm starting a business and taking an accelerated summer course so my time is short)
I can probably do a full review tomorrow afternoon.

theofficialgman:
This PR is finalized from me now unless anyone requests changes.

manage
```diff
@@ -299,8 +356,14 @@ $app"
...
-  queue="$(validate_apps_gui "$queue")"
-  [ -z "$queue" ] && exit 0
+  if [[ "$queue" =~ ^"update;filelist:".* ]];then
+    # the queue is being used to pass a list of files to update
+    # skip app validation
+    true
+  else
+    queue="$(validate_apps_gui "$queue")"
+    [ -z "$queue" ] && exit 0
+  fi
```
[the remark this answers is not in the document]
theofficialgman: it doesn't
this is under the `if [ "$1" == 'daemon' ];then` where the only input that is ever passed is one line.

https://github.com/Botspot/pi-apps/blob/a2b205e40f8eae27e4affc3c5508a2f6d9671391/manage#L214C1-L366

`"${DIRECTORY}/manage" daemon` is only ever called with one line of input
https://github.com/Botspot/pi-apps/blob/a2b205e40f8eae27e4affc3c5508a2f6d9671391/api#L699
https://github.com/Botspot/pi-apps/blob/a2b205e40f8eae27e4affc3c5508a2f6d9671391/api#L716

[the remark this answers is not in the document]
theofficialgman: I would like to change it so that we could pass multiple lines of queue to daemon at once but that would require a refactor of the `terminal_manage` function which, as shown above, is the only function that calls `"${DIRECTORY}/manage" daemon`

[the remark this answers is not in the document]
theofficialgman: >Why not rather change validate_apps_gui to skip any lines that include `update;filelist:`?

implemented in https://github.com/Botspot/pi-apps/pull/2340/commits/93c9ed847a6cf8f93f13406cd671196f4e7f1851 even though multiple lines of input can never occur as mentioned above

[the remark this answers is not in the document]
theofficialgman: > I would like to change it so that we could pass multiple lines of queue to `"${DIRECTORY}/manage" daemon` at once but that would require a refactor of the `terminal_manage` function which, as shown above, is the only function that calls `"${DIRECTORY}/manage" daemon`. the problem is `terminal_manage` takes two arguments, the first is the action, and the second is the appname.

solved and implemented with
https://github.com/Botspot/pi-apps/pull/2340/commits/b3933ec55d0366c55c94fac74fc148f88c77fbf6
now being used in the updater (for speed)
https://github.com/Botspot/pi-apps/pull/2340/commits/d5b8f386c9c18ac3c84611e638b95ae925e497a0

api
```diff
@@ -675,6 +675,73 @@ debian_ppa_installer() { #setup a PPA on a Debian distro. Arguments: ppa_name di
...
+  else
+    error "terminal_manage(): Must specify an action: either 'install' or 'uninstall' or 'update' or 'refresh'"
+  fi
+  
+  app="$2" #one app name (or filelist of the format 'filelist:path/to/file:path/to/another/file')
+  terminal_manage_multi "$action $app"
+}
+
+terminal_manage_multi() { #function to install/uninstall/update/refresh multiple apps (or filelists) - uses a terminal and refreshes the app list
+  queue="$1" #one or multiple actions and app names (or filelists of the format 'filelist:path/to/file:path/to/another/file')
```
theofficialgman, line 698: also just to note, I did not bother to add any complicated checks on the contents of the `queue` here since the `"${DIRECTORY}/manage" daemon "$queue"` already validates the contents itself. doing so here would be redundant code.

api
```diff
@@ -675,6 +675,73 @@ debian_ppa_installer() { #setup a PPA on a Debian distro. Arguments: ppa_name di
...
+      ' "Terminal Output"
+    
+    # Check if terminal-run failed to launch. GUI users don't see any terminal output if it fails (since there is no terminal open) so we need to prompt them with a GUI window
+    if [ "$?" != 0 ]; then
+      echo -e "Unable to open a terminal.\nDebug output below.\n$(DEBUG=1 "${DIRECTORY}/etc/terminal-run" 2>&1)" | yad --center --window-icon="${DIRECTORY}/icons/logo.png" \
+      --width=700 --height=300 --text-info --title="Error occured when calling terminal-run" \
+      --image="${DIRECTORY}/icons/error.png" --image-on-top --fontname=12 \
+      --button='OK'
+      if echo "$queue" | grep -q "^update filelist:" ;then #list of files separated by :
+        updatable_apps='' updatable_files="$(echo "$queue" | grep "^update filelist:" | sed 's/^update filelist://g' | tr ':' '\n')" no_status=true update_now_cli
```
Botspot, line 740: This is an api function that calls `update_now_cli`, a function that I am pretty sure is still in the `manage` script, though I have not finished reviewing your changes. Did you assume that `manage` is already sourced when it reaches this point in this terminal_manage_multi function?

api
```diff
@@ -675,6 +675,73 @@ debian_ppa_installer() { #setup a PPA on a Debian distro. Arguments: ppa_name di
     apt_update || exit 1
   fi
 }
+
+terminal_manage() { # wrapper for the original terminal_manage function to terminal_mange_multi
+  #function to install/uninstall/update/refresh one app (or filelist) - uses a terminal and refreshes the app list
```
Botspot, line 680: Is there a backwards-compatibility reason you chose to keep terminal_manage and terminal_manage_multi? My first inclination would be to either add multiline functionality to terminal_manage or to rename the function, updating all scripts which call the previous name.

updater
```diff
@@ -515,45 +516,53 @@ update_now_cli() { #input: updatable_files and updatable_apps variables
...
+          break
+        fi
+        sleep 0.01
+        attempts=$((attempts + 1))
+      done
+    fi
+  }
+  #update any files with terminal_manage. if it fails terminal-run will update the files in the background. This is a safety measure to update pi-apps scripts if terminal fails to launch.
+  if [ ! -z "$updatable_files" ]; then
+    queue+="update "$(echo "filelist:$updatable_files" | tr '\n' ':' | sed "s,:$,,")""$'\n'
```
Botspot, line 545: Is it intentional to leave the $() subshell outside of double-quotes here?

Botspot:
I like the majority of this PR but did not like the dissonance between the updater's preview of individual files but the manage daemon's lumping them altogether. I have changed manage-daemon to display each updatable file on a separate line, with an informative icon based on the minetype of the file (to match what updater does), while removing your introduction of a new delimiter your `filelist` approach.
I also disliked how the manage "update" mode was being hijacked to update a list of files: I have always intended for manage to be a close approximation to dpkg for debian, and this worked against that ideal.

Now a new action type is introduced for updating files: `update_file`. When this action is encountered by manage daemon, it will directly run update_now_cli instead of launching a subprocess of `manage`.

It is late here and I was not able to test any of these changes yet. But they seem to be fairly trivial and I doubt it will take me long to do it later.

api
theofficialgman: Yes, read the commit description

updater
theofficialgman: not it wasn't intentional but I don't think it caused any issues

theofficialgman:
> Now a new action type is introduced for updating files: `update_file`. When this action is encountered by manage daemon, it will directly run update_now_cli instead of launching a subprocess of `manage`.

The reason I "hijacked" `update` was so that I didn't have to add another action type. These are still updates, just file updates, and the action type was used to set the name that gets written to the GUI (`updating`, `updated`, etc)

updater
```diff
@@ -515,45 +516,53 @@ update_now_cli() { #input: updatable_files and updatable_apps variables
...
+          break
+        fi
+        sleep 0.01
+        attempts=$((attempts + 1))
+      done
+    fi
+  }
+  #update any files with terminal_manage. if it fails terminal-run will update the files in the background. This is a safety measure to update pi-apps scripts if terminal fails to launch.
+  if [ ! -z "$updatable_files" ]; then
+    queue+="$(echo "$updatable_files" | sed 's/^/update_file /g')"$'\n'
```
theofficialgman, line 545: this will add a file as a new line in the queue for each individual file which is what I was trying to avoid. running manage takes about 2 seconds minimum per action regardless of what you do (due to sourcing api and other things)
the input:
```
file1
file2
file3
```
```
update_file file1
update_file file2
update_file file3
```

manage
```diff
@@ -29,12 +30,12 @@ validate_apps_gui() { #Given a list of actions and apps, graphically notify the
...
   local queue="$1"
   
-  #Ensure that the first word of each line is 'install' or 'uninstall'
+  #Ensure that the first word of each line is 'install' or 'uninstall' or 'update' or 'refresh'
   local i=1 #track the line number in queue
   for line in $queue ;do
     action="$(echo "$line" | awk -F ';' '{print $1}')"
     
-    if [ "$action" != install ] && [ "$action" != uninstall ] && [ "$action" != update ];then
+    if [ "$action" != install ] && [ "$action" != uninstall ] && [ "$action" != update ] && [ "$action" != refresh ];then
```
theofficialgman, line 38: `update_file` will fail this check

manage
```diff
@@ -599,27 +690,17 @@ Or on Discord: \e[94m\e[4mhttps://discord.gg/RXSTvaUvuu\e[0m" | tee -a "$logfile
   #exit the manage script with the same exit-code of the app's script
   exit $exitcode
   
-elif [ "$1" == 'update' ];then #The manage script no longer handles updates. This mode is only for backwards-compatibility and uses the updater script
-  
-  warning "The manage script ONLY updates apps, and this mode has been replaced by the updater script.
-If you want to update Pi-Apps from the command-line, please use:
-~/pi-apps/updater cli-yes"
-  
+elif [ "$1" == 'update' ] || [ "$1" == 'refresh' ];then
```
theofficialgman, line 693: `update_file` has no corresponding action in the manage script. nothing will run for the files

theofficialgman:
> I like the majority of this PR but did not like the dissonance between the updater's preview of individual files but the manage daemon's lumping them altogether. I have changed manage-daemon to display each updatable file on a separate line, with an informative icon based on the minetype of the file (to match what updater does), while removing your introduction of a new delimiter your `filelist` approach. I also disliked how the manage "update" mode was being hijacked to update a list of files: I have always intended for manage to be a close approximation to dpkg for debian, and this worked against that ideal.

It was significantly better before. now the state of the PR is broken. I could have easily made the files appear as multiple lines in the YAD UI while still have them as a filelist so that they could be updated at once if you had asked. Dropping your commits to do that since this is too broken

manage
Botspot: Manage daemon will avoid running manage at all for update_file.

Botspot: See line 479.

manage
```diff
@@ -219,8 +235,23 @@ if [ "$1" == 'daemon' ];then
     # $4 specifies where the output should go. Default is to send the output to the yadlist pipe, but if value is 'stdout', output to stdout.
     local output="$4"
     
+    #choose an icon for this app or file being updated/installed
     if [ -f "${DIRECTORY}/apps/$app/icon-24.png" ];then
       local icon="${DIRECTORY}/apps/$app/icon-24.png"
+    elif [ "$action" == update_file ];then
+      #determine mimetype of file to display an informative icon in the list
+      if [ "$(file -b --mime-type "${DIRECTORY}/${file}")" == 'text/x-shellscript' ];then
```
theofficialgman, line 243: `file` variable does not exist

theofficialgman:
> It was significantly better before. now the state of the PR is broken. I could have easily made the files appear as multiple lines in the YAD UI while still have them as a filelist so that they could be updated at once if you had asked. Dropping your commits to do that since this is too broken

and now I have done this. easy https://github.com/Botspot/pi-apps/pull/2340/commits/cdaf24318e60831146513f1ca71ba6c2cc34ae29

api
theofficialgman: `update_now_cli` has not changed (except fixing the error code) and still resides in the `updater` script. `terminal_manage_multi` is being called from the `updater` so the function is guaranteed to be available

theofficialgman:
@Botspot please do not push. changes are final and fully functional.
Feel free to leave another review

updater
theofficialgman: In this specific usecase, because we are writing to one line anyway, it makes no difference if it is quoted or not.

theofficialgman:
Rebased on master.

you can test on a local install by
```
cd ~/pi-apps/update/pi-apps
git pull origin updater-manage
```
and then using pi-apps builtin updater

theofficialgman:
also just to note, if you want another `action` called `update-files` or something similar I can do that. I just don't want to call something like `update_file` for each app in its own instance (like your previous changes did). that will take much longer to execute (1-2 seconds per file updated) than calling it once for all files to update.

manage
theofficialgman: yeah I saw that. that only occurs if you run the action through the manage daemon. if you simply wanted to use manage to update files you can't do that with your changes.

Botspot:
> Rebased on master.
> 
> you can test on a local install by
> 
> ```
> cd ~/pi-apps/update/pi-apps
> git pull origin updater-manage
> ```

This test procedure exposed an oversight with the way this PR renames functions in `updater`. It appears we need to keep a wrapper function named `update_now_gui_apps`, otherwise updating any apps alongside the script changes in this PR will fail.

The old updater runs terminal-run, but terminal-run sources the new `updater` and then tries to run update_now_gui_apps, which no longer exists in this PR.

theofficialgman:
> The old updater runs terminal-run, but terminal-run sources the new `updater` and then tries to run update_now_gui_apps, which no longer exists in this PR.

Hmm... yes that is correct. The old update procedure updated all files first (`updatable_apps='' no_status=true update_now_cli`) and then updated the apps by calling `update_now_gui_apps`.

Now if we wanted to I could make it so that instead of this https://github.com/Botspot/pi-apps/pull/2340/commits/8135b939f2ad4f8aa1e0405089616e0bc22ec393 we have `update_now_gui_apps` wrap directly to `update_now_gui`. This should work because the terminal-run sources the new updater which inturn sources the new api. So all new functions would exist and the new manage-daemon would start working immediatly.

Personally though, I think I would rather have the deprecated original `update_now_gui_apps` function like you did.

theofficialgman:
rebase on master and squashed a couple of commits. not content changes

theofficialgman:
ok now it has been properly rebased and commits squashed (using git in github desktop on linux which didn't cause any trouble)

manage
```diff
@@ -51,19 +53,20 @@ validate_apps_gui() { #Given a list of actions and apps, graphically notify the
...
+    elif ([ "$action" == update ] || [ "$action" == refresh ]) && [ ! -d "${DIRECTORY}/update/pi-apps/apps/${app}" ];then
+      
       yad --class Pi-Apps --name "Pi Apps" --text="Invalid app "\""<b>$app</b>"\"". Cannot $action it." \
         --text-align=center --center --title='Error' --window-icon="${DIRECTORY}/icons/logo.png" \
         --button=OK!"${DIRECTORY}/icons/check.png":0
       
       #remove the app from list
       queue="$(echo "$queue" | sed ${i}d)"
-    elif [ "$action" != update ] && [ ! -d "${DIRECTORY}/apps/${app}" ];then
+    elif [ "$action" != update ] &&  [ "$action" != refresh ] && [ ! -d "${DIRECTORY}/apps/${app}" ];then
```
theofficialgman: no, this was correct before. a refresh by definition requires there to be a locally installed app. revert this part please

Botspot: I understand that was *your* definition, but I saw fit to expand the meaning to: "copying an app from update/ to apps/ when you don't want it to be reinstalled." It seems simpler this way.

manage
```diff
@@ -445,82 +447,91 @@ $app"
...
-      
-      #refresh the list in queue-viewer window
-      write_list "$queue"
-      
-      #echo "Action is $action, app is $app"
-      
+    #exit loop if queue is complete and no new actions were added to the queue
+    if [ "${current_line_num}" -gt "$(echo "$queue" | wc -l)" ];then
+      break
+    fi
```
theofficialgman, line 478: I personally prefer `if` statements that execute if true rather than break out of the loop. however I understand why you made this change in this case (to get rid of the else statement) so I will let it slide

Botspot:
Changes I made, in no particular order:
- skip sourcing api in manage script if using the `refresh` mode
- switched to the `refresh` copying a new app from the update folder to the main apps folder
- using a new process_exists function
- increased the height of manage daemon list viewing window, when not part of the dual-pane layout
- only reorder the queue if new items were added
- render the queue list as a background process; only render it again if the previous rendering has completed; this prevents the queue list from being a bottleneck when refreshing 100+ apps (yes I tested for this situation!) Now it will end up refreshing the list for roughly every 10 fast actions completed, on my system.
- render the "done" list item outside of the loop (code structure improvement)
- updater list now has lines like `refresh-app: $app` and `update-app: $app`, rather than `app: $app`. This expands the usage of refreshing an app from only compressed_update to any update operation that does not require a reinstall.
- Fix alphabetization of app updates so that it matches the order presented to user in `list_updates_gui`

manage
```diff
@@ -445,82 +447,91 @@ $app"
...
+    
+    #run manage script for app installs, uninstalls, or updates. Not for file-updates or app refreshes.
+    if [ "$action" == update ] && [[ "$app" =~ ^filelist:.* ]];then #list of files separated by ':'
+      # use the second argument to pass our list of files to update
+      updatable_apps='' updatable_files="$(echo "$app" | sed 's/^filelist://g' | tr ':' '\n')" no_status=true update_now_cli
+      exitcode=$?
+    else
+      echo manage
       "${DIRECTORY}/manage" "$action" "$app"
+      echo done
```
theofficialgman: looks to me like some debugging output is still enabled here

Botspot: I noticed and fixed this already.

manage
```diff
@@ -708,22 +719,30 @@ Or on Discord: \e[94m\e[4mhttps://discord.gg/RXSTvaUvuu\e[0m" | tee -a "$logfile
...
   #for this operation, a program name must be specified.
   app="$2"
   if [ -z "$app" ];then
     error "For this operation, you must specify which app to operate on."
   fi
   
-  # make sure update_app function is available
-  typeset -f update_app &>/dev/null || source "${DIRECTORY}/updater" source
+  #get functions from updater script
+  source "${DIRECTORY}/updater" source
```
theofficialgman: this seems to be in conflict with your changes above that made sure the `updater` was already sourced. what is the point of making sure it was sourced if you don't use it at all for `refresh` and source it anyway for `update` actions?

manage
theofficialgman: I see, you added functionality that depends on this in the next commit.

In the future it would be better to make this part of the change along with that next commit, rather than before. or note it.
https://github.com/Botspot/pi-apps/pull/2340/commits/50e4723fd826b7453f549118c53f31c74fdf096c#diff-196c23750e800642e7c008dc0416536449f295069c8161c23283c86334f8d251R323

manage
Botspot: Good catch. I had briefly tried having manage-daemon directly use updater' functions for both the `update` and `refresh` actions as a time-saving measure. I later decided against it as it created duplicate codepaths. I will remove the residual `$sourced_updater` code shortly.

updater
```diff
@@ -557,27 +558,24 @@ update_now_gui() { #input: updatable_files and updatable_apps variables
...
-    for app in $updatable_apps ;do
-      # if app is not a compressed update app, do a regular update
-      if ! echo "${refresh_apps}" | grep -q "^$app"; then
-        queue+="update $app"$'\n'
-      fi
-    done
+    queue+="$(echo "$updatable_apps" | list_subtract "$refreshable_apps" | sed 's/^/refresh /g')"$'\n'
   fi
   queue="${queue::-1}" #remove final newline character
+  queue="$(tac <<<"$queue")" # reverse order because manage daemon reverses it again; affects alphabetization, not priority
```
theofficialgman, line 575: Good catch. I did not consider the tac in the daemon (which was intented only reverse the order which the pending pipes were read) also reversed the file contents themselves.

manage
Botspot: Oh wait. It is necessary because updating files uses `update_now_cli`.

manage
```diff
@@ -20,7 +20,9 @@ if [ -z "$1" ];then
 fi
 
 set -a #make all functions in the api available to apps
-source "${DIRECTORY}/api" || error "failed to source ${DIRECTORY}/api"
+if [ "$1" != refresh ] || ! command -v generate_logo >/dev/null ;then
```
theofficialgman, line 23: Wouldn't it be better to check for `status` or `status_green` rather than `generate_logo`, a function that isn't used by `refresh`? Not that it really makes any difference

manage
```diff
@@ -708,22 +719,30 @@ Or on Discord: \e[94m\e[4mhttps://discord.gg/RXSTvaUvuu\e[0m" | tee -a "$logfile
...
+  #move old program to trash - this allows app-developers to recover their work if update was accidental
+  gio trash "${DIRECTORY}/apps/${app}" 2>/dev/null
+  #failsafe
+  rm -rf "${DIRECTORY}/apps/${app}"
+  
+  #copy new version from update/ to apps/
+  cp -rf "${DIRECTORY}/update/pi-apps/apps/${app}" "${DIRECTORY}/apps/${app}"
   
-elif [ "$1" == 'update' ] || [ "$1" == 'refresh' ];then
+  status_green Done
```
theofficialgman, line 726: this isn't very descriptive. is there a reason `status_green` was changed from
```
  if [ "$failed" == 'true' ]; then
    echo -e "\e[91mFailed to update ${app}.\e[0m"
    return 1
  else
    status_green "${app} was updated successfully."
    return 0
  fi
  ```
 ?

File: manage [hunk 21]
```diff
@@ -708,22 +719,30 @@ Or on Discord: \e[94m\e[4mhttps://discord.gg/RXSTvaUvuu\e[0m" | tee -a "$logfile
   #exit the manage script with the same exit-code of the app's script
   exit $exitcode
   
-elif [ "$1" == 'update' ] && [[ "$2" =~ ^filelist:.* ]];then #list of files separated by :
-  # use the second argument to pass our list of files to update
-  #get functions from updater script
-  source "${DIRECTORY}/updater" source
-
-  updatable_apps='' updatable_files="$(echo "$2" | sed 's/^filelist://g' | tr ':' '\n')" no_status=true update_now_cli
```
theofficialgman: Still not a huge fan of putting this inside of the daemon when all other functionality resides as its own manage action. Why do you want it inside of the daemon? There is no signifianct time penalty for calling manage anymore.

theofficialgman:
Lets let this cook for another night or two. 

Most of your changes are good or at least neutral but I do want to rewrite commit history for some things so additions make more sense (eg: https://github.com/Botspot/pi-apps/pull/2340#discussion_r1218798495)  and removals and re-additions don't happen. That can happen when it is all done though so don't worry about it.

updater
```diff
@@ -561,7 +558,7 @@ update_now_gui() { #input: updatable_files and updatable_apps variables
   
   #contruct a queue to send to terminal_manage_multi.
   if [ ! -z "$updatable_files" ]; then
-    queue+="update filelist:$(echo -n "$updatable_files" | tr '\n' ':')"$'\n'
+    queue+="$(echo "$updatable_files" | sed 's/^/update-file /g')"$'\n'
```
theofficialgman, line 561: same thing as before. updating one file with changes that are not compatible with the repo is not good practice even if your next commit completes the changes.  just make one commit next time

Botspot:
I have again changed the daemon list to list each updatable file on its own separate line at the queue level. I find it unnecessary to list them separately, then consolidate them on one line, split them apart for every list rendering, and finally split them apart when copying the files. Yes, this calls update_now_cli a bunch of times, but it has a negligible effect on speed, as `updater` is only sourced once.

updater
Botspot: Is there a way to commit to two files simultaneously on the github.com web interface?

Botspot: I would normally use `git col` but I'm basing off my local install which is based on `master` and I'd rather not switch branches.

manage
```diff
@@ -37,7 +37,7 @@ validate_apps_gui() { #Given a list of actions and apps, graphically notify the
   for line in $queue ;do
     action="$(echo "$line" | awk -F ';' '{print $1}')"
     
-    if [ "$action" != install ] && [ "$action" != uninstall ] && [ "$action" != update ] && [ "$action" != refresh ];then
+    if [ "$action" != install ] && [ "$action" != uninstall ] && [ "$action" != update ] && [ "$action" != refresh ] && [ "$action" != update-file ];then
```
theofficialgman, line 40: `terminal_manage` should be updated as well even if it is not used currently for `update-file` action (`terminal_manage_multi` is used with your change here https://github.com/Botspot/pi-apps/pull/2340/commits/e12ff09bbc4477b5846f76abbbd43a49e6161a19#diff-196c23750e800642e7c008dc0416536449f295069c8161c23283c86334f8d251R561 )

manage
Botspot: Because I consider `manage` to be a user-facing script for dealing with apps. 

I don't want our docs to say that manage can update files, but I also don't want it to be an undocumented feature that is accessible from command-line args.

manage
Botspot: Refreshing many apps at once is common. Since refreshing an app never generates any output, there is no need for large title blocks at the beginning and end. Rather, each refreshed app generates just one line in terminal: the status and the completion message at the end. (which I should improve to say "failed" if it fails)

manage
```diff
@@ -495,10 +488,9 @@ $app"
       write_list_pid=$!
     fi
     
-    #run manage script for app installs, uninstalls, or updates. Not for file-updates or app refreshes.
-    if [ "$action" == update ] && [[ "$app" =~ ^filelist:.* ]];then #list of files separated by ':'
-      # use the second argument to pass our list of files to update
-      updatable_apps='' updatable_files="$(echo "$app" | sed 's/^filelist://g' | tr ':' '\n')" no_status=true update_now_cli
+    #run manage script for app installs, uninstalls, or updates. Avoid using it for file-updates because that is out of the scope for manage script.
+    if [ "$action" == update-file ];then
+      updatable_apps='' updatable_files="$app" no_status=true update_now_cli
```
theofficialgman, line 493: we have come back full circle. now you have re-implemented the changes I didn't want. Updating one file at a time which duplicates functionality. the `.git` folder will get re-copied each and every update. this folder can get large which will slow down the process unnecessarily.
```
  #delete .git folder, then copy the new one
  rm -rf "${DIRECTORY}/.git" || sudo rm -rf "${DIRECTORY}/.git" || error "Failed to delete old ${DIRECTORY}/.git folder!"
  cp -a "${DIRECTORY}/update/pi-apps/.git" "${DIRECTORY}" || error "Failed to copy new .git folder!"
```

manage
Botspot: I wanted something that would have no chance of an overlap with an actual `status` command in /usr/bin.

updater
theofficialgman: > Is there a way to commit to two files simultaneously on the github.com web interface?

yup, for about 2 years now. dropdown next to edit, and open in github.dev. its basically vscode in the browser

manage
Botspot: Fair point here. But I see the filelist implementation being too clunky compared to its usefulness. Either I could adopt what you did to bundle all app refreshes to apply at once - to detect all update-file lines and push them all through update_now_cli.
Or, I could create a duplicate codepath to avoid using update_now_cli altogether.

manage
theofficialgman: `daemon` mode is not supposed to be user facing at all and thats a big part of the manage script. it is accessible from command line args. so I think you are already very far away from your ideal here.

I don't see how this is different, it just needlessly complicates it.

manage
theofficialgman: why not just split `update_now_cli` apart into multiple smaller functions

theofficialgman: basically this. just like update_app

```
update_file() {  #file name
  local file="$1"
  [ -z "$file" ] && error "update_file(): no file specified!"  
  mkdir -p "$(dirname "${DIRECTORY}/${file}")"
  
  #copy new version
  if cp -f "${DIRECTORY}/update/pi-apps/${file}" "${DIRECTORY}/${file}" ;then
    status_green "${file} file was copied successfully."
  else
    echo -e "\e[91mFailed to copy ${DIRECTORY}/update/pi-apps/${file}! \e[0m"
  fi
}

update_git() {
  #delete .git folder, then copy the new one
  rm -rf "${DIRECTORY}/.git" || sudo rm -rf "${DIRECTORY}/.git" || error "Failed to delete old ${DIRECTORY}/.git folder!"
  cp -a "${DIRECTORY}/update/pi-apps/.git" "${DIRECTORY}" || error "Failed to copy new .git folder!"
}

update_now_cli() { #input: updatable_files and updatable_apps variables
  
  local IFS=$'\n'
  for file in $updatable_files ;do
    update_file "$file"
  done
  
  for app in $updatable_apps ;do
    update_app "$app"
  done
  
  update_git
  
  [ "$no_status" != true ] && status_green "\nPi-Apps updates complete."
  true
}
```

then just call `update_file` directly in manage daemon and at the end call `update_git` if `update-file`, `update`, or `refresh` existed in the `queue`

manage
theofficialgman: thats why you use `typeset -f status &>/dev/null` and not `command -v`

updater
```diff
@@ -515,45 +516,53 @@ update_now_cli() { #input: updatable_files and updatable_apps variables
...
+      attempts=0
+      while [ ! -f "${DIRECTORY}/data/manage-daemon/pid" ] || [ ! -f "/proc/$(cat "${DIRECTORY}/data/manage-daemon/pid")/status" ];do
+        if [ "$attempts" -gt 500 ];then
+          break
+        fi
+        sleep 0.01
+        attempts=$((attempts + 1))
+      done
+    fi
+  }
```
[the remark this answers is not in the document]
theofficialgman: removed with https://github.com/Botspot/pi-apps/pull/2340/commits/4faf1266fa60a32b14c00ec3a203c12b51e528e3

theofficialgman:
> Changes I made, in no particular order:

@Botspot I fixed multiple bugs in your changes. the most eggregious of which was you incorrectly called `refresh` for apps that needed to have `update` called. I squashed the fix into the original commit for that one.

the other one I left here https://github.com/Botspot/pi-apps/pull/2340/commits/58e8ffdf928f868a01dba94c59ce0c7a334a4c74 but it will be squashed into the original commit as well by the time that you read this so you won't see it in the history.

manage
theofficialgman: resolved with https://github.com/Botspot/pi-apps/pull/2340/commits/29e05be3979c3203ea30d480577af23400d8cc06 no longer needed

manage
theofficialgman: I have come around to including it in the daemon now that they actually use the updater script functions
https://github.com/Botspot/pi-apps/pull/2340/commits/fa729751f8a515c8cbe407736149af4b4e7970d3

I have removed the refresh from manage as a result https://github.com/Botspot/pi-apps/pull/2340/commits/29e05be3979c3203ea30d480577af23400d8cc06

manage
```diff
@@ -476,14 +476,9 @@ $app"
...
-    if [ -z "$write_list_pid" ];then
-      write_list "$queue" &
-      write_list_pid=$!
-    elif ! process_exists "$write_list_pid" ;then
-      write_list "$queue" &
-      write_list_pid=$!
-    fi
+    #refresh the list in queue-viewer window as a background process
+    write_list "$queue" &
+    write_list_pid=$!
```
Botspot: The risk with this change is that when many apps are refreshed concurrently, it spawns a new write_list background process for every app - a thread bomb. My system slowed to a standstill when refreshing 30+ apps, and the yad list was blank and unresponsive until close to the end of the 100-app stress test.

theofficialgman: ideally we could kill any PID that is not the most recent one waiting.

I just ran my own simulation and it got through tons very quickly but then yad coredumped 😆

theofficialgman: for now I have dropped that commit untill a better solution can be made

Botspot:
During a long list of app refreshes, I kept getting errors saying that refresh_app was not a command. This was caused by you only sourcing updater in one of the if conditions for when new lines are added to the queue. For the first loop iteration the first if condition will be true, rather than the second, skipping the sourcing of updater.
I will fix this on my end as I work out a solution for the delayed list refreshing issue.

Botspot:
- I've added a secondary background process that will run if `write_list` is already running.
- secondary_write_list (a pseudoname - not a function name) waits for the first `write_list` process to complete, then refreshes the list with its own version of the list.
- secondary_write_list will be called every time write_list is busy, but older processes are killed so that only the newest secondary_write_list process is running at any one time.

I performed one test and it seemed to work. Briefly, after the app refreshes finished and while Cura was being reinstalled, the list was 10+ items behind. Then a few seconds into Cura's reinstallation, the list refreshed to show the current status.
2023-06-06-233332_1920x1080_scrot

Botspot:
I think this is ready to be merged. @theofficialgman, you mentioned something about not squashing commits during merge. Could you explain this further, specifically what difference it would make? Is leaving all commits individually more of a documentation decision or because you prefer to have authored a number of commits on `master` rather than one?

theofficialgman:
> During a long list of app refreshes, I kept getting errors saying that refresh_app was not a command. This was caused by you only sourcing updater in one of the if conditions for when new lines are added to the queue. For the first loop iteration the first if condition will be true, rather than the second, bypassing the sourcing of updater.

Ah yup my bad. Saw your fix, that was my intention.

Botspot:
It sure seems like this perfecting of commits is taking a long time. Not that it affects me, but I am growing uncertain if it should be worth your time.

theofficialgman:
> It sure seems like this perfecting of commits is taking a long time. Not that it affects me, but I am growing uncertain if it should be worth your time.

Don't worry about it. I find it fun and listen to music in the background anyway. Relatively large changes like this deserve clean commits anyway.

theofficialgman:
merged

Issue #807, opened 2021-07-16
Wine64


SuperPi911:
[Wine64.zip](https://github.com/Botspot/pi-apps/files/6830593/Wine64.zip)

SuperPi911:
Anyone?

Itai-Nelken:
problems with your scripts:
it uses `apt`. you should use pi-apps's `pkg-install`.
you are not handling errors.

both of those issues can be fixed if you use pi-apps's app creation wizard. you can find it in pi-apps settings.

Itai-Nelken:
sorry for taking so long to answer

SuperPi911:
i created the app with pi-apps settings and clicked "new app"

SuperPi911:
could you help me ?

Itai-Nelken:
> i created the app with pi-apps settings and clicked "new app"

you deleted all the code that already was there. take a look at the [app creation tutorial](https://github.com/Botspot/pi-apps/wiki/Creating-an-app)

> could you help me ?

maybe later, right now I don't have time.. sorry.

Tejas-Singh-py:
> could you help me ?

Okay, I am going to make the scripts, but you'll have to check it as I am not on 64-bit RaspiOS



> maybe later, right now I don't have time.. sorry.

@Itai-Nelken Can you make me a collaborator as well, or that power is in  @Botspot 's hand

Tejas-Singh-py:
Sorry I tried, but the script needs a complete rewrite. And also you cannot just install Wine you need to install it on top Box64 or maybe Box86, see my commit #794

SuperPi911:
You need box 64

SuperPi911:
https://user-images.githubusercontent.com/78208253/126446811-040cea2d-5b50-42f8-8bb5-2140af7eb55f.mp4

after you install Box64 you can install wine64 ONLY on 64BIT Raspbian         you can view in my video how it runs

Botspot:
Can Wine64 run 32-bit Windows apps? I'd be surprised if it **couldn't**.

Itai-Nelken:
I don't think it can. It calls wine32 for that I think.

Itai-Nelken:
@SuperPi911 sorry for the long time, we will eventually come to this pr.

SuperPi911:
[Wine64.zip](https://github.com/Botspot/pi-apps/files/7026805/Wine64.zip)


I think now it is finished!

Itai-Nelken:
[...]

Issue #2160, opened 2022-10-22
Some apps should/need to be updated


theofficialgman:
This is a meta issue for applications that we are aware of that need/should be updated due to considerable addition of features and/or bugfixes from upstream:

- AntimicroX
  - New releases have been made upstream with support for Gyro/motion controls but require that a very recent version of SDL2 be bundled with the application
  - maybe we should consider using the flatpak instead
- Stunt Rally
  - needs to be re-built against static libboost so that the deb can work on multiple distros
  - version 2.7 was released recently https://github.com/stuntrally/stuntrally/releases/tag/2.7

Rak1ta:
> Xsnow

The compiled version 3.5.3 works on Buster armhf, but I don't know which packages should be kept/removed as dependencies.

theofficialgman:
the dependencies for the armhf package are as follows
```
Depends: libc6 (>= 2.29), libcairo2 (>= 1.14.0), libgcc-s1 (>= 3.5), libgdk-pixbuf-2.0-0 (>= 2.22.0), libglib2.0-0 (>= 2.28.0), libgsl25 (>= 2.6), libgtk-3-0 (>= 3.9.10), libstdc++6 (>= 5), libx11-6, libxext6, libxpm4
```
libc6 is 2.28 on buster. other packages may also be too new. I have not checked the binary itself to see if it is compatible

theofficialgman:
turns out building OBS 28.X itself hasn't been too hard once I got QT6 installed. ffmpeg 4.X+ so far doesn't seem like a hard requirement. the problem comes when building the browser source CEF which now is built on the version where the ARM64 glibc reversion script was disabled in chromium. it STILL has not been re-enabled, I'm working on getting google chromium developers to do that upstream here: https://chromium-review.googlesource.com/c/chromium/src/+/3938615/2

theofficialgman:
update, the patch was merged at chromium a few weeks ago and it has made its way to chromium 111

everypizza1:
What about Godot? I think you can get 4.0 on the pi.

theofficialgman:
> What about Godot? I think you can get 4.0 on the pi.

[...]

PR #2341, opened 2023-05-30
Wine update to 8.11, new app Wine (x64), add ARM64 suport to Notepad++

theofficialgman:
closes https://github.com/Botspot/pi-apps/issues/807
works towards https://github.com/Botspot/pi-apps/issues/2160

built on ubuntu bionic 18.04 x86_64 with options `--enable-archs=i386,x86_64` to enable the new WoW64 mode (runs x86 32bit windows binaries with only x86_64 Linux userspace). WOW64 is currently non-functional in box64 https://github.com/ptitSeb/box64/issues/465 so only x64 applications (not x86) are expected to work.

~~extraneous features of the x86 wine install script were removed for testing of a minimal install. they can be added back (with the proper versioning) once deemed stable~~


apps/Notepad ++/install
```diff
@@ -0,0 +1,40 @@
...
+#Checking if wine is installed
+if ! command -v wine &>/dev/null ; then
+  echo 'Wine Is Not Installed Yet'
+  #Installing wine
+  if [ $arch == 32 ];then
+    "$DIRECTORY/manage" install 'Wine (x86)' || error 'Failed to install wine'
+  elif [ $arch == 64 ];then
+    "$DIRECTORY/manage" install 'Wine (x64)' || error 'Failed to install wine'
+  fi
+fi
```
Botspot, line 21: I suggest replacing this with `install-if-not-installed`. Otherwise, NPP would silently fail to launch on systems where the user already has wine installed from elsewhere. Chances are, if wine is installed from elsewhere, it is either an arm version from the repos, or a franken wine install from following a tutorial.

Botspot:
If this window appears during wine installation, it means that you did not place Mono in the correct system-wide location.
image
This dialog will not appear if you have a wine prefix with mono already in it.
This issue on winetricks repo should help provide context here. https://github.com/Winetricks/winetricks/issues/1236

theofficialgman:
> If this window appears during wine installation, it means that you did not place Mono in the correct system-wide location. image This dialog will not appear if you have a wine prefix with mono already in it. This issue on winetricks repo should help provide context here. [Winetricks/winetricks#1236](https://github.com/Winetricks/winetricks/issues/1236)

yup I am still debugging it. according to wine documentation it has been placed correctly and I have tried all permutations of what are listed in the documentation. this makes me think that systemwide mono is broken in recent wine versions

apps/Notepad ++/install
theofficialgman: this was copied from the install-32. this can be implemented

theofficialgman:
ah nevermind, /opt/wine-8.10/share/wine/mono works.... no clue why /opt/wine-8.10/mono is not working. according to docs it should.

apps/Notepad ++/install
theofficialgman: https://github.com/Botspot/pi-apps/pull/2341/commits/d29b3114c6b4d645a618eb0a84baa74897d6c569

theofficialgman:
also blocked by https://bugs.winehq.org/show_bug.cgi?id=55070 
This issue can happen on both x86 and x64 Wine but is hard to diagnose because not all OS/architecture/kernel combinations have the issue.

My Nintendo Switch on Ubuntu Bionic arm64 on kernel 4.9 has the issue but my Jetson Orin on Ubuntu Focal arm64 on kernel 5.10 does not have the issue.

My pi3 on Debian Bullseye armhf on kernel 6.1 has the issue.

theofficialgman:
see https://bugs.winehq.org/show_bug.cgi?id=55070 for a patch which I have no applied to the Wine-8.11 builds

theofficialgman:
winetricks is broken on wow64. currently working on fixing upstream winetricks

PR #2356, opened 2023-06-10
Wine (x86): update to Wine 8.10

theofficialgman:
works towards https://github.com/Botspot/pi-apps/issues/2160

built on ubuntu bionic 18.04 i386

independent of https://github.com/Botspot/pi-apps/pull/2341 but uses the same folder structure.

### DIFF
--- a/apps/Wine (x86)/credits
+++ b/apps/Wine (x86)/credits
@@ -1,1 +1,3 @@
-Script by Itai-Nelken and Botspot
+Wine Developers: https://wiki.winehq.org/Who%27s_Who
+Box86 Developer: @ptitSeb on GitHub
+Script Writer and Packager: @theofficialgman and other pi-apps contributors on GitHub
--- a/apps/Wine (x86)/description
+++ b/apps/Wine (x86)/description
@@ -1,4 +1,4 @@
-Run x86 Windows apps with a box86-emulated version of x86 wine.
+Run x86 Windows apps with a box86-emulated version of x86 wine-8.10.
 To run: wine path/to/file.exe
 To configure wine: go to 'Wine configuration' in main menu or type in terminal: wine winecfg
 Not all Windows applications will work under Wine. It's a good idea to check online if your program can run, or if something can be tweaked to make it work.
--- a/apps/Wine (x86)/install-32
+++ b/apps/Wine (x86)/install-32
@@ -295,6 +295,7 @@ Categories=System;" | sudo tee /usr/share/applications/wine-regenerate.desktop
 status_green "Done"
 
 cat << EOF | sudo tee /usr/local/bin/generate-wine-prefix >/dev/null
+#!/bin/bash
 echo
 
 #set up functions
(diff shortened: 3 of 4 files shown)


Botspot:
What is the goal behind placing wine in `/opt/wine-${version}` instead of the formerly more general `/opt/wine` directory? Do you intend for users to have multiple wine versions installed? Do you just consider it good practice?

theofficialgman:
> What is the goal behind placing wine in `/opt/wine-${version}` instead of the formerly more general `/opt/wine` directory? Do you intend for users to have multiple wine versions installed? Do you just consider it good practice?

this was an attempt at being closer to winehq official builds. they place builds in `/opt/wine-stable` `/opt/wine-devel` and `/opt/wine-staging` respectively. I thought it best we not just locate our builds in `/opt/wine` since it does not provide any description for what it is.

It could also allow a user to install mulitple versions easily though pi-apps will not do that.

theofficialgman:
I am considering changing the name to be even more specific thought. like `/opt/wine-amd64-i386-8.10` to indicate what  cpu support the wine package has. this way if we ever do decide to add wine x86 to 64bit installs (the version that requires 32bit userspace) then we can add a package called `/opt/wine-i386-8.10`

also this would allow us to ship ARM64 wine as well (since that does exist) under `/opt/wine-arm64-8.10`

theofficialgman:
also leaving this here as a reminder to myself and @Botspot we need to circle back around to fixing this as well before merging
https://github.com/Botspot/pi-apps/pull/2257#issuecomment-1411331286

theofficialgman:
the above commit "should" work on actual arm64/armhf kernels but in the CI it doesn't work since the host system's /boot folder doesn't get mounted into the chroot. so the first check fails and it tries to modprobe anyway.

not really much I can do about that since I don't really want to mount it. probably going to add a check to skip this test and assume the kernel is good when running the CI like we did in the api for nproc
https://github.com/Botspot/pi-apps/blob/24cd052ec5f0341f8eaae8a80c25dda75546ee31/api#L2883-L2896

theofficialgman:
ok I was not aware till now. QEMU redefines `uname` and actually make `uname -m` return the architecture of the emulated system (so in this case `armv7`) https://github.com/qemu/qemu/blob/master/linux-user/uname.c

thats nice I guess but not what I was banking on to make the CI "just work". looks like the check for GITHUB_ACTIONS will have to be added after all. the above cleanup and changes are still nice though

theofficialgman:
https://github.com/Botspot/pi-apps/actions/runs/5239470976/jobs/9459333705#step:3:7284

hmm... I think its just saying there is no file because it can't execute it. systemd-binfmt doesn't get restarted when box86 installs from within the chroot because it can't restart it from there
```
Restarting systemd-binfmt...
Running in chroot, ignoring command 'restart'
```
and since I didn't enable and install libc6:i386 on the host I assume the binfmt for i386 doesn't exist on the host.

theofficialgman:
blocked by https://bugs.winehq.org/show_bug.cgi?id=55070

theofficialgman:
closing in favor of single PR since both x86 and x64 are blocked by the same issue and will be merged at once when corrected upstream https://github.com/Botspot/pi-apps/pull/2341

PR #2370, opened 2023-06-27
Yad gui speed improvements

Botspot:
Avoids running a gui script subprocess every time a category or app is clicked. This has long been the behavior of the xlunch mode but I've had plans to extend it to yad. My past attempts failed but this time around it looks pretty solid. The interface is noticeably smoother and more snappy. I have not benchmarked it.
The other major benefit from these changes is that the app details window no longer vanishes when you try to drag it around the screen, resize it, or when you bring another window into focus. The details window will now close only when the user begins to do something else.

I have completed basic testing of the interface and have corrected all bugs found. It is entirely possible that edge cases have gone untested. While debugging, I found a lot of cases where a dialog (app list, details window, search) was supposed to close but no longer does, so I would not be too surprised if there are a couple more cases where that particular issue crops up.

I'm interested in feedback from @theofficialgman and maybe some testing from @Rak1ta if possible. (Rak1ta you seem to find a lot of bugs!)


theofficialgman:
I am currently out of town. Will take a look mid/late next week.

Rak1ta:
@Botspot i checked it all out. I noticed such a bug, I attach a video. If I write the text, you will not be able to understand me.

Until updated:
https://github.com/Botspot/pi-apps/assets/88277343/2a1df3c4-b728-451d-be91-c25bb64af84a

After the update:
https://github.com/Botspot/pi-apps/assets/88277343/3620da3e-6514-4d69-8d15-6625754fb863

Rak1ta:
Previously, an additional window was tied to the menu of the list of programs (although it was not taken into account that this could go outside the screen). Now as I understand it, you have made a reference to a certain place on the screen.
Is it possible to make it clear to all windows that they should be with the main menu? And taking into account the boundaries of the screen.

Безымянный
Безымянный2

Rak1ta:
I managed to make two windows, but it's hard to get hold of. I have an ssd, pi applications manage to open and close all windows quickly.
Безымянный3

Botspot:
The main issue uncovered by Rak1ta is that this implementation breaks how the `get_positions` function works. It needs the app list's window ID, provided by Yad's $YAD_XID variable that is sent to subprocesses. Now that yad is not starting each details window as a subprocess, the $YAD_XID variable remains unknown to the responses loop.

Botspot:
> Previously, an additional window was tied to the menu of the list of programs (although it was not taken into account that this could go outside the screen). Now as I understand it, you have made a reference to a certain place on the screen. Is it possible to make it clear to all windows that they should be with the main menu? And taking into account the boundaries of the screen.
> 
> I write about the fact that I cannot control the position of the pop-up menu, which can close other windows.

I have solved this issue. I am still looking for a way to prevent clicks from stacking up when clicking the Search button multiple times.

gui
```diff
@@ -320,67 +111,72 @@ guimode="$(cat "${DIRECTORY}/data/settings/App List Style")"
 #In Xlunch mode, the window must be centered, but xlunch only handles absolute offsets.
 get_positions() {
   #determine screen_width and screen_height
-  screen_dimensions="$(xrandr --nograb --current | awk -F 'connected |\\+|\\('  '/ connected.*[0-9]+x[0-9]+\+/ && $2 {printf $2 ", "}' | sed -n -e 's/^.*primary //p' | tr 'x+' ' ' | tr ',+' ' ')"
-  if [ -z "$screen_dimensions" ];then
-    # if screen_dimensions is empty, this could be a single monitor wayland display which does not have the word "primary" in the output
-    # workaround is to get the first output returned for the connected display
-    screen_dimensions="$(xrandr --nograb --current | awk -F 'connected |\\+|\\('  '/ connected.*[0-9]+x[0-9]+\+/ && $2 {printf $2 ", "}' | tr 'x+' ' ' | tr ',+' ' ')"
+  if [ -z "$screen_width" ] || [ -z "$screen_height" ];then
```
theofficialgman, line 114: I believe this will cause the UI to appear incorrectly if the screen resolution changes after pi-apps is first opened or if the user moves the GUI to another monitor. This is because the screen_width and screen_height variables will be set after the first call of get_positions and will not be updated again. This is a very fast function to run and there isn't any measureable benefit to not running dimensions/width/height every time

both of these were not issues

gui
```diff
@@ -111,67 +111,72 @@ guimode="$(cat "${DIRECTORY}/data/settings/App List Style")"
...
-      # this is the size of the window borders (including title bar)
-      border_widths=($(xprop _NET_FRAME_EXTENTS -id "$main_yad_window" | sed -n  's/_NET_FRAME_EXTENTS.*=//p' | tr ',' '\n'))
-      #total dimensions for both yad windows side by side
-      if [ $screen_width -le 1000 ] || [ $screen_height -le 600 ];then
-        #gui size for small screens
-        height=${main_yad_window_dim[1]}
-        width=600
+      
+      #skip calculating offsets if yad window is in the same place as last time
+      if [ -z "$last_window_info" ] || [ -z "$geometry1" ] || [ -z "$geometry2" ] || [ "$last_window_info" != "$window_info" ];then
```
theofficialgman, line 135: same issue as above. best to just drop this commit unless there was actually some functional issues that were fixed.

theofficialgman:
@Botspot the major issue I forsee with this has to do with the update to this change. I have not tested these changes but I know that on update, the old GUI script will be running. Clicking on any app/settings/etc will call the gui script with the 3rd argument which functionality you have removed. this will either result in the wrong thing executing or nothing.

gui
```diff
@@ -502,49 +458,110 @@ if [[ "$guimode" == yad* ]];then
...
+      "${DIRECTORY}/preload" yad "$prefix" > "$pipe" 2>/dev/null
+    fi
+  }
+  
+  kill_details() { #close the details window in 1 second. Uses the $detailspid variable. To close it immediately add an argument of 0
+    (
+    if [ "$1" != 0 ];then
+      sleep 1.5
+    fi
+    pkill -P "$detailspid" 2>/dev/null
```
theofficialgman, line 486: this kills child PIDs of the parent process (the details window)
this (I think) will kill any running or backgrounded `terminal_manage` calls spawned from that detailspid

Botspot:
@theofficialgman I think it was unintentional to push a snapdrop commit to this branch.

theofficialgman:
> @theofficialgman I think it was unintentional to push a snapdrop commit to this branch.

what snapdrop commit 😆 . yes it was unintentional. thought I had master checked out

gui
Botspot: I had not considered this possibility, but it does not seem to kill the terminal in practice. Not sure why not.

gui
Botspot: I made these changes to get_positions to see how much of a difference it would make when rapidly selecting between apps, and the difference was more noticeable when there was high load on the system, like a program being compiled, a video playing, or a sudo apt upgrade.
Otherwise the difference is not really noticeable.

gui
Botspot: Screen_width is already only really used for the initial launch to choose one of two window sizing options.
Before I made these changes, if the screen resolution changed after launch, the main yad window would not receive the sizing update but the details window would now appear in "small mode". That is a situation we have not really planned for, but nobody has reported such a situation, so I doubt it happens often.

If we really wanted to account for changing screen resolutions, it would take more edits than reverting this commit. Either the app list would need to be closed and relaunched with new dimensions, or something would need to resize it on-the-fly.

Botspot:
> @Botspot the major issue I forsee with this has to do with the update to this change. I have not tested these changes but I know that on update, the old GUI script will be running. Clicking on any app/settings/etc will call the gui script with the 3rd argument which functionality you have removed. this will either result in the wrong thing executing or nothing happening at all.
> 
> If we want to keep these changes, I think we will need to bump `GUI_FORMAT_VERSION` to 3 and provide a simple backwards compatibility layer if `GUI_FORMAT_VERSION` is currently set to <=2 (from the previously running GUI)

I just tried simulating an update from old to new. When I clicked on a category in the old gui, the new gui launched, covering the old with an identical app list. From there it worked fine. I'm inclined to leave it like this - it seems better than having `updater` kill the gui the way it did for GUI_FORMAT_VERSION=2.

Botspot:
After thinking about it more, I went ahead with GUI_FORMAT_VERSION=3.

theofficialgman:
> After thinking about it more, I went ahead with GUI_FORMAT_VERSION=3.

hmm... the old GUI did not close. something changed to make this runonce not affective anymore

theofficialgman:
I found a way to kill the yad windows (by using the knowledge we have of what arguments it was started with to get the pid from ps). this kills both the old gui script and the old yad window, however the new yad window pops up in the top left with a strange size and it closes after 30 seconds (the timer from the terminal-manage script that I guess the new gui got spawned as a child PID from)

theofficialgman:
The benifits of these changes are clear. However I don't think there is a way to integrate them without asking the user to re-start pi-apps. We can do that easily by adding an if statement near the top of the GUI script that checks the `GUI_FORMAT_VERSION` that is currently set (so from the old gui it will be set to 2) and prompt the user to close and re-start pi-apps.

I tried `setsid` as well which makes the process a child of init. But the problem there is because it is a child of init, it struggles to get the window size and doesn't seem to be able to launch its own terminal-run processes (so installing/uninstalling/etc all fail)

theofficialgman:
also I'm going to be honest I don't see any speed differences with this PR compared to master. If you see any that must mean you are pretty IO limited (on your storage). Since the only difference is this PR has all of the functions already in bash, while master has to read from storage (or a cache) and execute the gui script again.

theofficialgman:
I've tried every iteration of forking (&), disown, nohup, setsid. everything either dies when the timer ends (still a child process) or fails to run terminal-run (like mentioned before)

gui
theofficialgman: I think in practice this is impossible to hit because in order to kill the detailspid that has the manage child process, you would need to click uninstall/install and another app name in the list at the exact same time. Since the partent process (the details window) pid disappears once the yad details window closes on clicking the install/uninstall button, the kill_details pkill does nothing since that pid does not exist (even though it has child PIDs).

theofficialgman:
outside of the outright removal of the mange, api, settings, or other fundamental scripts I see no reason why we would ever need to close or restart the GUI on the user ever again from now on. Since the GUI is a self contained running script now (it does not call itself) an update can not cause a break in functionality.

gui
```diff
@@ -502,49 +458,108 @@ if [[ "$guimode" == yad* ]];then
...
+      echo -e "\f" > "$pipe"
+      "${DIRECTORY}/preload" yad "$prefix" > "$pipe" 2>/dev/null
+    fi
+  }
+  
+  kill_details() { #close the details window in 1 second. Uses the $detailspid variable. To close it immediately add an argument of 0
+    (
+    if [ "$1" != 0 ];then
+      sleep 1.5
+    fi
```
theofficialgman, line 485: what is the purpose of the sleep? is there a reason why you want to wait to kill the details window when the user clicks on something else?

Botspot: The purpose of this delay is to improve the UX. Previously, when the app details window is open and you select a different app, the first details window will immediately disappear, then about a second later, the new details window appears. If the background behind pi-apps is a dark color, this can cause the overall screen brightness to vary drastically. By aiming to close the old details window after the new one is in place, it makes the experience much more seamless feeling - more like a cohesive UI and less like hacked together dialogs that appear and disappear erratically.

theofficialgman:
> I have solved this issue. I am still looking for a way to prevent clicks from stacking up when clicking the Search button multiple times.

@Botspot solved with https://github.com/Botspot/pi-apps/pull/2370/commits/37fe8aa9c4a564adb58cf9efe925161187c702d0

gui
Botspot, line 1: This is genius.

Botspot:
> > I am still looking for a way to prevent clicks from stacking up when clicking the Search button multiple times.
> 
> @Botspot solved with [37fe8aa](https://github.com/Botspot/pi-apps/commit/37fe8aa9c4a564adb58cf9efe925161187c702d0)

Well done. That is quite a genius way of doing it.

Botspot:
I found a better way to transition to the new gui, hold on...

gui
theofficialgman: > The purpose of this delay is to improve the UX. Previously, when the app details window is open and you select a different app, the first details window will immediately disappear, then about a second later, the new details window appears. If the background behind pi-apps is a dark color but the system GTK theme is light, this can cause the overall screen brightness to vary drastically under normal usage. By aiming to close the old details window after the new one is in place, it makes the experience much more seamless feeling - more like a cohesive UI and less like hacked together dialogs that appear and disappear erratically.

that is because your SD card is slow. on my system with a decent SD card (sandisk A2 512GB, not even the best which is the samsung A2 pro plus), the new details window pops up almost immediately (~ 0.5 seconds) and then the old details window closes afterwards.

I can see how this is a clever way of making sure the background never shows through even if it is weird for me to see the old window close 1.5 seconds later (like if I move it before opening a new one)

gui
theofficialgman: can't find any issue with it

theofficialgman:
> Well done. That is quite a genius way of doing it.

took me a while to figure out how to do it without a subshell. not going to lie. 😆

gui
Botspot: Fascinating, because I'm booting from almost the fastest NVME drive money can buy, with a USB 3 enclosure. I've always been used to it taking 1-1.5 seconds for just about any type of yad window to appear.
I've long wonder if yad is poorly optimized, or if it waits for something to time out in the background.

Botspot: It could also be Raspbian's Mutter window manager to blame, with its open/close animations.

theofficialgman: Ultimately it is probably a combination of things.

Mutter is a compositing window manager (uses OpenGL). The pi has a pretty slow gpu and memory.
I use mutter as well (in budgie) but ofc that is with a gpu with a LOT more grunt than the pi.

The pi's ram is very slow (4-5GB/s) and higher latency in comparison to the TX1 (24GB/s).
The pi's very low cpu cache (compared to other chips also with 4 A72 cores) also really hurts its multithreading performance.

yad is also probably poorly optimized. I haven't read its code much (besides to conclude that newer versions are just broken).

theofficialgman: Also our background shellcheck that runs for 10-15 seconds at 100% of one thread surely doesn't help either 😆

theofficialgman:
found a bug (not to do with the latest commit)
having the update window open prevents clicking anywhere else in the GUI because it also isn't a background task and so the inputs stack

Botspot:
> found a bug (not to do with the latest commit) having the update window open prevents clicking anywhere else in the GUI because it also isn't a background task and so the inputs stack

Is the `--close-on-unfocus` flag not working on your system?
I added this flag to updater so that clicking anywhere else will close the updater window.

theofficialgman:
> Is the `--close-on-unfocus` flag not working on your system? I added this flag to updater so that clicking anywhere else will close the updater window.

it works fine. the problem is the window isn't automatically focused. the focus is still on the main window when the updater opens. so if I clicks in the main window again I never focused the updater in the first place so it doesn't close.

It is actually really easy for me to tell what window is in focus because the top border changes color slightly. the details window focuses immediately.

theofficialgman:
I'm not actually sure why this is the case or what causes yad to focus one window but not another.
A simple test script works fine with the initial focus being on the first yad window, and then the focus changing to the second yad window
`yad & sleep 3; yad --close-on-unfocus`

theofficialgman:
interestingly the updater prior to this PR did focus automatically....

gui
```diff
@@ -502,49 +458,111 @@ if [[ "$guimode" == yad* ]];then
...
+  done < <(yad "${yadflags[@]}" \
+    --image="${DIRECTORY}/icons/logo-64.png" --image-on-top --text "$motd" \
+    --list --no-headers --column=:IMG  --column=Name --column=Sysname:HD --column=tip:HD --column=@fore@:HD \
+    --separator='\n' --window-icon="${DIRECTORY}/icons/logo.png" \
+    --tooltip-column=4 \
+    --dclick-action "bash -c "\""return_sysname %s"\""" --select-action="bash -c "\""return_sysname %s"\""" \
+    --button="!${DIRECTORY}/icons/search.png"!'Search':"bash -c "\""return_sysname '' '' Search/"\""" \
+    --button="!${DIRECTORY}/icons/options.png"!'Settings':"bash -c "\""echo Settings/ ; kill "\$"YAD_PID"\""" \
+    "$geometry1" < <(tail -f --retry $pipe))
```
theofficialgman, line 551: you removed this https://github.com/Botspot/pi-apps/commit/ac8ee8a6505ab2a7cdc85341bca4d21ca6416f74 workaround

I image it is still necessary and will have to check. Its purpose was to kill the tail command when the yad dialog closed.

theofficialgman: hmm... actually this method seems to avoid the issue entirely. interesting.

theofficialgman:
> interestingly the updater prior to this PR did focus automatically....

same deal on ubuntu focal with default gnome DE (mutter)

gui
theofficialgman: added it back with https://github.com/Botspot/pi-apps/pull/2370/commits/8efbbe90534d9d0fe12f6c047374b41bfce10d89

theofficialgman:
yeah so none of the windows are being focused that spawn in that while loop
the only reason that the details window is focused is because of this
```
    #Make sure app details window is focused (necessary for openbox window manager)
    (i=0
      while [ "$(xdotool getactivewindow getwindowname)" != "Details of ${app}" ] && [ $i -lt 20 ]; do
        wmctrl -a "Details of ${app}" #Raise the window named "Details of ${app}"
        sleep 0.2
        i=$((i+1))
      done) &
```

before this PR I could remove that snippit above and the details and all other windows were focused when they were opened... something about how yad is being called from the same process probably or something

Botspot:
On your system, could you check if a yad window when run with `--close-on-unfocus` will launch unfocused?

theofficialgman:
> On your system, could you check if a yad window when run with `--close-on-unfocus` will launch unfocused?

it does not launch unfocused.

I believe the root issue is that the main yad window now is never inactive. You have its output redirected to the while loop and that must place it in a different state then before where the button actions broke out into case statements.

Going to try some stuff locally but it might have to wait till tomorrow.

theofficialgman:
Took a WHILE but I figured out the issue stems from the how yad is calling the commands. yad was directly calling another yad window before this PR. now the gui script is calling the secondary yad commands.

yad doesn't like that and won't raise if you do that.

the while loop is unnecessary, so I got rid of it :)

theofficialgman:
ah crap this breaks killing the windows ofc... 
will look into this tomorrow or later

Botspot:
> ah crap this breaks killing the windows ofc... will look into this tomorrow or later. probably going to have to change the detailspid to a pipe so that anything can read it.

It also introduces possible issues if multiple action threads are running concurrently. A while loop has more control over what is being run, and when, and can easily pass information forward from previous iterations. I'm hoping you can find another way to fix the window focusing issue.
Have you tried experimenting with different `--class` and `--name` values?

theofficialgman:
> It also introduces possible issues if multiple action threads are running concurrently. A while loop has more control over what is being run, and when, and can easily pass information forward from previous iterations. I'm hoping you can find another way to fix the window focusing issue that only seems to be happening on your system. Have you tried experimenting with different `--class` and `--name` values? Have you tried a more recent version of yad?

Two systems. Different hardware. Different DEs. Different OS. Make sure to test yourself from the . desktop file since it doesn't happen when running pi-apps/gui from terminal.

theofficialgman:
@Botspot before the PR, before https://github.com/Botspot/pi-apps/pull/2370/commits/dc1fa200935337ea020d22f5e8c8911de31c72b2 , and after on piOS bullseye with the default DE (64bit on a pi3), the updater window is not focused on launch. So this DE does not work regardless of what we do and always has this issue https://github.com/Botspot/pi-apps/pull/2370#issuecomment-1624589613

theofficialgman:
> It also introduces possible issues if multiple action threads are running concurrently.

the gui functioned this way before the PR and there were no issues.

theofficialgman:
@Botspot I have tested more systems. in summary

has the issue (updater not focused on launch) before the PR:
piOS with default DE

has the issue (updater not focused on launch) with the PR by before my fix (https://github.com/Botspot/pi-apps/commit/dc1fa200935337ea020d22f5e8c8911de31c72b2):
piOS with default DE
Ubuntu Bionic 18.04 (xorg) with mutter on Budgie DE
Ubuntu Focal 20.04 (xorg) with mutter on Ubuntu/Gnome DE
Ubuntu Lunar 23.04 (xorg) with kwin on KDE Plasma

has the issue (updater not focused on launch) with the PR but after my fix (https://github.com/Botspot/pi-apps/commit/dc1fa200935337ea020d22f5e8c8911de31c72b2):
piOS with default DE

So my fix corrects all DEs (3) that have the bug as a result of this change and does nothing for the broken DEs (piOS) that never properly raised the window in the first place

theofficialgman:
> Have you tried experimenting with different `--class` and `--name` values?

no affect.

theofficialgman:
> ah crap this breaks killing the windows ofc... will look into this tomorrow or later. probably going to have to change the detailspid to a pipe so that anything can read it.

solved with https://github.com/Botspot/pi-apps/pull/2370/commits/212e53938b561d28ce0e4afe094e4a61165c25a5

theofficialgman:
@Botspot I really would have preferred that do a rebase. It's impossible to have issues if you do once since got forced you to resolve conflicts manually as you do one.

Botspot:
> @Botspot I really would have preferred that do a rebase. It's impossible to have issues if you do one since you get forced to resolve conflicts manually during the process. Using github actions with vscode as the editor its a breeze to resolve any

I actually was forced to resolve conflicts, but I did it incorrectly. I should have tested gui after resolving the conflicts, not before.

theofficialgman:
fyi, the github actions should have been an indication for you not to merge
shellcheck failed

you are welcome to add a rule to prevent merging unless all checks pass. I can't do that since I don't have permissions on this repo to edit settings. https://stackoverflow.com/questions/60776412/github-actions-is-there-a-way-to-make-it-mandatory-for-pull-request-to-merge

PR #2373, opened 2023-06-30
add box64'd discord release and ptb

libewa:


Botspot:
1. What situations would a Pi user need or want the emulated discord client when ArmCord is basically identical, but has options for themes and better security settings?
2. Not sure what you did here, but there is a lot of duplicated file structure. Take a look at https://github.com/Botspot/pi-apps/pull/2373/files. There appears to be three duplicates of everything, under different folder names.

theofficialgman:
I personally don't see box64'd chromium to be very viable. For one, the required ram to emulate chromium multiplies by a factor of 2-4X which is not suitable for background use as discord is intended. Performance is also lower by 2-4X compared to native.

libewa:
ok, some answers:
> What situations would a Pi user need or want the emulated discord client?

Rich Presence, as in using Riitag RPC etc without booting your PC

> Not sure what you did here, but there is a lot of duplicated file structure. There appears to be three duplicates of everything, under different folder names.

That Discord and Discord PTB are (almost) identical. For some reason though, the Discord PTB folder ... contains another copy of the Discord folder.

> What does PTB refer to?

Discord PTB is the Public Test Build (aka beta) of the Discord client

theofficialgman:
> Rich Presence, as in using Riitag RPC etc without booting your PC

I suggest using an alterative implementation independent of discord, seen here https://github.com/Botspot/pi-apps/pull/2377#issuecomment-1617211854

It really isn't viable or worth running the official client through emulation.

libewa:
at least to me it seems as if armcord is somehow ... slower

Botspot:
> at least to me it seems as if armcord is somehow ... slower

Slower than an emulated chrome binary?

libewa:
ihnc how but yes

Issue #2350, opened 2023-06-03
Download/Install script for Cascaydia Cove NF


libewa:
### What is the name of the app?

Cascaydia Cove NF

### (Optional) Where is the app hosted?

https://www.nerdfonts.com/

### About the app

Cascaydia Cove is a version of Microsoft's Cascadia Code font with added symbols for apps like Oh My Posh and Oh My Fish.

### Upload file or Add PR Link

~~[Cascaydia Cove NF.zip](https://github.com/Botspot/pi-apps/files/11642018/Cascaydia.Cove.NF.zip)~~

PR #2375 

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.
- [X] I have confirmed that this app can run on Raspberry Pi.

libewa:
[Cascaydia Cove NF.zip](https://github.com/Botspot/pi-apps/files/11642090/Cascaydia.Cove.NF.zip)
fixed a typo

libewa:
i should stop making typos
[Cascaydia Cove NF.zip](https://github.com/Botspot/pi-apps/files/11642109/Cascaydia.Cove.NF.zip)

theofficialgman:
there is an edit button under the `...` in the top right of each comment
no need to make a new comment with changes

PR #2375, opened 2023-07-01
Caskaydia Cove NF font

libewa:
closes https://github.com/Botspot/pi-apps/issues/2350


apps/Caskaydia Cove NF/install
```diff
@@ -0,0 +1,14 @@
...
+mkdir cascadia
+cd cascadia
+wget https://github.com/ryanoasis/nerd-fonts/releases/download/v3.0.1/CascadiaCode.zip || error "failed to download font"
+unzip CascadiaCode.zip || error "failed to extract zip"
+mkdir ~/.fonts
+mkdir ~/.fonts/Caskaydia
+rm -- *Propo*
+rm -- *Mono*
+rm CascadiaCode.zip
+mv * ~/.fonts/Caskaydia/
```
theofficialgman, line 12: this only applies to the current user. when possible it is best to save files to the global directories so that they apply for all users. that would be `/usr/share/fonts` (for system installed fonts) and `/usr/local/share/fonts` for user installed fonts (the latter of which you should use)

also, why are specific fonts being removed after downloading? are they not desireable for some reason? maybe you should take that up with upstream ryanoasis if you do not like them.

apps/Caskaydia Cove NF/uninstall
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+rm -r ~/.fonts/Casaydia
```
theofficialgman, line 3: see above

libewa: there's also a typo there lol

apps/Caskaydia Cove NF/uninstall
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-rm -r ~/.fonts/Casaydia
+sudo rm -r /usr/local/share/fonts/Caskaydia
```
theofficialgman, line 3: -rf so that a failure due to the folder already not existing does not cause the script to exit with an error

apps/Caskaydia Cove NF/install
```diff
@@ -1,13 +1,14 @@
...
 
+cd /tmp
 mkdir cascadia
 cd cascadia
 wget https://github.com/ryanoasis/nerd-fonts/releases/download/v3.0.1/CascadiaCode.zip || error "failed to download font"
 unzip CascadiaCode.zip || error "failed to extract zip"
-mkdir ~/.fonts
-mkdir ~/.fonts/Caskaydia
+sudo mkdir /usr/local/share/fonts
+sudo mkdir /usr/local/share/fonts/Caskaydia
```
theofficialgman, line 9: use `sudo mkdir -p /usr/local/share/fonts/Caskaydia` instead of both of these lines. it will create subfolders as necessary

apps/Caskaydia Cove NF/install
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+cd /tmp
+mkdir cascadia
+cd cascadia
+wget https://github.com/ryanoasis/nerd-fonts/releases/download/v3.0.1/CascadiaCode.zip || error "failed to download font"
+unzip CascadiaCode.zip || error "failed to extract zip"
+sudo mkdir /usr/local/share/fonts
+sudo mkdir /usr/local/share/fonts/Caskaydia
+rm -- *Propo*
```
theofficialgman, line 10: I still think all fonts should be left. no reason to remove any since they just take up a bit more space.

apps/Caskaydia Cove NF/description
```diff
@@ -0,0 +1,2 @@
+A fun, new monospaced font that includes programming ligatures and is designed to enhance the modern look and feel of the Windows Terminal.
+This font will be installed in/usr/local/share/fonts and will be available in most programs on next restart.
```
theofficialgman, line 2: missing space `in/usr/local/share/fonts`

theofficialgman:
a bit more context
https://github.com/Botspot/pi-apps/commit/13121322a7bdf23e79c24b2b654e4e09591b8b39#r120626265
please place it in `/usr/local/share/fonts/truetype/Caskaydia` since these are truetype fonts

Issue #2346, opened 2023-06-01
Oh My Posh install script


libewa:
### What is the name of the app?

Oh My Posh

### (Optional) Where is the app hosted?

https://github.com/jandedobbeleer/oh-my-posh
I download the official install script, which detects the architecture, from https://ohmyposh.dev.

### About the app

Oh My Posh adds customizable prompts to many shells, including bash, zsh, pwsh and fish.
Prompts are configured via json files and activated by adding a command to `~/.bashrc`

### Upload file or Add PR Link

~~[Oh My Posh.zip](https://github.com/Botspot/pi-apps/files/11627915/Oh.My.Posh.zip)~~
PR #2376

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi Apps app list.
- [X] I have confirmed that this app can run on Raspberry Pi.

theofficialgman:
multiple issues with your install and uninstall scripts

I suggest you take a look at what the `https://ohmyposh.dev/install.sh` script actually does and integrate this into the install script in pi-apps itself. there is no reason to run their wrapper script which has no error checking. all it does is download the latest release binary and themes from the github releases and add them to the proper installation directory.

you should not leave orphaned files on the users install or execute extraneous commands either (eg: don't run purge_packages if you never ran install_packages)

also, please post a response with a picture of what your terminal looks like before and after installing this. I am having trouble understanding what it is for, what it does, and why you want it from the description and the website.

libewa:
i also went on the omp discord for instructions on how to uninstall, and jan said to not use the install script on linux, but to use go install

libewa:
[Oh My Posh.zip](https://github.com/Botspot/pi-apps/files/11635863/Oh.My.Posh.zip)
entirely new script using go install

theofficialgman:
please refer to https://github.com/Botspot/pi-apps/issues/2346#issuecomment-1572636375
I do not want go install. pi-apps should manage the version of "oh my posh" by installing from github releases just like the https://ohmyposh.dev/install.sh does

also I still need the the last request there answered.

libewa:
ah yeah

libewa:
Screenshot: 
![image](https://github.com/Botspot/pi-apps/assets/67926131/c349708a-814c-4b01-abf6-2f9455c94731)
[...]

PR #2376, opened 2023-07-01
Oh My Posh

libewa:
closes https://github.com/Botspot/pi-apps/issues/2346


apps/Oh My Posh/install-32
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+wget -O /tmp/oh-my-posh https://github.com/JanDeDobbeleer/oh-my-posh/releases/latest/download/posh-linux-arm
```
theofficialgman, line 3: I would still like this to be versioned (use the format like shown in the example install script from the documentation https://pi-apps.io/wiki/development/Creating-an-app/#making-an-install-script )

You should additionally add an auto-updater so that pi-apps can keep the users install of Oh My Posh up to date. I don't have great documentation written for how to do so but this commit should be a good start showing how to obtain the version number from github releases and how pi-apps expects to be able to update the version number of the application automatically https://github.com/Botspot/pi-apps/commit/fd584580618f627f6508f686a5a01b1b7cc67cfc

apps/Oh My Posh/install-64
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+wget -O /tmp/oh-my-posh https://github.com/JanDeDobbeleer/oh-my-posh/releases/latest/download/posh-linux-arm64
```
theofficialgman, line 3: see above

etc/categories
```diff
@@ -46,6 +46,8 @@ Deluge|Internet
 Descent 1|Games
 Descent 2|Games
 Deskreen|Internet
+Discord|Internet/Communication
+Discord PTB|Internet/Communication
```
theofficialgman, line 50: applications are added that are not part of this PR and Oh My Posh is not added in the categories file

libewa: rip switched up branches

etc/categories
```diff
@@ -129,6 +129,7 @@ Node.js|Tools
 Notejot|Office
 Notepad ++|Programming
 OBS Studio|Multimedia
+Oh My Posh|Terminals/Appearance
```
theofficialgman, line 132: not a valid category

Issue #2387, opened 2023-07-11
Pi Apps moved to preferences?


steveiliop56:
Hello,

Today I formated my raspberry pi and then installed pi-apps with these commands:

```Shellscript
git clone https://github.com/Botspot/pi-apps
mv pi-apps/ .pi-apps/
~/.pi-apps/install
```

While the install worked the pi-apps app appeared in the Preferences section rather than the Accessories one. I tried uninstalling and reinstalling 2 times and I had the same result. So did pi-apps change location?

Botspot:
Hello @steveiliop56, yes this is a known issue.
It was a mistake caused by @theofficialgman's recent commit https://github.com/Botspot/pi-apps/commit/7e381a2ef769e56873197b7683972e7c6ea312da
I've asked him to fix it in a way that it solves the problem he was having, while avoiding moving the pi-apps launcher to Preferences on PiOS.

steveiliop56:
So it's a bug and it will be fixed in the next update?

Botspot:
> So it's a bug and it will be fixed in the next update?

It's a bug and will be fixed soon. If @theofficialgman does not get to it today, then I will fix it myself.

theofficialgman:
Will look into it this afternoon local time.

theofficialgman:
piOS uses the LXDE desktop environment by default. the default categories are controlled by the XML found in `/etc/xdg/menus/lxde-pi-applications.menu`

read the below contents you will find that the accessories submenu is generated based on applications that have a category that contains `Utility and NOT (Accessibility OR System)`

after the help menu there is a `DesktopSettings` menu which I believe is the `Preferences` menu (it is strange because in the XML it is named differently. This menu is generated last and contains only categories `(Settings OR PackageManager OR System)` that were not already allocated above.

Since pi-apps currently uses `Categories=Utility;System;PackageManager;` it failed the Accessories submenu checks so DesktopSettings` picked it up due to System and PackageManager existing.

kde discover: `Categories=Qt;KDE;System;`
gnome software: `Categories=GNOME;GTK;System;PackageManager;`
synaptic package manager: `Categories=PackageManager;GTK;System;Settings;`
snap store: `Categories=GNOME;GTK;System;PackageManager;`

[...]

PR #2388, opened 2023-07-11
add note that pi-apps.desktop file location in applications lists has been corrected

theofficialgman:
closes https://github.com/Botspot/pi-apps/issues/2387


Botspot:
> I am content on leaving it how it is now @Botspot since it is show it always should have been in the first place.

It is not your place to make this call.
I will make a compromise: if on PiOS, the launcher will be returned to Accessories (for the same reason why PiKISS GUI and PiGro are not in Preferences), but if the OS is not PiOS it will go in Preferences.

theofficialgman:
> It is not your place to make this call. If this was intentional to begin with, it should have been a PR from the start. 

I already said it was not intentional. hence why it was never a PR in the first place https://github.com/Botspot/pi-apps/commit/7e381a2ef769e56873197b7683972e7c6ea312da#commitcomment-121236764

Issue #2395, opened 2023-07-20
ArmCord - Version Mismatch


TheAlienDrew:
### Confirmations

- [X] I searched for similar bug reports in https://github.com/Botspot/pi-apps/issues and found none were relevant.
- [X] I tried updating Pi Apps and the issue is still not fixed. [How to update Pi Apps?](https://github.com/Botspot/pi-apps#to-update-pi-apps)
- [X] IMPORTANT: I promise to respond to comments on this issue and write a good bug report https://www.softwaretestinghelp.com/how-to-write-good-bug-report/

### What happened?

After updating ArmCord to the latest version, for some strange reason that shows an older version even though it's updated.

### Description

If you install the latest version of ArmCord on a Raspberry Pi, launch it and sign in, then right click on the status icon... It shows version 3.1.3 instead of the newest version 3.2.1.

I originally reported this to ArmCord, thinking it was an issue with the app itself, but they say it's not them. https://github.com/ArmCord/ArmCord/issues/425

I did have an older version installed prior before installing the newer version so I'm not sure if that affects anything.

However, I also uninstalled and reinstalled ArmCord, but it still exhibited the same issue.

### What are your system specs (run the following command in your terminal)?

```bash
OS: Debian GNU/Linux 11 (bullseye)
OS architecture: 64-bit
Last updated Pi-Apps on: 07/18/2023
Latest Pi-Apps version: 07/19/2023
Kernel: aarch64 6.1.21-v8+
Device model: Raspberry Pi 4 Model B Rev 1.1 
Cpu name: Cortex-A72
Ram size: 3.79 GB
Raspberry Pi OS image version: 2022-09-22
Language: en_US.UTF-8
```

theofficialgman:
I don't think it is use. our scripts are up to date
https://github.com/Botspot/pi-apps/blob/master/apps/ArmCord/install-32
https://github.com/Botspot/pi-apps/blob/master/apps/ArmCord/install-64

send all install/uninstall logs in `~/pi-apps/logs/` folder for armcord here

TheAlienDrew:
> I don't think it is pi-apps. our scripts are up to date
> https://github.com/Botspot/pi-apps/blob/master/apps/ArmCord/install-32
> https://github.com/Botspot/pi-apps/blob/master/apps/ArmCord/install-64
> 
> send all install/uninstall logs in `~/pi-apps/logs/` folder for armcord here
> 

[...]

PR #2396, opened 2023-07-20
api: add minimum version to dummy deb for local and url packages

theofficialgman:
closes https://github.com/Botspot/pi-apps/issues/2395

it has yet to be seen if we need to account for different versions of the same app appearing in a previous dummy deb. the current `uniq`  doesn't handle that

### DIFF
--- a/api
+++ b/api
@@ -489,7 +489,9 @@ $(echo "$existing_deps" | sed -z 's|, |\n|g')"
   { #create dummy apt package that depends on the packages this app requires
     
     #this stores the comma-separated list of dependency packages. It removes duplicate entries.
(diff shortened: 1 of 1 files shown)


theofficialgman:
> it has yet to be seen if we need to account for different versions of the same app appearing in a previous dummy deb. the current `uniq` doesn't handle that

fixed

Botspot:
Is this output expected and acceptable?
```
update-alternatives: warning: forcing reinstallation of alternative /opt/ArmCord/armcord because link group armcord is broken
```

theofficialgman:
> It could be circumstantial, but this is the first time I have noticed it.

never seen that before but its not from us or anything we did here. I'm assuming it has something to do with how armcord installs/uninstalls.

can't find anything helpful on the internet as to what it means and what causes it

PR #2400, opened 2023-07-23
Switch to adoptium from openjdk

theofficialgman:
switches to temurin java from openjdk due to openjdk 8/11 having been removed in bookworm.

Some install scripts had cases specific to ubuntu to use openjdk so those have been left since ubuntu shows no interest in removing any openjdk LTS (8/11/17) from their distros.

### DIFF
--- a/api
+++ b/api
@@ -693,6 +693,45 @@ debian_ppa_installer() { #setup a PPA on a Debian distro. Arguments: ppa_name di
   fi
 }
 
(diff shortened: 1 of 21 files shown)


theofficialgman:
force push because shellcheck caught my syntax errors https://github.com/Botspot/pi-apps/actions/runs/5638706031/job/15273111342?pr=2400#step:4:3263

PR #2404, opened 2023-07-29
README.md | Made readme more consistent

NajmAjmal:
Improvements include
 - Making readme more consistent
 - Removal of "-"

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ This will not uninstall any apps that you installed through Pi-Apps.
 ```
 </details>
 
-## To run Pi-Apps
+## To Run Pi-Apps
 - From the start menu: Accessories -> Pi Apps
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,9 +76,9 @@ This will not uninstall any apps that you installed through Pi-Apps.
 ```
 </details>
 
-## To run Pi-Apps
```
Botspot, line 79: "To" is capitalized because it starts a complete phrase, while Pi-Apps is capitalized because it is a proper name.
I suspect that your reason to capitalize "Run" is to capitalize every word, like the "To Update Pi-Apps" title below it.
Honestly, in the name of consistency, I would prefer having "update" lowercase instead of this change.
Any comments?

theofficialgman: agreed with Botspot, update should be lowercase

PR #2408, opened 2023-07-30
Install | Rephrased apt update fail

NajmAjmal:
Made it grammatically correct


Botspot:
I'm fairly confident that both before and after are grammatically correct. Your confusion may stem from thinking of the word "Pi-Apps" as a plural noun, but in most of our documentation we treat "Pi-Apps" as a singular noun.

theofficialgman:
Actually both are grammatically correct for both plural and singular nouns.

@NajmAjmal 
Pi-Apps is the name of the project itself. As Botspot said that makes it a singular noun.

I also prefer the original wording.

PR #2409, opened 2023-08-01
settings | Fixed Name Typo

NajmAjmal:
Made Pi Apps -> Pi-Apps.
Now it is more consistent.

### DIFF
--- a/settings
+++ b/settings
@@ -48,7 +48,7 @@ fi
 if [ ! -f ~/.local/share/applications/pi-apps-settings.desktop ];then
   echo "Creating Settings menu button"
   echo "[Desktop Entry]
-Name=Pi Apps Settings
+Name=Pi-Apps Settings
 Comment=Configure Pi-Apps or create an App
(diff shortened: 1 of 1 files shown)


theofficialgman:
actually its not a typo on this only makes it inconsistent. 
all our .desktop files from the install scripts have the name "Pi Apps". https://github.com/Botspot/pi-apps/blob/master/install

even though I do think that they should probably all be "Pi-Apps" personally

theofficialgman:
@Botspot you missed the place that this PR actually covered. The settings script. Might be others.

Botspot:
> @Botspot you missed the place that this PR actually covered. The settings script. Might be others.

Look more closely at that commit again. I made changes until `grep -r 'Name=Pi Apps' ~/pi-apps` returned nothing but matches from the update folder.

theofficialgman:
Oh I must have missed it. I didn't see settings in the commit before. My bad.

Issue #2401, opened 2023-07-26
PiOS Bookworm Incompatibilities


theofficialgman:
Apps with known compatibility issues (installing or running) on Bookworm that need to be fixed in no particular order

Uses pip:
- [x] CommanderPi
- [x] Powerline Shell https://github.com/Botspot/pi-apps/actions/runs/5674579997/job/15378383497#step:3:1184
- [x] PiGro
- [x] Quartz https://github.com/Botspot/pi-apps/actions/runs/5674580679/job/15378397179#step:3:1195
- [x] BlockPi (removed) (upstream https://github.com/alienzhangyw/BlockPi/issues/15 ) https://github.com/Botspot/pi-apps/actions/runs/5674572818/job/15378360281#step:3:1132
- [x] Pi-Apps Terminal Plugin (python) (removed)
- [x] Email Checker (removed) https://github.com/Botspot/pi-apps/actions/runs/5674575175/job/15378372480#step:3:1177
- [x] Sysmon (removed) (upstream https://github.com/t0xic0der/sysmon/issues/52 ) https://github.com/Botspot/pi-apps/actions/runs/5674582104/job/15378394996#step:3:1179

Other errors:
- [x] Prusaslicer (https://github.com/davidk/PrusaSlicer-ARM.AppImage/issues/29)
- [x] Stunt Rally (disabled, see https://github.com/Botspot/pi-apps/issues/2160)  https://github.com/Botspot/pi-apps/actions/runs/5674581872/job/15378396902#step:3:1256
- [ ] VeraCrypt (upstream https://github.com/veracrypt/VeraCrypt/issues/1046) https://github.com/Botspot/pi-apps/actions/runs/5674583305/job/15378399222#step:3:1212
- [x] KeePassXC https://github.com/Botspot/pi-apps/actions/runs/5674576980/job/15378423712#step:3:1212
- [x] Lokinet (already disabled) https://github.com/Botspot/pi-apps/actions/runs/5674577731/job/15378421562#step:3:1182
- [x] Marathon https://github.com/Botspot/pi-apps/actions/runs/5674577904/job/15378418240#step:3:2880
- [x] AstroMenace https://github.com/Botspot/pi-apps/actions/runs/5674572478/job/15378359527#step:3:2170
- [x] BlueJ Java IDE (upstream PiOS) https://github.com/k-pet-group/BlueJ-Greenfoot/issues/2249
[...]

PR #2419, opened 2023-08-04
api: add `remove_deprecated_app` and `userinput_func` functions

theofficialgman:
also remove deprecated apps:
Email Checker
Pi-Apps Terminal Plugin (python)
Sysmon
BlockPi

work towards https://github.com/Botspot/pi-apps/issues/2401

I have tested this
```
case $(ps -o stat= -p $$) in
  *+*) # Running in foreground
```
within a pi-apps runonce and it does work to detect when it is running as part of a foreground process.
We need to be running in a foregroud process with GUI available in order to prompt the user to chose to uninstall the application before it is removed.

the process is foreground with GUI available when the runonce-entries are run as part of the manage-daemon updater as well as after any install/uninstall from the pi-apps GUI.


theofficialgman:
this is still a draft as I am looking for feedback/cleanup for `userinput_func`

etc/runonce-entries
```diff
@@ -62,6 +62,30 @@ runonce <<"EOF"
   refresh_all_pkgapp_status
 EOF
 
+# remove deprecated apps (only when running in a visible terminal)
```
Botspot, line 65: Presumably the importance of running this in a visible terminal is to ensure this runonce only runs when the user is already installing an app and a terminal is visible to display the progress of uninstalling deprecated apps?

Botspot:
Reviewed now. Sorry for the delay - I have a lot of stuff to do in the real world.

userinput_func looks rather flexible. Do you have any other specific places in mind to replace current dialogs?
My main two concerns are:
- Feeling constrained to use the few options userinput_func offers, resulting in potentially suboptimal dialog designs for unforeseen situations. (Take the Arduino color choice dialog for an example of how yad freedom resulted in a better design)
- Muddying the water if we try to get app-contributors to use this. Now they have to understand when to use yad and when to use userinput_func. For people in the future who try to understand the source code, there would now be two general dialog commands to watch out for and understand. (Cue XKCD 927)

etc/runonce-entries
theofficialgman: Run once runs as part of set-status. That happens when pi-apps opens in a background terminal, as well as during any update, install, or uninstall in a foreground terminal.

There are two reasons it needs to be only in foreground.

Background terminals cannot and should not be used since the user won't see any terminal output from any uninstalls executed from deprecated apps. 

In addition, if sudo is needed and the user does not have passwordless sudo (so any normal Linux distro that isn't piOS) the runonce would hang in the background waiting for input in a terminal that doesn't have any way for the user to enter a password.
sudo_popup isn't an option since it uses pkexec which has its own set of environment variables (clears local environment) and would break many scripts. Also every sudo prompt would ask for the password which would not be ideal.

theofficialgman:
> userinput_func looks rather flexible. Do you have any other specific places in mind to replace current dialogs? 

It isn't intended to replace current dialogs.

> * Feeling constrained to use the few options userinput_func offers, resulting in potentially suboptimal dialog designs for unforeseen situations. (Take the Arduino color choice dialog for an example of how yad freedom resulted in a better design)
> * Muddying the water if we try to get app-contributors to use this. Now they have to understand when to use yad and when to use userinput_func. For people in the future who try to understand the source code, there would now be two general dialog commands to watch out for and understand. (Cue XKCD 927)
> 
> To be clear, neither of those concerns make me disapprove userinput_func, but hopefully they help show why I'm not as enthusiastic about it as you are.

See above, it's not a replacement for custom dialogs. It's a tool for script writers to use at their choice not needing to understand yad and duplicate work. An earlier version `userinput_func` is already shipped in multiple scripts and I have seen multiple PRs from contributors also copy my function (verbatim) and put it into their own scripts. It just makes sense to clean it up and unify it so that

PR #2428, opened 2023-08-17
Icons update

Rak1ta:


Botspot:
Can you provide a source for when the icon was updated? It appears to not have changed in 7 years, and is the old (current) version. https://github.com/linuxmint/timeshift/blob/master/icons/timeshift.png

Rak1ta:
@Botspot I'm looking at an icon from an application package. To match the icon in the application menu with the icon in pi-apps (only if there is no new version of the icon for Bullseye and higher).

Botspot:
I quickly installed Timeshift to verify. Thanks @Rak1ta!

Issue #2354, opened 2023-06-07
malware on Tor browser pi-apps


ghost:
i've scanned Tor browser on pi-apps with ClamTk and found that snowflake and obs4 contain Unix.Malware.kaiji-1000 chinese malware. I wonder if anyone else has identified this.

Botspot:
Can confirm.
```
pi@raspberrypi:~ $ clamscan /home/pi/tor-browser/Browser/TorBrowser/Tor/PluggableTransports/snowflake-client
/home/pi/tor-browser/Browser/TorBrowser/Tor/PluggableTransports/snowflake-client: Unix.Malware.Kaiji-10003916-0 FOUND

----------- SCAN SUMMARY -----------
Known viruses: 8668536
Engine version: 0.103.8
Scanned directories: 0
Scanned files: 1
Infected files: 1
Data scanned: 10.97 MB
Data read: 10.31 MB (ratio 1.06:1)
Time: 63.508 sec (1 m 3 s)
Start Date: 2023:06:07 18:56:37
End Date:   2023:06:07 18:57:40
pi@raspberrypi:~ $ clamscan /home/pi/tor-browser/Browser/TorBrowser/Tor/PluggableTransports/obfs4proxy
/home/pi/tor-browser/Browser/TorBrowser/Tor/PluggableTransports/obfs4proxy: Unix.Malware.Kaiji-10003916-0 FOUND

----------- SCAN SUMMARY -----------
Known viruses: 8668536
Engine version: 0.103.8
Scanned directories: 0
Scanned files: 1
Infected files: 1
Data scanned: 6.52 MB
Data read: 6.12 MB (ratio 1.06:1)
Time: 62.836 sec (1 m 2 s)
Start Date: 2023:06:07 18:58:51
End Date:   2023:06:07 18:59:54
```
Pi-Apps downloads tor-browser from sourceforge: https://sourceforge.net/projects/tor-browser-ports
These findings are concerning and I will try to scan some older versions to see if this has been added recently or if it was absent at some time in the past.

Botspot:
At this time, I would like to avoid blaming the user on sourceforge for deliberately embedding malware in tor. The user seems legitimate and provides build instructions. It is possible that person's build server was compromised with Kaiji, or maybe clamscan has a false positive.

Botspot:
I downloaded four different versions from the sourceforge.

Found in:
- 11.5
- 11.0

Not found in:
- 10.5
- 10.5.10

[...]

PR #2429, opened 2023-08-18
Create SECURITY.md

KOSASIH:


Botspot:
Hello @KOSASIH, thank you for taking the time to suggest this.
Obviously you do not intend to have this file merged in its current state, as it is a placeholder with suggested information, rather than with real information.
It is important to note that Pi-Apps is a **rolling release**. There are no version numbers, but instead, Pi-Apps does a more complicated equivalent of `git pull` to stay up to date. In addition to that, the only practical security 'vulnerabilities' that could be introduced would be either through an app's installation script, or from the app itself. Pi-Apps core scripts would need a lot of changes to provide any meaningful attack vector.
The only recent vulnerability report that I can remember was when somebody thought that Tor Browser was malware. See #2354. And in that case, that user simply opened an issue which was perfectly acceptable.

@KOSASIH, I am not sure how familiar you are with the Pi-Apps project, but I do know your past experience with other projects is valuable. If you can, please try to explain a bit why you feel a SECURITY.md is important to add, given the context I've provided above.

theofficialgman:
closing as @KOSASIH appears to be mostly bot commits from any repo they create or fork

Issue #2198, opened 2022-12-16
Tester wanted: to get steamlink running on PiOS Bullseye


Botspot:
Steamlink runs fine on PiOS Buster, but on Bullseye it displays this error on launch:
![IMG_4137](https://user-images.githubusercontent.com/54716352/208020368-48dcf9c8-b61a-44aa-ac67-5a012871803f.jpg)
This is caused by a safety check in the `~/.local/share/SteamLink/steamlink.sh` script:
```bash
# Make sure we're not trying to display DRM frames on X11
if [ ! -f "$TOP/.ignore_x11" ] && [ "$DISPLAY" != "" ]; then
    OS_VERSION=$(fgrep VERSION_ID /etc/os-release | sed 's,.*"\([0-9]\+\)",\1,')
    if [ "$OS_VERSION" -gt 10 ]; then
        show_message --warning $"X11 is not supported on this version of Raspbian. You can run on the console or downgrade to Raspberry Pi Buster."
        exit 1
    fi
fi
```
While I can easily disable thus safety check to cause the fullscreen UI to load, this does nbot really prove anything. Reportedly, the problem with Bullseye only occurs when the game video stream starts. As I do not have immediate access to another computer on the network with Steam installed, I really have no way to test what graphical tweaks are necessary to make it fully work on PiOS Bullseye.

Tester requested! I want to see if switching to the FKMS GL driver allows the video stream to render, or if just running it in the console is enough. Additionally, I am interested in finding a way to run Steamlink from the graphical desktop, without any manual console-switching or changes to /boot/config.txt.
If this venture fails, then I see little point in continuing to carry Steamlink in Pi-Apps if it cannot run on the default OS going forward.

Those interested in workarounds can look to this tutorial: https://blog.iancolwell.ca/steamlink-aarch64

Botspot:
The tester needs to have a Pi4 that is running PiOS Bullseye, with access to a PC that is running Steam.

theofficialgman:
might be a good time to plug moonlight-qt (client software) and sunshine (an alterntive open source host software instead of nvidia geforce experience remote play) as steamlink alternatives.

there are specific debian repos of moonlight-qt for the pi (buster and bullseye 32/64bit) and nvidia jetson line (switch included) both with hardware accelerated decoding h264/hevc for fast and power efficient streaming

Rak1ta:
[...]

PR #2455, opened 2023-10-05
Anydesk arm64

Botspot:
Testing required on systems other than PiOS Bullseye 64-bit.

This uses multiarch and some dirty lib symlinking to get Anydesk running on PiOS 64-bit. It avoids installing libraspberrypi-dev:armhf, as that package conflicts with libraspberrypi-dev:arm64.

I doubt Anydesk would successfully open on a system other than PiOS due to missing libs.

### DIFF
--- a/apps/AnyDesk/install-64
+++ b/apps/AnyDesk/install-64
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+version=6.3.0-1
+
+function check-armhf() {
+	ARMHF="$(dpkg --print-foreign-architectures | grep "armhf")"
+}
+
+#add armhf architecture (multiarch)
+check-armhf
+if [[ "$ARMHF" == *"armhf"* ]]; then
+  echo "armhf arcitecture already added..."
+else
+  sudo dpkg --add-architecture armhf
+  check-armhf
+  if [[ "$ARMHF" != *"armhf"* ]]; then
+    error "armhf architecture should be added by now, but it isn't!"
+  fi
+fi
+
+unset rpi_arm_userspace
+# only install the libraspberrypi0 arm32 package if the user already has the libraspberrypi0 arm64 package installed
+if package_installed libraspberrypi0 ; then
+  rpi_arm_userspace="libraspberrypi0:armhf"
+fi
+
+install_packages https://download.anydesk.com/rpi/anydesk_${version}_armhf.deb libgles-dev:armhf libegl-dev:armhf libpolkit-gobject-1-0:armhf $rpi_arm_userspace || exit 1
+
+#As libraspberrypi-dev:armhf cannot be installed with multiarch, symlink the required libs from libraspberrypi0:armhf
+if [ -e /usr/lib/arm-linux-gnueabihf/libbcm_host.so.0 ] && [ -e /usr/lib/arm-linux-gnueabihf/libvcos.so.0 ] && [ -e /usr/lib/arm-linux-gnueabihf/libvchiq_arm.so.0 ];then
+  sudo ln -s /usr/lib/arm-linux-gnueabihf/libbcm_host.so.0 /usr/lib/arm-linux-gnueabihf/libbcm_host.so
+  sudo ln -s /usr/lib/arm-linux-gnueabihf/libvcos.so.0 /usr/lib/arm-linux-gnueabihf/libvcos.so
+  sudo ln -s /usr/lib/arm-linux-gnueabihf/libvchiq_arm.so.0 /usr/lib/arm-linux-gnueabihf/libvchiq_arm.so
+else
+  error "Could not locate needed libs to symlink! Please ask pi-apps developers for help about this."
+fi
+# Solve error on Bullseye: "anydesk: error while loading shared libraries: libbrcmGLESv2.so: cannot open shared object file: No such file or directory"
+
+[ ! -e /usr/lib/libbrcmGLESv2.so ] && sudo ln -s /usr/lib/arm-linux-gnueabihf/libGLESv2.so /usr/lib/arm-linux-gnueabihf/libbrcmGLESv2.so
+[ ! -e /usr/lib/libbrcmEGL.so ] && sudo ln -s /usr/lib/arm-linux-gnueabihf/libEGL.so /usr/lib/arm-linux-gnueabihf/libbrcmEGL.so
+exit 0
--- a/apps/AnyDesk/install-64
+++ b/apps/AnyDesk/install-64
@@ -24,18 +24,19 @@ if package_installed libraspberrypi0 ; then
   rpi_arm_userspace="libraspberrypi0:armhf"
 fi
 
-install_packages https://download.anydesk.com/rpi/anydesk_${version}_armhf.deb libgles-dev:armhf libegl-dev:armhf libpolkit-gobject-1-0:armhf $rpi_arm_userspace || exit 1
+install_packages https://download.anydesk.com/rpi/anydesk_${version}_armhf.deb libgles-dev:armhf libegl-dev:armhf libpolkit-gobject-1-0:armhf $rpi_arm_userspace patchelf || exit 1
 
-#As libraspberrypi-dev:armhf cannot be installed with multiarch, symlink the required libs from libraspberrypi0:armhf
-if [ -e /usr/lib/arm-linux-gnueabihf/libbcm_host.so.0 ] && [ -e /usr/lib/arm-linux-gnueabihf/libvcos.so.0 ] && [ -e /usr/lib/arm-linux-gnueabihf/libvchiq_arm.so.0 ];then
-  sudo ln -s /usr/lib/arm-linux-gnueabihf/libbcm_host.so.0 /usr/lib/arm-linux-gnueabihf/libbcm_host.so
-  sudo ln -s /usr/lib/arm-linux-gnueabihf/libvcos.so.0 /usr/lib/arm-linux-gnueabihf/libvcos.so
-  sudo ln -s /usr/lib/arm-linux-gnueabihf/libvchiq_arm.so.0 /usr/lib/arm-linux-gnueabihf/libvchiq_arm.so
-else
-  error "Could not locate needed libs to symlink! Please ask pi-apps developers for help about this."
-fi
-# Solve error on Bullseye: "anydesk: error while loading shared libraries: libbrcmGLESv2.so: cannot open shared object file: No such file or directory"
+#make anydesk search armhf libs first
+sudo patchelf --set-rpath /usr/lib/arm-linux-gnueabihf /usr/bin/anydesk
 
-[ ! -e /usr/lib/libbrcmGLESv2.so ] && sudo ln -s /usr/lib/arm-linux-gnueabihf/libGLESv2.so /usr/lib/arm-linux-gnueabihf/libbrcmGLESv2.so
-[ ! -e /usr/lib/libbrcmEGL.so ] && sudo ln -s /usr/lib/arm-linux-gnueabihf/libEGL.so /usr/lib/arm-linux-gnueabihf/libbrcmEGL.so
-exit 0
+#libraspberrypi0:armhf provides libs with so.0, but it is not multiarch compatible
+#so patch anydesk to use equivalent .so libs from libraspberrypi-dev:armhf
+
+sudo patchelf --replace-needed libbcm_host.so libbcm_host.so.0 /usr/bin/anydesk
+sudo patchelf --replace-needed libvcos.so libvcos.so.0 /usr/bin/anydesk
+sudo patchelf --replace-needed libvchiq_arm.so libvchiq_arm.so.0 /usr/bin/anydesk
+
+# Solve error: "anydesk: error while loading shared libraries: libbrcmGLESv2.so: cannot open shared object file: No such file or directory"
+
+sudo patchelf --replace-needed libbrcmGLESv2.so libGLESv2.so /usr/bin/anydesk
+sudo patchelf --replace-needed libbrcmEGL.so libEGL.so /usr/bin/anydesk
--- a/apps/AnyDesk/install-64
+++ b/apps/AnyDesk/install-64
@@ -9,7 +9,7 @@ function check-armhf() {
 #add armhf architecture (multiarch)
 check-armhf
 if [[ "$ARMHF" == *"armhf"* ]]; then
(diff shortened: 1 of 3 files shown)


apps/AnyDesk/install-64
```diff
@@ -0,0 +1,41 @@
...
+  sudo ln -s /usr/lib/arm-linux-gnueabihf/libbcm_host.so.0 /usr/lib/arm-linux-gnueabihf/libbcm_host.so
+  sudo ln -s /usr/lib/arm-linux-gnueabihf/libvcos.so.0 /usr/lib/arm-linux-gnueabihf/libvcos.so
+  sudo ln -s /usr/lib/arm-linux-gnueabihf/libvchiq_arm.so.0 /usr/lib/arm-linux-gnueabihf/libvchiq_arm.so
+else
+  error "Could not locate needed libs to symlink! Please ask pi-apps developers for help about this."
+fi
+# Solve error on Bullseye: "anydesk: error while loading shared libraries: libbrcmGLESv2.so: cannot open shared object file: No such file or directory"
+
+[ ! -e /usr/lib/libbrcmGLESv2.so ] && sudo ln -s /usr/lib/arm-linux-gnueabihf/libGLESv2.so /usr/lib/arm-linux-gnueabihf/libbrcmGLESv2.so
+[ ! -e /usr/lib/libbrcmEGL.so ] && sudo ln -s /usr/lib/arm-linux-gnueabihf/libEGL.so /usr/lib/arm-linux-gnueabihf/libbrcmEGL.so
```
theofficialgman, line 40: I don't like this at all... I see a high possibility for this to cause issues for other software given it is dynamically linkable from arm64 where you are placing the symlinks. If we were to go forward with this PR, I would want to use `patchelf` to change the dynamic link targets of the anydesk binary from `/usr/lib/libbrcmGLESv2.so` to `/usr/lib/arm-linux-gnueabihf/libbrcmEGL.so`

Regardless, I'm going to suggest that we wait for bookworm to release because they may decide to make 64bit the default option on the website and imager which will push anydesk to support it officially through a 64bit package. In addition, this appears to use the legacy broadcom gpu libraries, not sure if those will work at all on the pi5 given it has a different GPU and raspberry pi has been trying to deprecate and remove support for them.

theofficialgman:
@Botspot FYI, the reason this failed https://github.com/Botspot/pi-apps/actions/runs/6423970408 is because you tested as a PR# from the master branch. As the script is written I have it only import any updated app, not script files. That could be changed but its just not how it functions right now.

If you want to test a branch in its entirety I suggest selecting the branch from the dropdown and then putting in the app name as I did here -> https://github.com/Botspot/pi-apps/actions/runs/6477319941

theofficialgman:
weird... https://github.com/Botspot/pi-apps/actions/runs/6477319941/job/17587322949#step:3:1360 
`libraspberrypi0:armhf` is definitely in the repo on the web but local `Packages` files do NOT have it as confirmed by the empty output here from the debug logging https://github.com/Botspot/pi-apps/actions/runs/6477319941/job/17587322949#step:3:1380

worth investigating, maybe the buster arm64 image ships with corrupted local files or it is being hidden with a pin?

Botspot:
> weird... https://github.com/Botspot/pi-apps/actions/runs/6477319941/job/17587322949#step:3:1360 `libraspberrypi0:armhf` is definitely in the repo on the web but local `Packages` files do NOT have it as confirmed by the empty output here from the debug logging https://github.com/Botspot/pi-apps/actions/runs/6477319941/job/17587322949#step:3:1380
> 
> worth investigating, maybe the buster arm64 image ships with corrupted local files or it is being hidden with a pin?

Maybe Buster's libraspberrypi0 was never made compatible with multiarch. I think I remember something about that, but it's been awhile.

theofficialgman:
> but local `Packages` files do NOT have it as confirmed by the empty output here from the debug logging https://github.com/Botspot/pi-apps/actions/runs/6477319941/job/17587322949#step:3:1380

wait no. I have the debug logging sent directly to the logfile https://github.com/Botspot/pi-apps/blob/master/api#L1772 so it never prints the terminal. we see it in error reports because the logfile is uploaded but this CI does not cat the logfile.

I should add it so that install/uninstall failures cat the logfile in this CI tester.

theofficialgman:
Same as -> https://github.com/Botspot/pi-apps/issues/2198#issuecomment-1868199669

We should be encouraging upstream developers to write their applications generically, not conforming to old, unstable, hacky workarounds

Anydesk users are encouraged to contact support https://anydesk.com/en/support/contact-us

Botspot:
> We should be encouraging upstream developers to write their applications generically, not conforming to old, unstable, hacky workarounds

Re-opening, as we already tend to take the users' side and implement workarounds for things that vendors should really be responsible for. (Take zoom for example.)

apps/AnyDesk/install-64
Botspot: I'd be interested to hear more about patchelf

Botspot:
I would like to give this some priority because it is being sought after a lot by the google-searching population.

This is THE most popular install-app page we have.
20240307_16h34m39s_grim

And these are the search terms that are responsible.
20240307_16h33m53s_grim
The people have spoken, and I would like to try to deliver, as anydesk is not...

apps/AnyDesk/install-64
```diff
@@ -0,0 +1,42 @@
...
+  fi
+fi
+
+unset rpi_arm_userspace
+# only install the libraspberrypi0 arm32 package if the user already has the libraspberrypi0 arm64 package installed
+if package_installed libraspberrypi0 ; then
+  rpi_arm_userspace="libraspberrypi0:armhf"
+fi
+
+install_packages https://download.anydesk.com/rpi/anydesk_${version}_armhf.deb libgles-dev:armhf libegl-dev:armhf libpolkit-gobject-1-0:armhf $rpi_arm_userspace patchelf || exit 1
```
theofficialgman, line 27: likely unnecessary dependencies
development libraries are nearly never required for a package
control file for anydesk already pulls in egl and glvnd libraries which pull in everything else needed

Botspot: these are the libs installed by libraspberrypi0:armhf:
```
/usr/lib/arm-linux-gnueabihf/libbcm_host.so.0
/usr/lib/arm-linux-gnueabihf/libdebug_sym.so.0
/usr/lib/arm-linux-gnueabihf/libdtovl.so.0
/usr/lib/arm-linux-gnueabihf/libvchiq_arm.so.0
/usr/lib/arm-linux-gnueabihf/libvcos.so.0
```
I am not sure that the default deps pull those in, at least for armhf

theofficialgman:
this screen needs to force 4K pagesize, as does any other script that uses armhf on an arm64 kernel.
there will be errors at install and runtime on 16K pagesize kernels
eg:
```
Setting up libgdk-pixbuf-2.0-0:armhf (2.42.10+dfsg-1+b1) ...
/usr/lib/arm-linux-gnueabihf/gdk-pixbuf-2.0/gdk-pixbuf-query-loaders: error while loading shared libraries: libglib
-2.0.so.0: ELF load command address/offset not page-aligned
Setting up libcairo-gobject2:armhf (1.16.0-7+rpt1) ...
Setting up libatk1.0-0:armhf (2.46.0-5) ...
Setting up libpango-1.0-0:armhf (1.50.12+ds-1) ...
Setting up libpangoft2-1.0-0:armhf (1.50.12+ds-1) ...
Setting up libpangocairo-1.0-0:armhf (1.50.12+ds-1) ...
Setting up libgtk2.0-0:armhf (2.24.33-2+rpt1) ...
/usr/lib/arm-linux-gnueabihf/libgtk2.0-0/gtk-query-immodules-2.0: error while loading shared libraries: libgmodule-
2.0.so.0: ELF load command address/offset not page-aligned
Setting up librsvg2-2:armhf (2.54.7+dfsg-1~deb12u1) ...
Setting up librsvg2-common:armhf (2.54.7+dfsg-1~deb12u1) ...
Setting up libgail18:armhf (2.24.33-2+rpt1) ...
Setting up anydesk:armhf (6.3.0) ...
Created symlink /etc/systemd/system/multi-user.target.wants/anydesk.service → /etc/systemd/system/anydesk.service.
Setting up libgail-common:armhf (2.24.33-2+rpt1) ...
Processing triggers for libc-bin (2.36-9+rpt2+deb12u4) ...
Processing triggers for libgdk-pixbuf-2.0-0:armhf (2.42.10+dfsg-1+b1) ...
/usr/lib/arm-linux-gnueabihf/gdk-pixbuf-2.0/gdk-pixbuf-query-loaders: error while loading shared libraries: libglib
-2.0.so.0: ELF load command address/offset not page-aligned
```
and
```
anydesk: error while loading shared libraries: libatk-1.0.so.0: ELF load command address/offset not page-aligned
```

Botspot:
also it seems to fail to launch if on wayland

apps/AnyDesk/install-64
theofficialgman: I am referring to `libgles-dev:armhf libegl-dev:armhf`

apps/AnyDesk/install-64
```diff
@@ -0,0 +1,42 @@
...
+  if [[ "$ARMHF" != *"armhf"* ]]; then
+    error "armhf architecture should be added by now, but it isn't!"
+  fi
+fi
+
+unset rpi_arm_userspace
+# only install the libraspberrypi0 arm32 package if the user already has the libraspberrypi0 arm64 package installed
+if package_installed libraspberrypi0 ; then
+  rpi_arm_userspace="libraspberrypi0:armhf"
+fi
```
theofficialgman, line 25: the deprecated libraspberrypi0 libraries are required to run anydesk
```
libbcm_host.so
libvcos.so
libvchiq_arm.so
libbrcmGLESv2.so
libbrcmEGL.so
```
this needs to be changed to check if `libraspberrypi0:armhf` exists and if available always install it. otherwise exit the script.

apps/AnyDesk/install-64
Botspot: Without those 2 installed, and then reinstalling the anydesk package and allowing it to pull in deps:
```
$ anydesk
anydesk: error while loading shared libraries: libGLESv2.so: cannot open shared object file: No such file or directory
```

Botspot: Oh, I get it. AnyDesk should be able to use `libGLESv2.so.2` provided by libgles2, instead of specifically needing libGLESv2.so that is only provided by the -dev package.
Looks like this can be fixed with more elf patching.

theofficialgman: hmm... odd
```
/usr/lib/arm-linux-gnueabihf/libGLESv2.so	[libgles-dev](https://packages.debian.org/bookworm/libgles-dev) [armhf]
```
its not normal that a package uses this library.

those dependencies need to stay then

theofficialgman:
I have tested this on piOS bookworm on RPi5 with 4K pagesize kernel on piOS wayfire and it does not function.
running `anydesk` from terminal shows output but no GUI starts and I see no anydesk processes

Botspot:
> I have tested this on piOS bookworm on RPi5 with 4K pagesize kernel on piOS wayfire and it does not function. running `anydesk` from terminal shows output but no GUI starts and I see no anydesk processes

See my comment above: "also it seems to fail to launch if on wayland"

strace shows that it is having issues with the `/dev/vchiq` file not existing.

Supposedly anydesk runs on pi5 if you switch to X11. https://forums.raspberrypi.com/viewtopic.php?t=366277
I will try that shortly.

apps/AnyDesk/install-64
theofficialgman: @Botspot you missed my response ^
you need to revert those -dev changes from https://github.com/Botspot/pi-apps/pull/2455/commits/9b0ce0a37b64a196613866bd7da0af7f2c109288 . they are needed. I checked

Botspot: Is patching the elf to use the non-dev libs not sufficient?

theofficialgman:
> > I have tested this on piOS bookworm on RPi5 with 4K pagesize kernel on piOS wayfire and it does not function. running `anydesk` from terminal shows output but no GUI starts and I see no anydesk processes
> 
> See my comment above: "also it seems to fail to launch if on wayland"
> 
> strace shows that it is having issues with the `/dev/vchiq` file not existing.
> 
> Supposedly anydesk runs on pi5 if you switch to X11. https://forums.raspberrypi.com/viewtopic.php?t=366277 I will try that shortly.

I tried. `/dev/vchiq` still does not exist and (obviously) "* failed to open vchiq instance" appears

Botspot:
> I tried. `/dev/vchiq` still does not exist and (obviously) "* failed to open vchiq instance" appears

Maybe that is no longer a thing on Pi5?

theofficialgman:
and that is indicated by the forum thread
https://forums.raspberrypi.com/viewtopic.php?t=366277#p2197174

given the obvious issues with it not functioning as expected, I think this PR can be closed and we can continue to mark this as an upstream issue.

All anydesk needs to do is build their standard client for desktop linux for armhf and arm64 targets. no need for any deprecated broadcom library use

theofficialgman:
> Supposedly anydesk runs on pi5 if you switch to X11. https://forums.raspberrypi.com/viewtopic.php?t=366277 I will try that shortly.

nice... switching back to wayland in raspi-config now gives me a black screen on boot. good thing this was already a fresh install

Botspot:
> given the obvious issues with it not functioning as expected, I think this PR can be closed and we can continue to mark this as an upstream issue.
> 
> All anydesk needs to do is build their standard client for desktop linux for armhf and arm64 targets. no need for any deprecated broadcom library use

Strange how all these remote-connection tools all use vchiq. Looks like that is the issue with steamlink as well: https://forums.raspberrypi.com/viewtopic.php?p=2174716&hilit=%2Fdev%2Fvchiq#p2174716

Botspot:
Still going to implement the api changes from this branch

theofficialgman:
> Strange how all these remote-connection tools all use vchiq. Looks like that is the issue with steamlink as well: https://forums.raspberrypi.com/viewtopic.php?p=2174716&hilit=%2Fdev%2Fvchiq#p2174716

its the direct interface to the GPU. They were doing it previously for performance reasons and since all the upstream support wasn't done yet.

Now that is complete and, to my knowledge, that direct GPU access is no longer allowed (hence why the device node is missing).

PR #2483, opened 2023-11-21
Fix grammatical error in XMRig install script

Crilum:
Probably don't merge this until a bigger change has to be made, or the auto-updater updates the version, so we don't unnecessarily trigger updates.

Also, I don't mind if this never gets merged. I just noticed a couple things to make the logging prettier, and... :man_shrugging:

### DIFF
--- a/apps/XMRig/install
+++ b/apps/XMRig/install
@@ -274,5 +274,5 @@ fi
 if [ -s $HOME/.config/xmrig.json ];then
   status "\nSince your configuration is ready to use, you can run 'xmrig' in a terminal to start the miner.\nOtherwize, you can use command-line flags to specify Wallet Address, mining pool, and so forth.\nHave fun!"
 else
-  status "\nSince you don't have a configuration yet, you can either create one later at '$HOME/.config/xmrig.json', or run 'xmrig' with command-line flags specify Wallet Address, mining pool, and so forth.\nHave fun!"
+  status "\nSince you don't have a configuration yet, you can either create one later at '$HOME/.config/xmrig.json', or run 'xmrig' with command-line flags to specify Wallet Address, mining pool, and so forth.\nHave fun!"
 fi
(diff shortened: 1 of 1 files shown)


Botspot:
I had seen somewhere some xmrig devs referring to this as XMRig when the name was being used as a title. Same as us using `pi-apps` and `Pi-Apps` interchangeably.
It was a while back though, and it's possible I just made up the capitalized XMRig word usage. I'd recommend looking at xmrig documentation to see if they always spell it lowercase, or only capitalize the first letter, or capitalize the first three. 
Which by the way makes sense as the actual Monero cryptocurrency has codename XMR, just as Bitcoin's codename is BTC.

Crilum:
Seems like `XMRig` is the official name:

image
image
(from https://xmrig.com)

theofficialgman:
>  or the auto-updater updates the version

well actually I had to disable the auto-updater for xmrig because the xmrig install requires user action and our auto-updater is a computer, not a person, and is incapable of providing this user action 🤖 

if you change the version to latest and find that it is still able to be built and run you are welcome to add that to this PR and we can merge afterwards.

Crilum:
> if you change the version to latest `v6.20.0-mo1` and find that it is still able to be built and run you are welcome to add that to this PR and we can merge afterwards.

Okay. I'll try building with the newest version (in qemu, if it works it's probably a good idea to have someone with an actual Raspberry test it).
Also.. What's the `-mo1` part at the end of the version number? I don't see it on the GitHub Releases..
image

theofficialgman:
> Okay. I'll try building with the newest version (in qemu, if it works it's probably a good idea to have someone with an actual Raspberry test it). Also.. What's the `-mo1` part at the end of the version number? I don't see it on the GitHub Releases.. image

looks like you are on the wrong repo. refer to the install script for `MoneroOcean/xmrig`

Botspot:
Closing this PR as there is nothing to merge. If there is further progress building the latest version, it would be better to discuss that in a new issue.

Issue #2490, opened 2023-11-28
PiGro 24.01 Release Preparation


actionschnitzel:
Hi Pi-Apps Team,

I plan to release PiGro 24.01 in January.
Pi-Apps currently fetches PiGro from GitHub. Whenever I needed to make changes, I submitted a pull request to your repository. However, for the past six months, I have also been providing DEBIAN packages. What is more convenient for you?

This is not a critical issue. I just wanted to let you know in case you need to make major changes that take time.

**New Dependencies:**
- `git`
- `xterm`
- `python3-pil`
- `python3-pil.imagetk`
- `python3-psutil`
- `python3-distro`
- `python3-bs4`
- `python3-dev`
- `python3-requests`
- `mpg123`
- `lolcat`
- `wmctrl`
- `gdebi`
- `mousepad`

(Note: Git may not need to be installed via Pi-Apps)

**Link to Latest Release:**
[PiGro Latest Release](https://github.com/actionschnitzel/PiGro-Aid-/releases/latest)

**DEBIAN Package Download:**
As seen in your FreeTub installer:

```
version=23.04
https://github.com/actionschnitzel/PiGroAid-/releases/download/${version}/pigro_jci-${version%}_all.deb
```

**I would also like to submit updated icons:**
![icon-24](https://github.com/Botspot/pi-apps/assets/67635013/8c90d97e-0e15-48fc-8010-893ab16a10af)
![icon-64](https://github.com/Botspot/pi-apps/assets/67635013/35ada6e3-1a11-41eb-9fb5-cb899b184bb3)

Best regards
Timo

Botspot:
Thanks for the heads-up. We're fine with switching to your deb release. Do you have your own apt repo? If not, that's fine.
This looks like a very simple change. Pi-Apps' main strength is its flexibility. It's a script, so any app can install by any means necessary.
While I could go ahead and make these changes myself, it would probably save us both time if you opened a pull request with the proposed changes. and yes, it looks like a copy of the FreeTube scripts would be a great starting point. This also gives you the chance to shoutout the new features in the description. If you can't tell, I think PiGro is pretty awesome especially for intermediate users who want to get stuff done quickly.

actionschnitzel:
@theofficialgman Wow, good that you've taken a closer look. I am using a script that I found. Must fix this as soon as possible.

actionschnitzel:
Update:
[...]

PR #2492, opened 2023-11-30
PiGro Updated Install Method

actionschnitzel:
I have updated the description. I hope my English makes sense. Additionally, I have updated the icons. I created the install script following the example of the FreeTube installer, and it worked for me.

Regarding the uninstaller, I'm not entirely sure; I'm not a Bash-Ninja... 
Since the old installation method clones to HOME, and the new DEBIAN PACKAGE installs to /opt, the old path must be deleted.
DEBIAN PACKAGE modified as discussed:
closes https://github.com/Botspot/pi-apps/issues/2490

Testing process:

1.Existing PiGro installation via Pi-Apps
2.Replace ~/pi-apps/apps/PiGro
3.PiGro Uninstall (GitHub) (worked)
4.PiGro install (DEB) (worked)


theofficialgman:
> Regarding the uninstaller, I'm not entirely sure; I'm not a Bash-Ninja... Since the old installation method clones to HOME, and the new DEBIAN PACKAGE installs to /opt, the old path must be deleted. DEBIAN PACKAGE modified as discussed: #2490

I took the liberty of simplifying that. There is no need to check if the folders/files exist before attempting to delete them. the `f` option from `rm` already does that. It is a no-operation with exit code 0 if there is nothing to do.

In theory, on upgrades, the old uninstall script gets run and that folder in home and the old .desktop file would already be removed, but sometimes we see strange things from users so you never really know. For a change like this we generally put it in both the install and uninstall script which is what I have done.

PR #2518, opened 2024-01-07
PPSSPP: latest release 1.16.6 build from source

theofficialgman:
also add auto-updater

work towards https://github.com/Botspot/pi-apps/issues/2160


theofficialgman:
I have tested this on Nintendo Switch Jammy ARM64 and RPI5 PiOS Bookworm ARM32 and ARM64.
OpenGL and Vulkan (when drivers are installed) works on ARM64 and OpenGLES and Vulkan (when drivers are installed) works on ARM32.

theofficialgman:
CI passed on all supported targets https://github.com/Botspot/pi-apps/actions/runs/7440930205

Additionally I confirmed that RPI4 PiOS Bullseye ARM64 OpenGL and Vulkan (when drivers are installed) works. No reason to believe that ARM32 won't work there as well.

Issue #2528, opened 2024-01-15
Add WideVine support in Ubuntu through Pi-Apps


OlivierDoriath:
Hello!
As things stand, installing Chromium through Pi-Apps on Ubuntu seems to install the Snap package. Adding the Raspberry Pi OS package for their branch of Chromium along with the libwidevinecdm0 package (and possibly other required packages) would allow for Ubuntu users to access DRM restricted content on their Raspberry Pi without using Raspberry Pi OS where Chromium has native WideVine support out of the box. It might also fix Better Chromium not working on Ubuntu.

theofficialgman:
> Hello! As things stand, installing Chromium through Pi-Apps on Ubuntu seems to install the Snap package.

chromium from pi-apps installs the `chromium-browser` package through apt. On ubuntu they have made this a dummy package that implements a install script for their chromium snap package.

> Adding the Raspberry Pi OS package for their branch of Chromium along with the libwidevinecdm0 package (and possibly other required packages) would allow for Ubuntu users to access DRM restricted content on their Raspberry Pi without using Raspberry Pi OS where Chromium has native WideVine support out of the box.

its not so simple. there are a few ways that widevine can be patched. See the commit message here for more info -> https://github.com/theofficialgman/l4t-debs/commit/09c7234202a29b5803f68f456ab9503600aa4722 . I am not sure which way piOS decided to patch their widevine.

Snap chromium actually already checks for Widevine on all architectures thanks to my request -> https://bugs.launchpad.net/ubuntu/+source/chromium-browser/+bug/2008433 . If you place a compatible Widevine (such as my patched version for all distros that use GLIBC 2.36+ out of the box https://github.com/theofficialgman/testing/releases/download/gmans-releases/WidevineCdm-2.36.tar.gz, just extract to the proper directory) then it will load and be used.

OlivierDoriath:
Hey!

Thank you for the swift reply!

I tried doing just that (extracted the archive to $HOME/.local/lib/ and reopened Chromium (Version 120.0.6099.199 (Official Build) snap (64-bit)) but the issue persists.

Am I missing something?

theofficialgman:
> I tried doing just that (extracted the archive to $HOME/.local/lib/ and reopened Chromium (Version 120.0.6099.199 (Official Build) snap (64-bit))) but the issue persists.

[...]

PR #2529, opened 2024-01-15
Better Chromium: install Raspberry Pi Foundation chromium on rpi Ubuntu

theofficialgman:
suggestion from https://github.com/Botspot/pi-apps/issues/2528
closes https://github.com/Botspot/pi-apps/issues/2528

see https://github.com/Pi-Apps-Coders/chromium-debs for the repo that re-hosts the rpi debs

### DIFF
--- a/apps/Better Chromium/uninstall
+++ b/apps/Better Chromium/uninstall
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# rpi ubuntu (22.04+) use Pi OS packaged chromium for better hardware support
+purge_packages
+rm_external_repo "pi-apps-coders-chromium" force
+sudo rm -f /etc/apt/preferences.d/pi-apps-coders-chromium
+
 #determine where the Chromium customizations folder is.
 if [ -d /etc/chromium.d ];then
   folder=/etc/chromium.d
--- a/apps/Better Chromium/uninstall
+++ b/apps/Better Chromium/uninstall
@@ -1,9 +1,15 @@
 #!/bin/bash
 
-# rpi ubuntu (22.04+) use Pi OS packaged chromium for better hardware support
-purge_packages
-rm_external_repo "pi-apps-coders-chromium" force
-sudo rm -f /etc/apt/preferences.d/pi-apps-coders-chromium
+if [[ "$1" != "update" ]];then
+  # rpi ubuntu (22.04+) can use Pi OS packaged chromium for better hardware support
+  if [ -f /etc/apt/sources.list.d/pi-apps-coders-chromium.sources ]; then
+    warning "When installing Better Chromium, you switched to PiOS chromium build. If you want to revert to the snap version, here's commands to do that:\n\n"\
+'~/pi-apps/api rm_external_repo "pi-apps-coders-chromium" "force"\n'\
+'sudo rm -f /etc/apt/preferences.d/pi-apps-coders-chromium\n'\
+'sudo apt install --only-upgrade chromium-* -y\n'\
+'sudo apt purge libwidevinecdm0'
+  fi
+fi
 
 #determine where the Chromium customizations folder is.
 if [ -d /etc/chromium.d ];then
--- a/apps/Better Chromium/install
+++ b/apps/Better Chromium/install
@@ -6,6 +6,11 @@ get_model &>/dev/null
 ubuntu_pios_chromium_available=0
 if [[ "$model" == *"Raspberry Pi"* ]] && [[ "$__os_id" == Ubuntu ]] && printf '%s\n' "22.04" "$__os_release" | sort -CV; then
   ubuntu_pios_chromium_available=1
+  # Ubuntu must be 22.10+ (GLIBC 2.36+) to be compatible with widevine
+  pin_hide_widevine=1
+  if printf '%s\n' "22.10" "$__os_release" | sort -CV; then
+    pin_hide_widevine=0
+  fi
 fi
 
 #determine if chromium is installed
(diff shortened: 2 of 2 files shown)


apps/Better Chromium/uninstall
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# rpi ubuntu (22.04+) use Pi OS packaged chromium for better hardware support
+purge_packages
+rm_external_repo "chromium" force
```
Botspot, line 5: I suggest changing the repo name to something more unique, just in case some user out there has made their own `/etc/apt/sources.list.d/chromium.list`, `/usr/share/keyrings/chromium-archive-keyring.gpg`, or `/etc/apt/sources.list.d/chromium.sources` files.

Botspot:
Also fairly sure that for widevine to show up at all on ubuntu, this logic needs to be changed to show the widevine option for the `chromium` package.
```bash
#detect if Widevine is enabled
if [ -d /opt/WidevineCdm ];then
  widevine=TRUE
else
  widevine=FALSE
fi
#detect if Widevine can be enabled
if package_available libwidevinecdm0 && package_installed chromium-browser;then
  widevine_possible=TRUE
  widevine_message=(--field="Enable Widevine DRM (for Netflix, Spotify)":CHK "$widevine")
else
  widevine_possible=FALSE
  widevine_message=()
fi
```

theofficialgman:
> Also I am fairly sure that for widevine to show up at all on ubuntu, this logic needs to be changed to display the widevine option for the `chromium` package.
> 
> ```shell
> #detect if Widevine is enabled
> if [ -d /opt/WidevineCdm ];then
>   widevine=TRUE
> else
>   widevine=FALSE
> fi
> #detect if Widevine can be enabled
> if package_available libwidevinecdm0 && package_installed chromium-browser;then
>   widevine_possible=TRUE
>   widevine_message=(--field="Enable Widevine DRM (for Netflix, Spotify)":CHK "$widevine")
> else
>   widevine_possible=FALSE
>   widevine_message=()
> fi
> ```

Nope. The logic is good and working since the piOS widevine package is used. See the apt repo.

apps/Better Chromium/uninstall
theofficialgman: `pi-apps-coders-chromium` sound good or do you have another suggestion?

Botspot:
> Nope. The logic is good and working since the piOS `libwidevinecdm0` and `chromium-browser` packages are used. See the apt repo.


It appeared to me that the `chromium` package was to be used on ubuntu, as seen in this code snippet:
```bash
# Ubuntu must be 22.10+ (GLIBC 2.36+) to be compatible with widevine
  if printf '%s\n' "22.10" "$__os_release" | sort -CV; then
    echo 'Package: chromium*
Pin: release o=Pi-Apps-Coders chromium-debs
Pin-Priority: 1000
Package: libwidevinecdm0
Pin: release o=Pi-Apps-Coders chromium-debs
Pin-Priority: 500
Package: chromium*
Pin: release o=Ubuntu*
Pin-Priority: -1' | sudo tee /etc/apt/preferences.d/chromium >/dev/null
```
But now I see your wildcard character `*` which explains everything. Is there a reason you did not specify `chromium-browser*` instead?

apps/Better Chromium/uninstall
Botspot: Sounds good.

theofficialgman:
> But now I see your wildcard character `*` which explains everything. Is there a reason you did not specify `chromium-browser*` instead?

Yes. There are other packages that need to be included. Refer to the apt repo https://github.com/Pi-Apps-Coders/chromium-debs/releases/tag/apt-release

Botspot:
`chromium-*` then?

Botspot:
Also, surely you have ruled out the option for directly using the https://archive.raspberrypi.com/debian/ repo on ubuntu but restricting it to only offer the chromium packages, right?

theofficialgman:
> Also, surely you have ruled out the option for directly using the https://archive.raspberrypi.com/debian/ repo on ubuntu but restricting it to only offer the chromium packages, right?

@Botspot yes, that was previously done, see https://github.com/Botspot/pi-apps/pull/2529/commits/92a7aadeda55db505ac04c4ff999a6abb359b3d5 . I ruled it too dangerous if the user happened to remove the pin they could end up with a broken system. It also triggers the frankendebian check so that is another problem with using it.

theofficialgman:
> `chromium-*` then?

yeah that is fine as well. The pi-apps-coders chromium-debs apt repos contents are fixed so the exact pin format does not really matter. 
https://github.com/Pi-Apps-Coders/chromium-debs/blob/3845b4f2785e8058ea2e12f06f14e2bcb5f1ac9f/download-chromium.sh#L3-L11

theofficialgman:
I have tested every iteration of this. I think it is all working as intended and suggested here https://github.com/Botspot/pi-apps/issues/2528#issuecomment-1894625374

theofficialgman:
@Botspot is anything else needed here?

apps/Better Chromium/uninstall
```diff
@@ -1,5 +1,16 @@
 #!/bin/bash
 
+if [[ "$1" != "update" ]];then
+  # rpi ubuntu (22.04+) can use Pi OS packaged chromium for better hardware support
+  if [ -f /etc/apt/sources.list.d/pi-apps-coders-chromium.sources ]; then
+    warning "When installing Better Chromium, you switched to PiOS chromium build. If you want to revert to the snap version, here's commands to do that."
+    warning '~/pi-apps/api rm_external_repo "pi-apps-coders-chromium" "force"'
+    warning 'sudo rm -f /etc/apt/preferences.d/pi-apps-coders-chromium'
+    warning 'sudo apt install --only-upgrade chromium-*'
+    warning 'sudo apt purge libwidevinecdm0'
```
Botspot, line 10: Do you intend for these 5 lines to all be preceded by a flashing yellow triangle and "WARNING: "?

Botspot:
Looks good. Merge at will.

PR #2540, opened 2024-01-28
CI: commit directly to repo and create/update issue on failures

theofficialgman:
ready for merge after implementing background updates of app refreshes and script updates.

PR #2541, opened 2024-01-29
updater: add `update_now_background` function and use it in `onboot` …

theofficialgman:
…updater

use same system as `list_updates_gui` to differentiate app types and filter out all apps but app refreshes

prerequisite for https://github.com/Botspot/pi-apps/pull/2540

### DIFF
--- a/etc/setting-params/Check for updates
+++ b/etc/setting-params/Check for updates
@@ -1,4 +1,4 @@
-#How often should Pi-Apps check for updates on boot?
+#How often should Pi-Apps check for app updates and refresh Pi-Apps on boot?
 Daily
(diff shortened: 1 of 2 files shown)


theofficialgman:
I decided keeping the refresh feature as part of the "Check for updates" option itself rather than having another separate option in settings for it. If you think it would be better as a separate option that can be done, just let me know.

theofficialgman:
@Botspot any objections? I would like to merge. Seems to work well and seems pretty harmless.

Botspot:
I have reviewed the files and after clarifying some comments this looks good to go.

PR #2554, opened 2024-02-24
Fully defined apt repository package checking

theofficialgman:
resolves https://github.com/Botspot/pi-apps/commit/27f5fe11a3efafc4b07b46178d115ac031558183#r139018751

these modified and new functions fully define the checking of an apt repository for installed packages.

This prevents false positives that did occur before if multiple apt repositories shared a URI and SUITE but had different COMPONENTS

This also greatly improves support for `.sources` files where multiple stanzas are now properly supported and parsed

### DIFF
--- a/api
+++ b/api
@@ -183,47 +183,6 @@ package_is_new_enough() { #check if the $1 package has an available version grea
   fi
 }
 
-anything_installed_from_repo() { #Given an apt repository URL, determine if any packages from it are currently installed
-  [ -z "$1" ] && error "anything_installed_from_repo: A repository URL must be specified."
-  
-  #user input repo-url. Remove 'https://', and translate '/' to '_' to conform to apt file-naming standard, with trailing _ to ensure full matches
-  local url="$(echo "$1" | sed 's+.*://++g' | tr '/' '_')_"
-  
-  #find all package-lists pertaining to the url
-  local repofiles="$(ls /var/lib/apt/lists/*_Packages | grep -F "$url")"
-  
-  #for every repo-file, check if any of them have an installed file
-  local found=0
-  local IFS=$'\n'
-  local repofile
-  for repofile in $repofiles ;do
-    #search the repo-file for installed packages
-    
-    grep '^Package' "$repofile" | awk '{print $2}' | while read -r package ;do
-      if package_installed "$package" ;then
-        #this package is installed; check if the version available on this repo is the current version (prevents false positives from backports repos)
-        if [ "$(sed -n "/^Package: ${package}$/,/Version:/p" "$repofile" | grep '^Version: ' | awk '{print $2}')" == "$(package_installed_version "$package")" ];then
-          echo "Package installed: $package"
-          exit 1
-        fi
-      fi
-    done #if exit code is 1, search was successful. If exit code is 0, no packages from the repo were installed.
-    
-    found=$?
-    
-    if [ $found == 1 ];then
-      break
-    fi
-  done
-  
-  #return an exit code
-  if [ $found == 1 ];then
-    return 0
-  else
-    return 1
-  fi
-}
-
 anything_installed_from_uri_suite_component() { #Given an apt repository uri, suite, and component, determine if any packages from it are currently installed
   local uri="$1"
   local suite="$2"
--- a/api
+++ b/api
@@ -225,51 +225,42 @@ anything_installed_from_repo() { #Given an apt repository URL, determine if any 
 }
 
 anything_installed_from_uri_suite_component() { #Given an apt repository uri, suite, and component, determine if any packages from it are currently installed
-  [ -z "$1" ] && error "anything_installed_from_uri_suite_component: A repository uri must be specified."
-  [ -z "$2" ] && error "anything_installed_from_uri_suite_component: A repository suite must be specified."
+  local uri="$1"
+  local suite="$2"
+  component="$3" #can be left blank
+  [ -z "$uri" ] && error "anything_installed_from_uri_suite_component: A repository uri must be specified."
+  [ -z "$suite" ] && error "anything_installed_from_uri_suite_component: A repository suite must be specified."
   
-  #component is an optional specification
-  if [ -z "$3" ]; then
-    local filepath="/var/lib/apt/lists/$(echo "$1" | sed 's+.*://++g' | sed "s,/$,," | tr '/' '_')_$(echo "$2" | sed "s,/$,," | tr '/' '_')_"
+  #find part of path to apt list file(s) to search for
+  if [ -z "$component" ]; then
+    local filepath="/var/lib/apt/lists/$(echo "$1" | sed 's+.*://++g' | sed "s,/$,," | tr '/' '_')_$(echo "$suite" | sed "s,/$,," | tr '/' '_')_"
   else
-    local filepath="/var/lib/apt/lists/$(echo "$1" | sed 's+.*://++g' | sed "s,/$,," | tr '/' '_')_dists_$(echo "$2" | sed "s,/$,," | tr '/' '_')_$(echo "$3" | sed "s,/$,," | tr '/' '_')_"
+    local filepath="/var/lib/apt/lists/$(echo "$1" | sed 's+.*://++g' | sed "s,/$,," | tr '/' '_')_dists_$(echo "$suite" | sed "s,/$,," | tr '/' '_')_$(echo "$component" | sed "s,/$,," | tr '/' '_')_"
   fi
   debug $filepath
   
   #find all relevant package-lists
-  local repofiles="$(ls $filepath*Packages)"
+  local repofiles="$(ls ${filepath}*_Packages)"
   debug "$repofiles"
   
   #for every repo-file, check if any of them have an installed file
-  local found=0
   local IFS=$'\n'
   local repofile
+  local installed_packages="$(grep -xF 'Status: install ok installed' /var/lib/dpkg/status -B 2 | grep '^Package: ' | sed 's/^Package: //g' | sort)"
   for repofile in $repofiles ;do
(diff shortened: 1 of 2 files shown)


theofficialgman:
just some examples where the output before is wrong
```
gman@gman-jam:~$ anything_installed_from_repo https://Pi-Apps-Coders.github.io/box64-debs/debian
gman@gman-jam:~$ echo $?
1
```
no output, there should be `debian-tegrax1`

correct output now:
```
gman@gman-jam:~$ anything_installed_from_uri_suite_component https://Pi-Apps-Coders.github.io/box64-debs/debian/ ./
gman@gman-jam:~$ echo $?
0
```
`0` is a package was found

theofficialgman:
there were also false negatives. I had sublime text and sublime merge installed. I then uninstalled sublime text and it removed the shared apt repo while sublime merge apt package was still installed.

all these issues are fixed with this new function and its use

api
```diff
@@ -29,6 +29,10 @@ status_green() { #announce the success of a major action
   echo -e "\e[92m$1\e[0m" 1>&2
 }
 
+debug() { #an echo command that only runs when debug mode is on
+  [ "$pi_apps_debug" = true ] && echo "$1"
+}
+
```
theofficialgman, line 35: I am fine if we want to remove this function. I thought it was just nice to have rather than commenting out `echo` commands that we have littered throughout our scripts.

api
```diff
@@ -179,45 +183,39 @@ package_is_new_enough() { #check if the $1 package has an available version grea
...
   #for every repo-file, check if any of them have an installed file
-  local found=0
   local IFS=$'\n'
   local repofile
   for repofile in $repofiles ;do
     #search the repo-file for installed packages
+
+    local packages_in_repo="$(grep '^Package' "$repofile" | awk '{print $2}' | sort)"
+    local installed_packages="$(apt list --installed 2>/dev/null | tail -n +2 | awk -F/ '{print $1}')"
+    local apt_cache_policy_output="$(echo "$packages_in_repo" | list_intersect "$installed_packages" | tr '\n' ' ' | xargs -r apt-cache policy)"
```
Botspot, line 210: With tr used here, it seems to me like the output from this would always be a single line, meaning all matching packages would be passed to apt-cache in one run.
So it seems like the only use for xargs is shorthand to avoid running apt-cache if no packages from the repo are installed?

Botspot:
I have changed the how the list of installed packages is determined, directly using the dpjkg status file which is much faster.
This should be safe because on my system the original and new command return exactly the same output.
```
pi@raspberrypi:~ $ grep -xF 'Status: install ok installed' /var/lib/dpkg/status -B 2 | grep '^Package: ' | sed 's/^Package: //g' | sort | sha256sum
3ac304de04b498fe60eaaa74122744f08c5318c96eea28523fa860d16bfc40ca  -
pi@raspberrypi:~ $ apt list --installed 2>/dev/null | tail -n +2 | awk -F/ '{print $1}' | sort | sha256sum
3ac304de04b498fe60eaaa74122744f08c5318c96eea28523fa860d16bfc40ca  -
```
The `-B2` is necessary instead of `-B1` for the handful of packages that add `Essential: yes` between the `Package:` line and the `Status: install ok installed` line.

Botspot:
Also avoided finding the list of installed packages multiple times. Not sure why you had that inside the loop.

theofficialgman:
> Also fixed setting `installed_packages` multiple times. Not sure why you had that inside the loop.

that was just an oversight. no other reason

theofficialgman:
> After having trouble getting the outputs to match and finding out why, I realized that `-B2` is necessary instead of `-B1` for the handful of packages that add `Essential: yes` between the `Package:` line and the `Status: install ok installed` line. Meaning that our `package_installed()` function has been flawed this whole time.

Yup I discovered that yesterday as well before opening this PR. I intended to fix it in pi-apps but hadn't gotten around to it yet. I see you have already done so.

api
theofficialgman: apt-cache policy needs the input as individual arguments. it also does not read from stdin. these are the reasons why xargs is used.

theofficialgman:
> I have changed the how the list of installed packages is determined, directly using the dpkg status file which is much faster. This should be safe because on my system the original and new command return exactly the same output.

I may change this back. The intention was to only get output for `all` and the current dpkg architecture. Currently any architecture package (including a non-native architecture package) will return. In order to obtain the architecture information from dpkg, we would need to parse 8 lines for each package, instead of the current 2.

There is actually (at least one) case where the output from the grep command and the apt command differs once "correctly" sorted for architecture

using your dpkg command, it now becomes
```
grep -xE 'Architecture: ((arm64)|(all))$' /var/lib/dpkg/status -B 7 | grep -xF 'Status: install ok installed' -B 2 | grep '^Package: ' | sed 's/^Package: //g' | sort
```
using the apt-cache policy, it now becomes
```
apt list --installed 2>/dev/null | grep -E "/* * amd64|all " | awk -F/ '{print $1}' | sort
```
the one difference between the output is the dpkg command incorrectly filters out any manually held back packages while the apt command does not. I am currently holding back the `qemu-user-static` package on my main desktop linux install because of QEMU regressions that cause ARM64 segfaults https://gitlab.com/qemu-project/qemu/-/issues/1913. The output of such a held back package is `Status: hold ok installed`

to correct that and achieve the same result, now the command needs to be
```
grep -xE 'Architecture: ((arm64)|(all))$' /var/lib/dpkg/status -B 7 | grep -xE 'Status: ((install)|(hold)) ok installed$' -B 2 | grep '^Package: ' | sed 's/^Package: //g' | sort
```
now that is getting quite lengthy

I feel like the dpkg status grep, although faster and I knew that, might be prone to breakage in the future.

api
Botspot: I explained poorly.
If apt-cache is only being run once, then something like this should work if I understand this correctly
```bash
local apt_cache_policy_output="$(apt-cache policy $(echo "$packages_in_repo" | list_intersect "$installed_packages" | tr '\n' ' '))"
```

Botspot:
> I feel like the dpkg status grep, although faster and I knew that, might be prone to breakage in the future.

ok understand

api
theofficialgman: No I correctly interpreted what you said.

> apt-cache policy needs the input as individual arguments.

What you just sent puts all output as one input argument I think. Can't check now.

Botspot: No quotes around the inner `$()` will make each package into a separate argument.
If nothing else, this is simply something useful to know. I'm not necessarily requesting any changes here.

api
```diff
@@ -283,9 +283,9 @@ remove_repofile_if_unused() { #Given a sources.list.d file, delete it if nothing
       # determine what uri, suite, and components are in a file
       #each stanza can only have one matching section. if there are multiple matches the last one is used
       #case should be ignored for fields
-      local uri="$(sed -n "$line_start","$line_end"p  "$file" | grep -v '^#' | grep -i '^URIs: ' | sed 's/URIs: //Ig' | tail -1)"
-      local suites="$(sed -n "$line_start","$line_end"p "$file" | grep -v '^#' | grep -i '^Suites: ' | sed 's/Suites: //Ig' | tail -1)"
-      local components="$(sed -n "$line_start","$line_end"p "$file" | grep -v '^#' | grep -i '^Components: ' | sed 's/Components: //Ig' | tail -1)"
+      local uri="$(sed -n "$line_start","$line_end"p  "$file" | grep -v '^#' | grep -i '^URIs:' | sed 's/URIs://Ig' | awk '{print $1}' | tail -1)"
+      local suites="$(sed -n "$line_start","$line_end"p "$file" | grep -v '^#' | grep -i '^Suites:' | sed 's/Suites://Ig' | awk '{$1=$1};1' | tail -1)"
+      local components="$(sed -n "$line_start","$line_end"p "$file" | grep -v '^#' | grep -i '^Components:' | sed 's/Components://Ig' | awk '{$1=$1};1' | tail -1)"
```
Botspot, line 288: Why awk print 1 on the first line but something else for the next 2?

theofficialgman: see the commit message from https://github.com/Botspot/pi-apps/pull/2554/commits/3f3b3692b6ef4c415c7fc5f184a0083d620dc4cb

Botspot: > see the commit message from [3f3b369](https://github.com/Botspot/pi-apps/commit/3f3b3692b6ef4c415c7fc5f184a0083d620dc4cb)

I already read that. What I meant was, why does line 286 use `awk '{print $1}'` but lines 287 and 288 use `awk '{$1=$1};1'`?

Botspot: It would seem to me that `awk '{$1=$1};1'` could be used for line 286, and if not, you likely have a reason why which I do not see stated clearly anywhere.

theofficialgman: > I already read that. What I meant was, why does line 286 use `awk '{print $1}'` but lines 287 and 288 use `awk '{$1=$1};1'`?

see again:
> components and suites that can have multiple strings of output

meanwhile uri can only have one string of output. this is defined by the specification referenced in a comment https://github.com/Botspot/pi-apps/pull/2554/files#diff-14c2529eb4498c5d1ffd6915d05bf58a91bdda796af59f41d480d11c099d0479R264 and expanded upon in an explanation of the specification https://repolib.readthedocs.io/en/latest/deb822-format.html#uris

theofficialgman: ah hmm... looks like I missed something in the spec. it appears actually that `URIs:` can have multiple uris defined in one field. I guess that explains why it is plural. it will need a loop as well.

> The URI must specify the base of the Debian distribution archive, from which APT finds the information it needs. There must be a URI component present in order for the source to be valid; multipls URIs can be configured simultaneously by adding a space-separated list of URIs.

I will fix that later.

Botspot:
OK once you have `uris` sorted out I think this is ready to be merged.

PR #2558, opened 2024-03-01
rename `read_packages_file` to `pkgapp_packages_required` to better suite function

theofficialgman:
Replacement to https://github.com/Botspot/pi-apps/pull/2557

### DIFF
--- a/api
+++ b/api
@@ -123,9 +123,9 @@ package_installed() { #exit 0 if $1 package is installed, otherwise exit 1
 
 package_available() { #determine if the specified package-name exists in a local repository for the current dpkg architecture
   local package="$1"
-  local arch="$(dpkg --print-architecture)"
+  local dpkg_arch="$(dpkg --print-architecture)"
   [ -z "$package" ] && error "package_available(): no package name specified!"
-  local output="$(apt-cache policy -qq "$package":"$arch" | grep "Candidate:")"
+  local output="$(apt-cache policy "$package":"$dpkg_arch" | grep "Candidate:")"
   if [ -z "$output" ]; then
     return 1
   elif echo "$output" | grep -q "Candidate: (none)"; then
@@ -1468,13 +1468,13 @@ app_type() { #there are 'standard' apps, and there are 'package' apps - an alias
   #if neither conditional above evaluated to true, no output will be returned and the function exits with code 1
 }
 
-read_packages_file() { #Returns which packages are to be installed from a package-app - handle the '|' separater
-  #if all packages are not available that are needed to be installed, returns no output
+pkgapp_packages_required() { #Returns which packages are required during installation of a package-app - handle the '|' separater
+  #returns no output if not all required packages are available
   local app="$1"
-  [ -z "$app" ] && error "read_packages_file(): no app specified!"
+  [ -z "$app" ] && error "pkgapp_packages_required(): no app specified!"
   
   if [ ! -f "${DIRECTORY}/apps/$app/packages" ];then
-    error "read_packages_file(): This app '$app' does not have a packages file!"
+    error "pkgapp_packages_required(): This app '$app' does not have a packages file!"
   fi
   
   local IFS=' '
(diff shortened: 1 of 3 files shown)


Botspot:
I mean... if you want to rename the function we could do that.
But please still look at https://github.com/Botspot/pi-apps/pull/2557 as it does a lot more than tweak the output of `read_packages_file` in edge cases.

theofficialgman:
additionally solves:

- If the user manually tries installing a hidden package-app using manage, the error from apt will be helpful, rather than saying no packages were specified.

theofficialgman:
additionally solves:

- if both OR packages are available but the second one is installed and not the first, show as installed instead of uninstalled

theofficialgman:
I do not think we can go ahead with this PR (or the other one) unless this problem is solved
```
test1(){
> test2(){
> echo hi
> }
> }
gman@gman-jam:~/pi-apps$ test1
gman@gman-jam:~/pi-apps$ test2
hi
```
https://github.com/Botspot/pi-apps/pull/2558/commits/f4076613acfd2ab6b483484cb4143db145128cbd
functions are not local and there appears to be no way in bash to make them so. the `package_available` and `package_installed` functions get replaced after using the function that re-delares them

Botspot:
I will check for solutions, but there is always the subshell option

theofficialgman:
> I will check for solutions, but there is always the subshell option

yeah I think that is what we will end up needing to do
define the functions and run refresh_pkgapp_status in a subshell together

Botspot:
Good catch
```
pi@raspberrypi:~ $ DIRECTORY=$HOME/pi-apps
pi@raspberrypi:~ $ source ~/pi-apps/api
pi@raspberrypi:~ $ package_available chromium-browser
pi@raspberrypi:~ $ echo $?
0
pi@raspberrypi:~ $ refresh_all_pkgapp_status
Marking Pinta as hidden
Marking Steam Link as hidden
pi@raspberrypi:~ $ package_available chromium-browser
pi@raspberrypi:~ $ echo $?
1
```

theofficialgman:
I went with the subshell method. doesn't seem to hurt performance. if you find anything else that you like better feel free to implement

Botspot:
I thought of saving original values of the functions and setting them back, but a subshell seems more straightforward and safe.

Botspot:
I'm thinking of doing a squash-merge unless you want to do another force push to cut down on commit count.

theofficialgman:
I'll rewrite the history tomorrow so it makes more sense and merge it (with no actual code changes besides maybe adding `-qq` since this doesn't have it).

I'd rather leave this tonight so I can sleep on it and then check over again tomorrow.

theofficialgman:
also I just want to make it clear. I liked most of the changes from your PR (which is why I integrated them here, anything you made I'll make sure has a coauthor on it). There were just some things I had issues with.

Clearly our best work is when its done together or iterating on each other.

Botspot:
> also I just want to make it clear. I liked most of the changes from your PR (which is why I integrated them here, anything you made I'll make sure has a coauthor on it). There were just some things I had issues with.

Understand.

> Clearly our best work is when its done together or iterating on each other.

For sure. Pi-Apps would just be another PiKISS if it was not for teamwork.

Botspot:
> * if both OR packages are available but the second one is installed and not the first, show as installed instead of uninstalled

I just thought of a scenario where your solution to this is likely insufficient.
- Imagine I am on a system where both `chromium` and `chromium-browser` are available.
- The packages file contains `chromium--browser | chromium` as usual.
- `chromium` is installed.
- pi-apps needs to update Chromium (it should never need to do this - this is theoretical)
- Pi-Apps uninstalls chromium because `pkgapp_packages_required` told apt to purge `chromium`.
- Pi-Apps installs back `chromium-browser` because `pkgapp_packages_required` now uses the first available package in the OR list as neither one are installed at the time of its second execution.

theofficialgman:
> > * if both OR packages are available but the second one is installed and not the first, show as installed instead of uninstalled
> 
> I just thought of a scenario where your solution to this is likely insufficient.
> 
> * Imagine I am on a system where both `chromium` and `chromium-browser` are available.
> * The packages file contains `chromium-browser | chromium` as usual.
> * `chromium` is installed.
> * pi-apps needs to update Chromium (it should never need to do this - this is theoretical)
> * Pi-Apps uninstalls chromium because `pkgapp_packages_required` told apt to purge `chromium`.
> * Pi-Apps installs back `chromium-browser` because `pkgapp_packages_required` now uses the first available package in the OR list as neither one are installed at the time of its second execution.

your scenario hinges on this `pi-apps needs to update Chromium` which as you say cannot happen. it is easy to break many of the functions of pi-apps with output or actions that cannot occur.

if it is desirable that updates to package apps retain the previously installed packages (assuming those packages are still in the new packages file), then that should be handled in the updater for package apps, not the `pkgapp_packages_required` call. that means modifying `update_app` which is ends up being called for updating apps via all methods (gui, manage, cli, etc):
https://github.com/Botspot/pi-apps/blob/2c327832366bc7685fd56df860e18371b01a6cd1/updater#L436-L488

alternatively, you could modify `manage uninstall` to not uninstall package apps when triggered as part of an `update` (reminder that the 3rd argument `update` is passed to manage when triggered as part of an update). instead of uninstalling, mark the packages as automatically installed (instead of manually installed which is their current status). then, during the `manage install`, the packages will still be installed (but will get autoremoved during any `apt autoremove` command). that will result in `pkgapp_packages_required` returning the desired `chromium` output since it is still installed. `manage install` can continue as normal and (optionally) you could add an `apt autoremove` (or add `--autoremove` as a flag during the `apt install`) to remove any packages that were marked as automatically installed during `manage uninstall` that were not marked as manually installed during `manage install` (eg: like would be desirable for a package that got removed from the `packages` list during the update)

https://github.com/Botspot/pi-apps/blob/2c327832366bc7685fd56df860e18371b01a6cd1/manage#L596-L742

PR #2593, opened 2024-05-20
Tidy up settings script

trn1ty:
I noticed there was some tricky IFS usage to separate file names, but shell can do that in a for loop anyway, so I went ahead and cleaned all that up as well as making some other small changes for readability. Functionality should be equivalent except that `settings help` shows a help screen now.

### DIFF
--- a/settings
+++ b/settings
@@ -36,6 +36,14 @@ revert) # Overwrite all settings with the defaults.
   done
   exit 0
   ;;
+*)
+  printf "\
+Usage: %s (subcommand)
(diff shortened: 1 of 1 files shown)


Botspot:
`echo` and `printf` are both bash builtins, so I am not aware of any meaningful performance difference between the two.
So when you use printf just like echo, is there a reason for that?

Why exchange this:
`echo "$output"`
for this?
`printf '%s\n' "$output"`

theofficialgman:
Closing as stale as the author has not replied to comments.

PR #2594, opened 2024-05-20
Tidy up add_to_list in manage

trn1ty:
While looking at the add_to_list function in manage I noticed the function could probably be implemented as a single shell pipeline, so I went ahead and did that. Instead of bouncing local variables around and using Bash features to sculpt everything before spitting it out, this leans on sed(1p) to chip away at a stream as it slowly forms into the desired output.

### DIFF
--- a/manage
+++ b/manage
@@ -218,26 +218,19 @@ if [ "$1" == 'daemon' ];then
   #this folder will contain the runtime files necessary to make the daemon work.
   mkdir -p "${DIRECTORY}/data/manage-daemon"
   
-  #a list of pending steps is displayed with yad. This function adds an item to the list
-  add_to_list() { # $1 is action, $2 is app, $3 is exit code status ('0' = success), $4 is output location (usually blank, but can be 'stdout')
+  # This function adds an item to the list of pending actions displayed by yad.
+  #   $1 - action
+  #   $2 - app
+  #   $3 - exit status ('0' is success)
+  #   $4 - output location ('stdout' if standard output, otherwise the list)
+  add_to_list() {
     # $action is used for status messages, so trim off the -file if it's there.
     test "$1" = update-file \
       && action=update \
       || action="$1"
-    local app="$2"
-    local status="$3"
     
-    # $4 specifies where the output should go.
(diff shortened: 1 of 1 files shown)


Botspot:
What is the goal of this PR?

- performance
- code readability
- code robustness
- future maintainability
- minimizing lines of code
- minimizing external program calls

To me this may be more "tidy", but my gut reaction is that it would perform more slowly as it seems to be using sed more and internal shell builtins less. (echo, test, printf, eval are built-in to bash and are generally much faster than repeated calls to external programs like sed, awk, or grep)
Also we have inconsistent usage of the input variables throughout the function, sometimes using `$action`, other times using `$1`. Not sure what that does except force a developer to memorize what `$1`, `$2`, `$3`, and `$4` mean to comprehend what is going on here.

Botspot:
Regarding code robustness, I see a few places where sed is given an app name as an argument. This could be problematic for app names with special characters in them. These are the special characters that some app names are using now:
`(` `)` `.` `+` `-` `'`
These app names could cause sed to perform strange regex edge cases. This all would need to be rigorously tested.

theofficialgman:
I find these changes overall much harder to read and they make it more difficult to extend the functionality of the function while "seemingly" providing no other benefit.

If you could respond to botspots first comment that would help.

theofficialgman:
still need this answered:

What is the goal of this PR?

- performance
- code readability
- code robustness
- future maintainability
- minimizing lines of code
- minimizing external program calls

theofficialgman:
> Also we have inconsistent usage of the input variables throughout the function, sometimes using `$action`, other times using `$1`. Not sure what that does except force a developer to memorize what `$1`, `$2`, `$3`, and `$4` mean to comprehend what is going on here.

To add to this, we generally prefer to rename the input arguments on purpose so that developers don't need to memorize the input argument order. If you look at most of the other functions in pi-apps where an input argument is used multiple times in the context of a function, we do this input argument renaming.

It is also our standard to define the use of the function in only one line before the function or inline with the function name with the input arguments defined within the function context, not before like you have done here.

This PR really only changes this function to not abide by these standards while providing no other value. For now I will close it.

PR #2595, opened 2024-05-20
WIP: Tidy up preload script

trn1ty:
This is a set of optimizations and cleanups for preload.

### DIFF
--- a/preload
+++ b/preload
@@ -148,7 +148,7 @@ if [ $reloadlist == 1 ];then
   if ! command -v app_prefix_category >/dev/null ;then
     source "${DIRECTORY}/api"
   fi
-  
+
   vfiles="$(
     app_prefix_category "$prefix" \
       | grep . \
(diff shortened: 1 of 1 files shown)


theofficialgman:
the if/else conversion to case statements are fine but I fail to see how anything else is an improvement... it looks harder to read/understand and is multiple lines longer.

theofficialgman:
also shellcheck shows a syntax error. refer to the action log

preload
```diff
@@ -37,25 +37,18 @@ function error {
   exit 1
 }
 
-#yad or xlunch format
+case "$1" in
+source)  return 0 ;;
```
theofficialgman, line 41: The `preload` script is not to be sourced. Why has this input argument been added?

preload
```diff
@@ -37,25 +37,18 @@ function error {
...
-
-#specifies a category to preload (can be left empty to load main page)
-prefix="$2"
-
-if [ "$prefix" == '/' ];then
-  prefix=
-fi
+# The prefix is a category to preload. By default (if '/' or empty) the main
+# page is preloaded.
+test "$2" = / || prefix="$2"
```
theofficialgman, line 51: this breaks the prefix variable and thus the preload script, it is no longer set when it is not `/`

theofficialgman:
Closing as stale as the author has not implemented requested changes and the if statement to case conversion makes no functional difference (we can implement that whenever if we wanted).

PR #2608, opened 2024-06-19
add VARA HF app

WheezyE:
I propose adding VARA HF to pi-apps.  

VARA HF is a small VB6 app with a large following. It is a software modem used by amateur radio enthusiasts to send data over radio transceivers. Many amateur ("ham") radio enthusiasts wish to use their radios outside on battery power and thus require small, power-efficient, computers like the Raspberry Pi to run these data-to-sound modulations-demodulations through their radio transceivers.

Many "hams" are also more hardware savvy than tech savvy and often need a little help installing things like wine, box86, and apps on their Raspberry Pi. I have run my own script, [Winelink](https://github.com/WheezyE/Winelink), for a few years with good success stably running VARA over the years. My main issue is maintaining the complex Wine installation in my script. I plan to continue to maintain my script as an experimental arm of installing ham apps on the Pi, but I would deem VARA 'stable' enough to go prime-time on pi-apps.

There are other VARA apps as well which could be added later or bundled into this app install (VARA Chat, VARA FM, VARA Sat), but VARA HF is the most sought-after by hams and should suffice on its own.


WheezyE:
It's worth noting there could be a way to run this script without autohotkey if we could determine with certainty which wineprefix the user was running.

apps/VARA HF/uninstall
```diff
@@ -0,0 +1,17 @@
...
+
+#Removing program files (keeping any VARA.ini settings files)
+mv ${WINEDIR}/drive_c/VARA/VARA.ini /tmp/VARA.ini 2>/dev/null
+rm -rf ${WINEDIR}/drive_c/VARA/* 2>/dev/null || error 'Could not remove VARA program files.'
+mv /tmp/VARA.ini ${WINEDIR}/drive_c/VARA/VARA.ini 2>/dev/null
+if [ $(ls -A "${WINEDIR}/drive_c/VARA" | wc -l) -eq 0 ]; then rm -rf "${WINEDIR}/drive_c/VARA" 2>/dev/null; fi #also delete directory if empty
+
+#Removing Desktop Entry
+rm ${APPDIR}/VARA/varahf.desktop 2>/dev/null || echo "Could not remove menu button."
+if [ $(ls -A "${APPDIR}/VARA" | wc -l) -eq 0 ]; then rm -rf "${APPDIR}/VARA" 2>/dev/null; fi #also delete directory if empty
```
Botspot, line 17: You can use `rmdir` for this.

apps/VARA HF/install
```diff
@@ -0,0 +1,66 @@
...
+sudo apt-get install -y p7zip-full || error 'Failed to install p7zip-full.'
+7z e -y -bsp0 -bso0 ${AHKDIR}/AutoHotkey_1.1.36.02_setup.exe AutoHotkeyU32.exe -o"${VARAHFDIR}" || error 'Failed to extract AutoHotKey archive.'
+echo -e "\nInstalling VARA HF . . ."
+BOX86_DYNAREC=0 BOX86_NOBANNER=1 BOX86_DYNAREC_BIGBLOCK=0 WINEDEBUG=-all wine ${VARAHFDIR}/AutoHotkeyU32.exe ${VARAHFDIR}/varahf_install.ahk || error 'AutoHotKey failed to launch.'
+
+#Removing tmp files
+rm -rf ${AHKDIR} ${VARAHFDIR}
+
+#Adding the user to the USB dialout group so that they can access radio USB CAT control later.
+sudo usermod -a -G dialout $USER
```
Botspot, line 49: I have to run usermod on every boot to keep a normal user in a group. Have you checked if this change is persistent across reboots?

apps/VARA HF/install
```diff
@@ -0,0 +1,66 @@
...
+# creating varahf_install.ahk (autohotkey script)
+echo '; AHK script to make VARA installer run completely silent'                                               > ${VARAHFDIR}/varahf_install.ahk
+echo 'SetTitleMatchMode, 2'                                                                                    >> ${VARAHFDIR}/varahf_install.ahk
+echo 'SetTitleMatchMode, slow'                                                                                 >> ${VARAHFDIR}/varahf_install.ahk
+echo '        Run, VARA setup (Run as Administrator).exe /SILENT /c:"VARA setup (Run as Administrator).exe"'   >> ${VARAHFDIR}/varahf_install.ahk
+echo '        WinWait, VARA Setup ; Wait for the "VARA installed successfully" window'                         >> ${VARAHFDIR}/varahf_install.ahk
+echo '        ControlClick, Button1, VARA Setup ; Click the OK button'                                         >> ${VARAHFDIR}/varahf_install.ahk
+echo '        WinWaitClose'                                                                                    >> ${VARAHFDIR}/varahf_install.ahk
+# running varahf_install.ahk
+sudo apt-get install -y p7zip-full || error 'Failed to install p7zip-full.'
```
Botspot, line 40: We ask that you not use `apt-get` directly but instead use our `install_packages` function.

apps/VARA HF/install
```diff
@@ -0,0 +1,66 @@
...
+#Downloading VARA HF
+VARAHFDIR=/tmp/VARAHF
+mkdir "${VARAHFDIR}" 2>/dev/null
+wget -q -r -l1 -np -nd "https://downloads.winlink.org/VARA%20Products" -A "VARA HF*setup.zip" -P "${VARAHFDIR}" || echo "failed to download"
+
+#Extracting VARA HF installer archives
+unzip -o "${VARAHFDIR}/VARA HF*setup.zip" -d "${VARAHFDIR}" || error 'Failed to unzip VARA HF archive'
+
+#Running VARA HF setup with autohotkey so we can auto-close its OK button
+# note: Instead of installing, we could just extract VARA's program files, copy them into wine, and manually install OCX files, but we would have to know which wineprefix the user is running.
```
Botspot, line 25: You could create a separate wineprefix for this, or use the default one pi-apps sets up at `~/.wine` if `$WINEPREFIX` is unset.

apps/VARA HF/install
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+#version=4.8.7 #version changes often
```
Botspot, line 3: Not sure if you intentionally commented out this `version=` line.

apps/VARA HF/install
```diff
@@ -0,0 +1,66 @@
...
+#Configuring wineprefix for VARA programs
+BOX86_DYNAREC=0 BOX64_DYNAREC=0 BOX86_NOBANNER=1 BOX64_NOBANNER=1 winetricks -q vb6run pdh_nt4 sound=alsa
+
+#Downloading VARA HF
+VARAHFDIR=/tmp/VARAHF
+mkdir "${VARAHFDIR}" 2>/dev/null
+wget -q -r -l1 -np -nd "https://downloads.winlink.org/VARA%20Products" -A "VARA HF*setup.zip" -P "${VARAHFDIR}" || echo "failed to download"
+
+#Extracting VARA HF installer archives
+unzip -o "${VARAHFDIR}/VARA HF*setup.zip" -d "${VARAHFDIR}" || error 'Failed to unzip VARA HF archive'
```
Botspot, line 22: It's good practice to remove files from /tmp when you are done using them.

Botspot:
As you are aware, Pi-Apps only supports GUI systems. We do not try to support headless usage or systems without a desktop environment installed. Would that be a problem here? I do know that for using a Pi in the field, more as a device than a desktop PC, many people are encouraged to use the `Lite` version of the OS.

Also could you estimate how many people would likely find this useful, if added to Pi-Apps? I am unfamiliar with the HAM crowd, and am not sure if 50 unique people using this is a lot or a little.

theofficialgman:
> It's worth noting there could be a way to run this script without autohotkey if we could determine with certainty which wineprefix the user was running.

the default location is `~/.wine`. unless the `WINEPREFIX` is set in the environment then that location will not change. So you can simply read the `WINEPREFIX` variable, if it is empty then use `~/.wine`

eg:

```
if [ -z "${WINEPREFIX}" ];
  wine prefix is unset or set to empty string. so location is `~/.wine`
else
  wine prefix is set to some location. so location is `$WINEPREFIX`
fi
```

apps/VARA HF/install
theofficialgman: @Botspot your experience is not normal. can you please show a example of a user being removed from a group on reboot?

apps/VARA HF/install
```diff
@@ -0,0 +1,66 @@
...
+  "$DIRECTORY/manage" install-if-not-installed 'Wine (x64)' || error 'Failed to install wine'
+fi
+
+#Configuring wineprefix for VARA programs
+BOX86_DYNAREC=0 BOX64_DYNAREC=0 BOX86_NOBANNER=1 BOX64_NOBANNER=1 winetricks -q vb6run pdh_nt4 sound=alsa
+
+#Downloading VARA HF
+VARAHFDIR=/tmp/VARAHF
+mkdir "${VARAHFDIR}" 2>/dev/null
+wget -q -r -l1 -np -nd "https://downloads.winlink.org/VARA%20Products" -A "VARA HF*setup.zip" -P "${VARAHFDIR}" || echo "failed to download"
```
theofficialgman, line 19: this should use `error` and not `echo`

apps/VARA HF/uninstall
WheezyE: Oh that's much easier, thank you!

apps/VARA HF/install
WheezyE: Ah yeah that was leftover from some testing thing I was doing. Thank you for spotting this!

WheezyE:
I accidentally started a code review and am not sure how to cancel it - sorry about that.

WheezyE:
> As you are aware, Pi-Apps only supports GUI systems. We do not try to support headless usage or systems without a desktop environment installed. Would that be a problem here? I do know that for using a Pi in the field, more as a device than a desktop PC, many people are encouraged to use the `Lite` version of the OS.

I believe this shouldn't be an issue since nobody has commented on it while using [my Winelink project script](https://github.com/WheezyE/Winelink/blob/1925b178cb4b65e6898e671e0797482705e93b77/install_winelink.sh#L35-L42), to install VARA HF in the past - my script also forces the user to not run in headless mode (checks for xsession).

```bash
function run_main()
{
    export WINEDEBUG=-all # silence winedbg for this instance of the terminal
    local ARG="$1" # store the first argument passed to the script file as a variable here (i.e. 'bash install_winelink.sh vara_only')


    ### Pre-installation
    run_checkpermissions
    run_checkxhost
```
...
```bash
function run_checkxhost()  # Check to see if an xserver is running (ie are we in SSH? because we will need a GUI for the script to work)
{
    if ! xhost >& /dev/null ; then # credits: xylo04 - thank you for this function!
        echo "No X window session, this script must be run with a GUI"
        run_giveup
    fi
}
```

> Also could you estimate how many people would likely find this useful, if added to Pi-Apps? I am unfamiliar with the HAM crowd, and am not sure if 50 unique people using this is a lot or a little.

That's a really good question. I'm not really sure how many people would actually be using this. I can maybe estimate somewhat based on 870 posts related to VARA on KM4ACK's message board, which started out as a ham radio on Raspberry Pi forum.
https://groups.io/g/KM4ACK-Pi/search?p=created%2C%2C%2C20%2C2%2C0%2C0&q=vara

apps/VARA HF/install
```diff
@@ -53,7 +49,7 @@ echo "[Desktop Entry]
 Name=VARA HF
 GenericName=VARA HF
 Comment=VARA HF is a shareware ham radio OFDM software modem for RMS Express and other messaging clients.
-Exec=env WINEDEBUG=-all wine \"C:\\VARA\\VARA.exe\"
+Exec=env WINEDEBUG=-all wine "${PREFIXDIR}"/drive_c/VARA/VARA.exe
```
Botspot, line 53: The way these double-quotes are escaped in the echo command, they will not show up in the .desktop file.
They will be necessary if the user's wineprefix contains whitespace characters. I don't recall if single-quote characters work, but to escape the double-quotes, replace them with this: `"\""`

apps/VARA HF/uninstall
```diff
@@ -2,12 +2,8 @@
...
 APPDIR="${HOME}"/.local/share/applications
 
-#Finding your default wineprefix directory
-if [ -z "${WINEPREFIX}" ]; then
-    PREFIXDIR="${HOME}"/.wine #if "WINEPREFIX=" envvar is unset or set to empty string, wineprefix location is "~/.wine"
-else
-    PREFIXDIR="${WINEPREFIX}" #if "WINEPREFIX=" envvar was previously set, wineprefix location is "$WINEPREFIX"
-fi
+#Finding the wineprefix where VARA.exe is installed
+PREFIXDIR=$(find ${HOME} -name "VARA.exe" | grep -P -o "(.*)(?=/drive_c/)")
```
Botspot, line 6: Beware that nothing is preventing this variable from containing multiple lines. It is up to you if you actually want to deal with that possibility though.

Botspot:
@WheezyE Let me know when you are done with the changes and ready for me to test installation on my machine.

WheezyE:
I believe I'm ready for you to test on your machine. Please let me know if you have ideas about how to handle this conundrum about knowing with certainty where the wineprefix is located. I supposed the only way for sure would be to have VARA in its own wineprefix but I would like to avoid that. _I could have the script check the .desktop button for the program too, but a user could edit this in theory and that would then wreck an uninstaller._

Botspot:
> I believe I'm ready for you to test on your machine. Please let me know if you have ideas about how to handle this conundrum about knowing with certainty where the wineprefix is located. I supposed the only way for sure would be to have VARA in its own wineprefix but I would like to avoid that. _I could have the script check the .desktop button for the program too, but a user could edit this in theory and that would then wreck an uninstaller._ The uninstall could also fail if the user deleted their wineprefix after installation - should I remove any `|| error 'messages'` if the VARA program files cannot be deleted?

- I don't like this:
    ```bash
    find "${HOME}" -name "VARA.exe"
    ```
    My home folder has some files not readable by the default user (config files, keys, etc), and it returns a lot of "Permission denied" errors, in addition to taking nearly a minute to complete.
    Maybe just use the same simple wineprefix detection as found in the install script, but if VARA is not found, use our `warning` function to tell the user that vara was not found, so if they moved it elsewhere that it needs to be manually deleted.
- Why are you installing the menu launcher to `~/.local/share/applications/VARA` rather than to `~/.local/share/applications`?
- Looks like the double-quote thing in the menu launcher still needs to be addressed.
- I would suggest you avoid obsessing over user tampering to the wineprefix location and path in the launcher. Anyone doing that likely has their reasons. But I would suggest making sure each part of the uninstall script can run multiple times.
  - For example, if the last line (`rmdir`) returns code 1, which it will on subsequent re-runs, then the whole script will return code 1 because it is the last command. To fix this, add `|| true` to the end of that line before the comment.
  - Nest your "Unregistering OCX/DLL files" lines into an `if` statement that will skip this if those files are missing.
  - Use `rm -f` for times when the file to remove is allowed to not exist. You have used this in most places, but it is missing in one place in `uninstall`.

> If things look good, I'll squash commits.

Don't worry about that; I tend to use the "Squash and merge" button.

WheezyE:
That sounds really good to me. I keep thinking of edge conditions, but I think you're right: If a user is configuring their own wineprefixes then they can also probably figure out how to remove programs manually. I'll rely on default or `WINEPREFIX=`. Been busy this week but I'll edit and test soon

WheezyE:
Ok, I think I've got these changes implemented now. Feel free to test again on your end at your leisure. Had a busy week! (got some stuff sorted out for a move overseas though, which is cool)

---

> My home folder has some files not readable by the default user (config files, keys, etc), and it returns a lot of "Permission denied" errors, in addition to taking nearly a minute to complete.
> Maybe just use the same simple wineprefix detection as found in the install script, 

That sounds good: I'm now either using default wineprefix or setting to user-defined wineprefix var.

> but if VARA is not found, use our warning function to tell the user that vara was not found, so if they moved it elsewhere that it needs to be manually deleted.

Added warnings instead of errors/echos

>Why are you installing the menu launcher to ~/.local/share/applications/VARA rather than to ~/.local/share/applications?

There are multiple VARA programs that a user could install (VARA HF, FM, Sat, Chat, Terminal). I plan to implement those later in my own script (or in Pi-Apps if you like - an easy modification from this script) and thought it would be nice to have the VARA apps icons separate from other program icons (like how wine has its own program icon folder in the ~/.local/share/applications/ directory). I can also remove this is you like.

> Looks like the double-quote thing in the menu launcher still needs to be addressed.

Oh I see what you mean now. I've added quotes to the program shortcut now for edge cases when the user's wineprefix path contains spaces.

> I would suggest you avoid obsessing over user tampering to the wineprefix location and path in the launcher. Anyone doing that likely has their reasons. But I would suggest making sure each part of the uninstall script can run multiple times.
For example, if the last line (rmdir) returns code 1, which it will on subsequent re-runs, then the whole script will return code 1 because it is the last command. To fix this, add || true to the end of that line before the comment.

I've added `|| true` now to prevent inadvertent code 1 return in cases where directory could not be found/deleted

> Nest your "Unregistering OCX/DLL files" lines into an if statement that will skip this if those files are missing.
That's not a bad idea. It will speed up the installation quite a bit if the files don't exist.

I've now implemented a for loop to check for the files (since winetricks advised me in the past that using -f on globs is a bad idea). The unregister function works fastest in wine if I feed it a glob, so if the for loop finds any files, then wine will be called and unregister any ocx files it finds in the directory.

> Use rm -f for times when the file to remove is allowed to not exist. You have used this in most places, but it is missing in one place in uninstall.

Oh I see what you mean about the program icon removal. I think it might be better to keep this and then warn the user in cases when the rm didn't work. I've updated the warning message now to alert the user if the icon can't be found. If we need to add a `|| warning '' && true` or something let me know

WheezyE:
Just wanted to circle back to this PR to see if you guys needed anything else from me.
Thank you again for your help with code reviews and tips for compatibility / format / style.

Botspot:
Do you want to remove the psapi.dll file when uninstalling?

apps/VARA HF/description
```diff
@@ -0,0 +1,8 @@
+A shareware HF modem for amateur radio messaging.
+
+VARA modems require a messaging app (such as Pat, Winlink Express, VarAC, or VARAChat), an amateur radio transceiver, and cables to connect audio and PTT control from your Pi to the transceiver.
+
+Run from Menu: Menu -> Accessories -> VARA HF
+Run from Terminal: wine "C:\VARA\VARA.exe"
+
+This app runs with the help of box86 and wine.
```
theofficialgman, line 8: Box86/Box64 and Wine

WheezyE:
> Do you want to remove the psapi.dll file when uninstalling?

That's a good question.

You know, I think I can just actually remove the `cp -n "${PREFIXDIR}"/drive_c/VARA/OCX/psapi.dll "${PREFIXDIR}"/drive_c/windows/system32/psapi.dll` line Wine already has an emulated psapi.dll that my Winelink script has always relied on - and which has worked just fine.

_More info:_
_I've just now tested this DLL vs the emulated Wine version of this DLL that's included with Wine (set to built-in and native in winecfg) and the program doesn't seem to function any differently using either one._

_The VARA HF installer also does not install this by default on Windows 11.  The DLL is also not an Active X component for VB6 (doesn't get registered by the VARA installer).  I'm assuming it doesn't get seen/invoked by the VARA.exe either on runtime since it's not in the same directory as the VARA.exe (instead it's in a subdirectory - would that be invoked as normal Windows program behavior?).  Anyway, I think it's not needed._

WheezyE:
Thanks guys!!

PR #2622, opened 2024-07-19
api: `purge_packages" don't autoremove on updates

theofficialgman:


api
```diff
@@ -734,7 +734,7 @@ purge_packages() { #Allow dependencies of the $app app to be autoremoved.
     
     status "Purging the $package_name package..."
 
-    if [[ "$script_input" == "update" ]]; then
+    if [ "$script_input" == "update" ]; then
```
theofficialgman, line 737: api has `#!/bin/bash` which guarantees bash (or equivalent implementation) execution

the use of double brackets is common in the api script. Even though there is no wildcard pattern matching being done (which would require double brackets) I see no reason why not to just use double brackets exclusively

Botspot: I will admit it is a very unimportant design preference of mine, but the way I learned bash gradually through online searches, double brackets mean regex.
Whenever I see double brackets I'm expecting to see regex, and when it's not there I end up re-reading the expression multiple times to check if I am missing any `*` characters.

Because you brought it up, I will try to drop the habit, but no promises on a timeline for that. :)

PR #2625, opened 2024-07-26
api apt exit code checking

Botspot:
Decided not to make changes to purge_packages because this already seems like quite an edge case

### DIFF
--- a/api
+++ b/api
@@ -371,8 +371,9 @@ apt_update() { #run an apt update with error-checking and minimal output
   apt_lock_wait
   
   status "Running \e[7msudo apt update\e[27m..."
-  output="$(sudo -E apt update --allow-releaseinfo-change "$@" 2>&1 | less_apt | tee /dev/stderr)"
-  exitcode=$?
+  set -o pipefail
+  local output="$(sudo -E apt update --allow-releaseinfo-change "$@" 2>&1 | less_apt | tee /dev/stderr)"
+  local exitcode=$?
   status "apt update complete."
   
   #inform user about autoremovable packages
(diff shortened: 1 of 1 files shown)


api
```diff
@@ -687,7 +692,7 @@ Package: $package_name" > /tmp/$package_name/DEBIAN/control
     done
     
     errors="$(echo "$output" | grep '^[(E)|(Err]:')"
-    if [ ! -z "$errors" ];then
+    if [ ! -z "$errors" ] || [ "$exitcode" != 0 ];then
```
theofficialgman, line 695: this will catch the issue but the debug output will not make sense.

it would be better to add an `else` case below this with custom output indicating that apt exited with an error without any `E` or `Err` output.

api
```diff
@@ -370,6 +370,8 @@ apt_update() { #run an apt update with error-checking and minimal output
   apt_lock_wait
   
   status "Running \e[7msudo apt update\e[27m..."
+  local output
+  set -o pipefail
   output="$(sudo -E apt update --allow-releaseinfo-change "$@" 2>&1 | less_apt | tee /dev/stderr)"
   exitcode=$?
```
theofficialgman, line 376: we can take the opportunity to make this local as well

api
Botspot: In this if statement, other specific messages from apt are checked for. Splitting this into an elif would mean either skipping those checks or duplicating them. Is that worth it?

theofficialgman: then simply check for the exitcode again inside if statement along with the other cases to avoid needing to duplicate them.

APT only has `0` and `100` exit codes programmed into it (manpage says so https://manpages.ubuntu.com/manpages/noble/man8/apt.8.html) . If it exits with something no equaling one of those two values it is an undefined error (like a segfault).

PR #2640, opened 2024-08-29
add MTG Forge game

Archargelod:


apps/Forge/install
```diff
@@ -0,0 +1,74 @@
...
+
+status "Extracting the game archive."
+tar -xvf "/tmp/forge-gui-desktop-${subversion}.tar.bz2" --directory="$game_folder" || error 'Failed to extract the game archive!'
+
+# problem: forge-adventure doesn't launch with java versions > 8
+# workaround: create `forge-mobile.sh` script that's a modified copy of `forge.sh` script
+status "Creating workaround for adventure shell script."
+jarname="$(find "$game_folder" -name 'forge-gui-mobile*.jar' | xargs basename)"
+cp "$game_folder/forge.sh" "$game_folder/forge-mobile.sh"
+sed -i "s|forge-gui-desktop-.*-jar-with-dependencies.jar|$jarname|" "$game_folder/forge-mobile.sh"
```
theofficialgman, line 47: What is the goal here? if the program doesn't launch with java > 8 then why are you installing java 17?

Archargelod: > with java > 8 then why are you installing java 17?

the app itself does work with java version > 8, but iirc launch scripts for "adventure" mode doesn't have proper flags so it just fails.

Why not just install java 8? Because it doesn't work with java 8 😂. Forge needs a java fx library that's not available in opensource builds of java 8, apparently it has support only for official oracle's build of java 8 and that's certainly harder to legally obtain. 

Maybe it's possible to make it work with java 8 on linux, but with my limited knowledge of java - I couldn't. But all other versions of java should be supported, I chose java 17 because most software in pi-apps use that version, so good chances that it will be already installed.

theofficialgman: Can you link an upstream bug report for that issue (or PR to fix it if one is open). If it's just bad flags it would be better to have it corrected upstream.

Archargelod: > Can you link an upstream bug report for that issue (or PR to fix it if one is open). If it's just bad flags it would be better to have it corrected upstream.

Here's my issue from 1+ year ago - https://github.com/Card-Forge/forge/issues/2221, it was marked as stale and closed. Note, that I was absolutely unfamiliar with git at that time.  
Looking at it now, I could try to submit PR that fixes the issue and we will see if it gets somewhere 😅.

PR #2658, opened 2024-10-10
api: replace here-string in `log_diagnose` with process substitution

theofficialgman:
a here-string creates a temporary file on disk which can lead to failure to process all greps in the case of disk space exhaustion. a here-string is also subject to max argument length exhaustion on some systems/kernels leading to failure to process all greps.

process substitution occurs in memory and is not subject to the same limitations.

### DIFF
--- a/api
+++ b/api
@@ -1952,7 +1952,7 @@ log_diagnose() { #Given a logfile, explain errors to user, suggest fixes, and ca
 #repo issues below
 #------------------------------------------
   #check for 'E: The repository'
-  if grep -qF 'E: The repository' <<<"$errors" || grep -qF 'sources.list entry misspelt' <<<"$errors" || grep -qF 'component misspelt in' <<<"$errors" ;then
+  if grep -qF 'E: The repository' <(echo "$errors") || grep -qF 'sources.list entry misspelt' <(echo "$errors") || grep -qF 'component misspelt in' <(echo "$errors") ;then
     error_caption+=("APT reported a faulty repository, and you must fix it before Pi-Apps will work.
 
 To delete the repository:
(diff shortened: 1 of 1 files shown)


Botspot:
> a here-string creates a temporary file on disk

Are you sure?
20241009_21h20m24s_grim

api
```diff
@@ -2858,6 +2858,8 @@ If this error keeps happening, try installing the More RAM app from Pi-Apps. Fin
   #return the information
   local IFS=$'\n'
   echo -e "${error_type}\n${error_caption[*]}"
+
+  rm -f "$tmplogfile"
```
Botspot, line 2862: Where is this variable set? I don't see it being assigned anywhere.

theofficialgman: Sorry. Artifact of a previous refactor attempt.

theofficialgman:
> > a here-string creates a temporary file on disk
> 
> Are you sure?

See top comment https://askubuntu.com/questions/678915/whats-the-difference-between-and-in-bash

theofficialgman:
Also just an FYI regarding ChatGPT and its use for specific technical questions. The use of ChatGPT will not give you consistent good/accurate results. It is a large language model, aka, complicated autocomplete. The question you asked was very leading and didn't provide any background information for it to "attempt" to autocomplete an answer for. It wasn't trained to get the answer correct but instead to answer in a way that sounded "natural". It is unreasonable for you to ask a technical question like that and expect an accurate answer.

Also to provide justification from my statement beyond the askubuntu answer I checked the bash code and quickly found what is responsible for this. The short of it is except in a few specific cases (zero length and short files) a temporary file is made.

https://github.com/bminor/bash/blob/6794b5478f660256a1023712b5fc169196ed0a22/redir.c#L433-L578

Botspot:
Merge at will.

Issue #2669, opened 2024-10-26
Hangover


AndreRH:
### What is the name of the app?

Hangover

### Where is the app hosted?

https://github.com/AndreRH/hangover

Debian packages are attached to the release: https://github.com/AndreRH/hangover/releases/latest

### PR
https://github.com/Botspot/pi-apps/pull/2670

### About the app

Hangover runs simple Win64 and Win32 applications on arm64 Linux. Only the application is emulated instead of a whole Wine installation.

### Confirmations

- [X] I have confirmed that this app is legal and not piracy.
- [X] I have read the app eligibilty rubric https://github.com/Botspot/pi-apps/issues/185 and confirmed this app is suitable for addition in Pi-Apps.
- [X] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi-Apps app list.
- [X] I have confirmed that this app can run on all [Pi-Apps supported systems](https://pi-apps.io/install/).

Botspot:
Is there a tutorial for installing and using this?
```
$ sudo apt install /home/pi/Downloads/hangover-wine_9.20~bookworm_arm64.deb /home/pi/Downloads/hangover-libqemu_9.20~bookworm_arm64.deb
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'hangover-wine' instead of '/home/pi/Downloads/hangover-wine_9.20~bookworm_arm64.deb'
Note, selecting 'hangover-libqemu' instead of '/home/pi/Downloads/hangover-libqemu_9.20~bookworm_arm64.deb'
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 hangover-wine : Depends: hangover-libarm64ecfex but it is not installable
                 Recommends: libodbc1 but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```
Also, rather than explaining how to install it, consider making a pi-apps install script yourself with [these easy instructions.](https://pi-apps.io/wiki/development/Creating-an-app/)
It's really easy. If you have Pi-Apps installed, just go to the settings and click New App, which will walk you through every step.

AndreRH:
As it says hangover-wine depends on
hangover-libarm64ecfex

[...]

PR #2670, opened 2024-10-27
add Hangover

AndreRH:
As discussed here: https://github.com/Botspot/pi-apps/issues/2669


Botspot:
Testing installation on our supported systems here: https://github.com/Botspot/pi-apps/actions/runs/11543239179
@theofficialgman it looks like some of the runners need fixing - something about not knowing how to uncompress an xz image.

Botspot:
Alright, so far from the testing it looks like one issue is how the debs are compressed - they cannot use the zst compression algorithm.

theofficialgman:
Nothing wrong with the runners. Bullseye doesn't have deb tar.zst support

theofficialgman:
Regardless though this can't be used on the Switroot releases due to upstream FEX regressions (segfault on launch on versions newer than 2407)

Botspot:
> Nothing wrong with the runners (regarding tar.zst). Bullseye and older doesn't have deb tar.zst support (doesn't affect Ubuntu releases)

I should have clarified. This failure in noble-64bit appears to have happened before it reached the app-installation stage.
20241027_17h04m06s_grim

AndreRH:
> Regardless though this can't be used on the Switchroot Ubuntu Jammy/Noble releases due to upstream FEX regressions (segfault on launch on versions newer than 2407)
> 
> Maybe if Hangover uses Box64 by default it could be considered.

Hangover doesn't ship a Linux FEX build, but 2 DLLs based on FEX emulator core and I'd assume it would work on switchroot that way.
Box64 is also shipped as a DLL, but is only the default for 32-bit emulation.

Botspot:
Could you add a bit of usage information to the description?

theofficialgman:
> > Regardless though this can't be used on the Switchroot Ubuntu Jammy/Noble releases due to upstream FEX regressions (segfault on launch on versions newer than 2407)
> > Maybe if Hangover uses Box64 by default it could be considered.
> 
> Hangover doesn't ship a Linux FEX build, but 2 DLLs based on FEX emulator core and I'd assume it would work on switchroot that way. Box64 is also shipped as a DLL, but is only the default for 32-bit emulation.

As I said. Fex doesn't work, it segfaults on versions newer than 2407. The regression hasn't been determined yet.

theofficialgman:
> > Nothing wrong with the runners (regarding tar.zst). Bullseye and older doesn't have deb tar.zst support (doesn't affect Ubuntu releases)
> 
> I should have clarified. This failure in noble-64bit appears to have happened before it reached the app-installation stage. 20241027_17h04m06s_grim

Seems like a temporary server error or Ubuntu removed the URL. I'll check later.

AndreRH:
> > > Regardless though this can't be used on the Switchroot Ubuntu Jammy/Noble releases due to upstream FEX regressions (segfault on launch on versions newer than 2407)
> > > Maybe if Hangover uses Box64 by default it could be considered.
> > 
> > 
> > Hangover doesn't ship a Linux FEX build, but 2 DLLs based on FEX emulator core and I'd assume it would work on switchroot that way. Box64 is also shipped as a DLL, but is only the default for 32-bit emulation.
> 
> As I said. Fex doesn't work on the mentioned systems, it segfaults on versions newer than 2407. The regression hasn't been determined yet.

I'm pretty sure this is only true for the Linux FEX version, a Windows version used in Hangover is likely not affected at all. Is the Linux FEX issue for Switchroot tracked somewhere?

Botspot:
When can the compression on the debs be fixed so that it works somewhere other than Bookworm?

theofficialgman:
> I'm pretty sure this is only true for the Linux FEX version, a Windows version used in Hangover is likely not affected at all. Is the Linux FEX issue for Switchroot tracked somewhere?

@azkali is there an open issue for the problem I have described or is it just discord discussion? I have myself had any time to bisect what causes versions newer than 2407 to segfault on Ubuntu but you seem to have trouble on every fex version on fedora.

Azkali:
Only discussed on discord

AndreRH:
> When can the compression on the debs be fixed so that it works somewhere other than Bookworm?

The bullseye package is built on bullseye, and I haven't heard of issues other than on your ci. I can look into it for the next release in three weeks though.

AndreRH:
> > When can the compression on the debs be fixed so that it works somewhere other than Bookworm?
> 
> The bullseye package is built on bullseye, and I haven't heard of issues other than on your ci. I can look into it for the next release in three weeks though.

I found the issue, the FEX packages are built on Ubuntu 22.04, which defaults to zstd. I can fix that easily.

apps/Hangover/uninstall
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+pkill -9 wine
+command -v wineserver >/dev/null && wineserver -k
+
+status -n "Removing terminal commands... "
+sudo rm -rf /usr/local/bin/generate-wine-prefix
```
theofficialgman, line 7: wrong file removed

AndreRH:
> > > > When can the compression on the debs be fixed so that it works somewhere other than Bookworm?
> > > 
> > > 
> > > The bullseye package is built on bullseye, and I haven't heard of issues other than on your ci. I can look into it for the next release in three weeks though.
> > 
> > 
> > I found the issue, the FEX packages are built on Ubuntu 22.04, which defaults to zstd. I can fix that easily.
> 
> Building debs for lower versions of linux distros on a newer distro is almost always going to cause a myriad of issues. Most common being that GLIBC requirements cannot be satisfied.
> 
> Your control file is also manually created and static in the github repo. This is unacceptable and you need to generate the dependencies and their minimum versions at compile/packaging time in order to guarantee that the builds can actually run on the installed system. Use the cpack DEB generator for that (if using a fully defined cmake project) or the underlying program that cpack DEB generator relies on `dpkg-shlibdeps` for generating those dependencies for any existing binary.

I'll say it again, those FEX packages contain Windows DLLs. In case you don't know what that is, consider them binary blobs, like pictures, the don't care about GLIBC, distributions or dependencies.

apps/Hangover/uninstall
AndreRH: good catch, I pushed a fixed version

theofficialgman:
> > Building debs for lower versions of linux distros on a newer distro is almost always going to cause a myriad of issues. Most common being that GLIBC requirements cannot be satisfied.
> > Your control file is also manually created and static in the github repo. This is unacceptable and you need to generate the dependencies and their minimum versions at compile/packaging time in order to guarantee that the builds can actually run on the installed system. Use the cpack DEB generator for that (if using a fully defined cmake project) or the underlying program that cpack DEB generator relies on `dpkg-shlibdeps` for generating those dependencies for any existing binary.
> 
> I'll say it again, those FEX packages contain Windows DLLs. In case you don't know what that is, consider them binary blobs, like pictures, the don't care about GLIBC, distributions or dependencies.

I was already aware those are windows dynamic libraries, not linux dynamic libraries. The wine debs however do contain linux binaries and need to have their control files generated. I missed that this is already the case https://github.com/AndreRH/hangover/blob/20da7093a53e3ca842d93ba8dca37da6b7c42c4d/.packaging/debian11/wine/debian/control#L63-L65  https://github.com/AndreRH/hangover/blob/master/.packaging/debian11/wine/Dockerfile#L9 so you are good there.

apps/Hangover/install-64
```diff
@@ -0,0 +1,305 @@
+#!/bin/bash
+
+version=9.20
```
theofficialgman, line 3: unrelated

apps/Hangover/install-64
```diff
@@ -0,0 +1,305 @@
...
+echo '
+[HKEY_CURRENT_USER\Control Panel\Desktop]
+"FontSmoothing"="'\$MODE'"
+"FontSmoothingOrientation"=dword:0000000'\$ORIENTATION'
+"FontSmoothingType"=dword:0000000'\$TYPE'
+"FontSmoothingGamma"=dword:00000578' >> \$TMPFILE
+
+echo "  - Windows 10 appearance theme"
+mkdir -p "\$WINEPREFIX/drive_c/windows/Resources/Themes/Windows_10"
+cp -f /opt/wine-${version}/Windows_10.msstyles "\$WINEPREFIX/drive_c/windows/Resources/Themes/Windows_10"
```
theofficialgman, line 144: see above
Please try to understand out scripts before copy/pasting

apps/Hangover/install-64
```diff
@@ -0,0 +1,305 @@
...
+else
+  error "User error: You are not using a supported Pi-Apps distribution."
+fi
+
+cd /tmp || error "Could not move to /tmp folder"
+wget https://github.com/AndreRH/hangover/releases/download/hangover-${version}/hangover_${version}_${ho_distro}_${__os_codename}_arm64.tar || error "Failed to download Hangover!"
+tar -xf hangover_${version}_${ho_distro}_${__os_codename}_arm64.tar || error "Failed to extract Hangover!"
+rm -f hangover_${version}_${ho_distro}_${__os_codename}_arm64.tar
+install_packages /tmp/hangover-libarm64ecfex_${version}_arm64.deb /tmp/hangover-libqemu_${version}~${__os_codename}_arm64.deb /tmp/hangover-libwow64fex_${version}_arm64.deb /tmp/hangover-wine_${version}~${__os_codename}_arm64.deb || exit 1
+rm -f ./hangover-libarm64ecfex_${version}_arm64.deb ./hangover-libqemu_${version}~${__os_codename}_arm64.deb ./hangover-libwow64fex_${version}_arm64.deb ./hangover-wine_${version}~${__os_codename}_arm64.deb
```
theofficialgman, line 52: this will conflict with the existing `Wine (x86)` and `Wine (x64)`. At pi-apps we don't allow applications to break each other. If they must break then installing one must automatically uninstall another one (eg: https://github.com/Botspot/pi-apps/blob/1c2387ba5c528ae3c78e35d87aa93d5e68ed5f4a/apps/AndroidBuddy/uninstall#L4 )

Please consider using a different naming convention for `hangover` than `wine` to avoid the conflict.

apps/Hangover/install-64
AndreRH: solved by removing all GUI related registry changes

apps/Hangover/install-64
AndreRH: the version variable actually is used

apps/Hangover/install-64
AndreRH: I'd expect the uninstall to happen in the install of Hangover then, no?

apps/Hangover/install-64
theofficialgman: It is only used (incorrectly) in the copy/pasted script from `Wine (x64)` from pi-apps which. Hence why I wrote "unrelated"

apps/Hangover/install-64
theofficialgman: Yes, if currently installed it will uninstall during the Hangover installation.

AndreRH: pushed, though IMHO it can't conflict with `Wine (x86)`, cause it's 32-bit only and Hangover is 64-bit only

theofficialgman: The binary name for using `Wine (x86)` does conflict as both `Wine (x86)` and `Wine (x64)` use the binary name `wine` (as per upstream defaults)

theofficialgman:
> I'm pretty sure this is only true for the Linux FEX version, a Windows version used in Hangover is likely not affected at all. Is the Linux FEX issue for Switchroot tracked somewhere?

Now tracked here https://github.com/FEX-Emu/FEX/issues/4148
We (Switchroot) are still deciding on Discord how to proceed.

apps/Hangover/install-64
AndreRH: AFAICT `Wine (x86)` can't be installed on 64-bit systems, Hangover can't be installed on a 32-bit system, so both will never meet, no?

Botspot: We also have Wine (x64) which also has a wine command, which is installable on arm64 systems. It would seem to conflict.

theofficialgman: Now that statement is true. So its not needed for `Wine (x86)` because of that.

Botspot: But that's not a bad thing. It would likely be a bad idea for a user to have two wine versions installed and running at the same time anyway.

Botspot:
The zst compression error is still an issue on Bullseye.
https://github.com/Botspot/pi-apps/actions/runs/11637595657/job/32411051344#step:3:1672

AndreRH:
> The zst compression error is still an issue on Bullseye. https://github.com/Botspot/pi-apps/actions/runs/11637595657/job/32411051344#step:3:1672

I now made a point release which solves that, 9.20.1, and pushed the change here

apps/Hangover/description
```diff
@@ -0,0 +1,10 @@
+Hangover runs simple Win64 and Win32 applications on arm64 Linux. Only the application is emulated instead of a whole Wine installation.
+
+Examples on how to run:
+
+$ wine your_x86_64_application.exe # Uses FEX for 64-bit apps
+$ wine your_x86_application.exe # To use Box64 for 32-bit apps
+$ HODLL=libwow64fex.dll wine your_x86_application.exe # To use FEX for 32-bit apps
+$ HODLL=xtajit.dll   taskset -c 1 wine your_x86_application.exe # To use QEMU for 32-bit apps
```
theofficialgman, line 8: this seems to differ from your own description

https://github.com/AndreRH/hangover?tab=readme-ov-file#running

which is incorrect? because from your repo's description I would expect QEMU for 32bit apps to be selected with
```
HODLL=wowarmhw.dllwine your_x86_application.exe
```

Additionally, are there plans to make QEMU and Box64 options for the 64-bit emulator dll?

AndreRH: thanks for having a look, I added a note
64-bit support using Box64, yes, but I'm quite time limited, so it might take a while
QEMU just isn't worth the effort

theofficialgman:
Consider adding a pretty important wine patch that we have been carrying in our builds.
https://github.com/Pi-Apps-Coders/files/blob/main/CompileCommands/Apps/Wine%20(x64).md
It fixes (well really works around) broken wine `.desktop` files that can occur on some systems. More details and link to the bug report at that link.

AndreRH:
ping
Any other blockers?

Botspot:
Sorry for the wait. I just wanted to test this running something before I merged the PR.
Upon installing Hangover on my main Pi, I noticed a couple things for users to take note of. I've updated the description to include it. @AndreRH if these changes seem factual and fair to you, let me know and I will merge the PR.

It would be nice if Hangover and Wine played nice. But installing it on a system that has a wineprefix already present, it asked permission to remove and regenerate the prexfix due to the errors:
```
Checking Wine prefix at /home/pi/.wine...
To choose another prefix, set the WINEPREFIX variable.
Waiting 5 seconds... 
002c:err:environ:init_peb starting L"C:\windows\system32\wineboot.exe" in experimental ARM64EC mode
002c:err:module:load_arm64ec_module could not load L"C:\windows\system32\libarm64ecfex.dll", status c0000135
wine: could not load kernel32.dll, status c000007b
◢◣ WARNING: Your previously existing Wine prefix failed with an error (see above). You chose to remove it and so it will be re-generated.
```

theofficialgman:
@Botspot due to other obligations I can't add this currently myself but can you please add this (hangover as hidden) to both the Jetson overrides lists. Until there is a solution upstream or downstream the majority of these devices cannot run this due to their old kernels lacking currently required functionality. I don't want this to be a blocker for other devices.

Botspot:
@AndreRH was there a reason you removed the Winetricks components from the install script? Without it, it is harder to install DLLs needed for commonly used applications.

AndreRH:
Users have reported issues with winetricks and hangover. So until I found time to sort it out, I 'd rather not include it

AndreRH:
Winetricks currently has recipes which can't support arm64ec, so let's not include it

PR #2673, opened 2024-11-02
Fastfetch: new app (draft)

theofficialgman:
@Botspot any opinion on an icon to use for this? [fastfetch](https://github.com/fastfetch-cli/fastfetch) doesn't seem to have one (I just kept the neofetch one for now).

I'll add an auto-updater assuming you have no objections.

Test run https://github.com/Botspot/pi-apps/actions/runs/11645397719


Botspot:
I think keeping the same icon could inform users that it is to be considered similar to Neofetch, so that is not a bad thing.
Now if you do want a bit of icon distinction, you could tweak the colors a bit, or italicize the penguin to signify speed maybe. Those are just some initial thoughts of mine.

Merge when ready.

theofficialgman:
> I think keeping the same icon could inform users that it is to be considered similar to Neofetch, so that is not a bad thing. Now if you do want a bit of icon distinction, you could tweak the colors a bit, or italicize the penguin to signify speed maybe. Those are just some initial thoughts of mine.

Ok. For now I will just leave it. Maybe in the future they will have an icon for branding of their own and we can switch.

PR #2686, opened 2024-11-27
Fix deprecated winetricks verb name

lukaszsobala:
This is more of a cosmetic change but winetricks always complains about `vcrun2005sp1` when installing Wine through pi-apps.


Botspot:
Let's wait until we have some other change to make to Wine, so that it won't be needlessly reinstalled for everybody.

lukaszsobala:
I am not sure how this works but I think my wine was reinstalled anyway when Hangover was added (#2670) but this one still was not merged

PR #2652, opened 2024-10-03
Contributor update

theofficialgman:
@Botspot this is by no means final. I have created this PR so we can start to discuss the exact wording and the presentation of the information that we discussed in #dev-chat on the discord. The wording can (and likely should) fully change and was kept brief so as to not waste my time on something that we didn't agree on.

PR #2708, opened 2025-01-19
Add retry button, subterminal, summary window to manage daemon

Botspot:
This is part of a larger improvement to the install/uninstall/update experience that will be ushered in with the "contributor update" https://github.com/Botspot/pi-apps/pull/2652 PR. This PR should probably be merged first.

This adds a new Retry button as an alternative to "Send report". Hopefully users will see and use it in the event of internet issues, or when instructed to fix their system in some way, which should have these effects:

- Moderately reduce our volume of received error reports.
- Help users keep apps installed long-term, rather than abandoning the app in an uninstalled state if an update ever fails once.
- As a result, this should improve the UX quite a bit for those with unstable internet connections.

Asking for @theofficialgman to look over a couple specific changes which I think are fine, but could be problematic:

1. My change to `will_reinstall` which now allows returning code 0 even if the app is uninstalled. I made this change so that updated apps, successfully uninstalling but failing to install, would actually be installed rather than refreshed if the user clicks Retry.
I found one usage of will_reinstall which may have new bad behavior with this change, so I replaced it with directly checking an app's installation status. This fix itself may also cause further regressions. I'm not sure.
2. My choice to avoid running a manage subprocess to update each app. (See the "`#manage can handle this action, but avoid spawning subprocess`" comment)
I vaguely remember trying to implement this once before, and there was some reason why running the "update" action was safer through a manage subprocess, but I cannot find that now nor do I see how this could cause issues.


theofficialgman:
The overall idea of retrying an application seems ok. I'll review the contents of this probably next weekend.

Shellcheck is failing due to the automatic change of ubuntu-latest moving from 22.04 to 24.04. I have already reported QEMU errors in newer QEMU that is used in 24.04 a few months back with no fix. We need to downgrade that runner back to 22.04.

theofficialgman:
Oh I almost forgot to mention, github actions ARM64 Linux runners are now free for public repos as of earlier this week. I need to do some refactoring to move our CI actions over from the QEMU based runners to the native ARM64 runners.

> I have already reported QEMU errors in newer QEMU that is used in 24.04 a few months back with no fix.

Actually this part is not the cause of the shellcheck errors. Shellcheck is running natively... not sure why it is hanging. But the change from 22.04 to 24.04 is causing shellcheck to fail so I will revert that action to 22.04.

theofficialgman:
Can't reproduce the issue on local Ubuntu Noble 24.04 shellcheck. Seems to only occur on Github Actions Ubuntu Noble 24.04. I will have reported the issue here: https://github.com/actions/runner/issues/3667

theofficialgman:
Switching shellcheck from operating on all files at once to a single file at a time reduced github actions shellcheck runtime from ~3 minutes down to ~25 seconds while also eliminating the chance of hangs observed locally https://github.com/Botspot/pi-apps/commit/1fa28fe781404d24e0b224a86716c58950db116f

Botspot:
New commits change from where the terminal is launched. Now it launches from manage daemon instead of from terminal_manage_multi.
This paves the path for a good summary window implementation as seen in #2652, and also finally un-breaks the bookworm experience.

Old behavior:
20250119_22h01m09s_grim
New behavior:
20250119_22h01m08s_grim

Botspot:
New commits more or less render #2652 obsolete
In addition to the two earlier specifics I want reviewed, @theofficialgman I would like your feedback on how I have the donation messages set up. Do you want your username/pfp/caption included in the way I have added it? Would you prefer your name not be there at all?

Botspot:
20250122_21h54m30s_grim

Botspot:
Okay @theofficialgman some things have changed since my original comment. If you are only looking at this now, ignore previous comments. I only have these two things that I am requesting for you to look over:

1. My choice to avoid running a manage subprocess to update each app. (See the "#manage can handle this action, but avoid spawning subprocess" comment in the code)
I vaguely remember trying to implement this once before, and there was some reason why running the "update" action was safer through a manage subprocess, but I cannot find that now nor do I see how this could cause issues.
2. The way your donate message is presented. I need to know if you are fine with it, want to change it, or want it removed.

As you can see on https://github.com/sponsors/botspot, my current real-life situation is fairly precarious, so I would like for this PR to be merged soon.

theofficialgman:
> 2\. The way your donate message is presented. I need to know if you are fine with it, want to change it, or want it removed.

I see nothing wrong with the way it is presented. It might be worth adding an additional row linking to all our contributors https://github.com/Botspot/pi-apps/graphs/contributors to be fair. If they have github sponsorship enabled (and some of them do) you can over over their icon and see a "sponsor" button on that page.

> As you can see on https://github.com/sponsors/botspot, my current real-life situation is fairly precarious, so I would like for this PR to be merged soon.

Noted. I'll DM you shortly (since its too personal to discuss in public imho). I don't intend on blocking your donation changes but given the nature of the other changes wrapped up in this PR I really need to do a thorough read through and regression test on our supported systems before this can be merged and I can't do that till the weekend (Saturday).

updater
```diff
@@ -433,49 +444,45 @@ Refreshes the Pi-Apps details for apps that do not need to be reinstalled for th
...
-      
-      #fix edge case: if app is installed but uninstall script doesn't exist somehow, then pretend app was uninstalled so that the reinstall later will happen noninteractively
-      if [ "$(app_status "$app")" == installed ];then
-        echo 'uninstalled' > "${DIRECTORY}/data/status/${app}"
-      fi
     fi
-
+    
+    #fix edge case: force-pretend app was uninstalled so that the reinstall later will happen noninteractively
+    echo 'uninstalled' > "${DIRECTORY}/data/status/${app}"
```
theofficialgman, line 470: manage installs are allowed to occur as long as the app status is not "disabled". That is the only check we have implemented so setting the status should not matter at all. I think the previous "edge case" was impossible because of this so we should remove this entirely. I have checked this to be true, all status except "Disabled" are allowed to proceed when executing manage install.

updater
```diff
@@ -433,49 +444,45 @@ Refreshes the Pi-Apps details for apps that do not need to be reinstalled for th
   
 }
 
-update_app() { #first arg is app name
+update_app() { #given app name, uninstall it, update app folder, install it again
```
theofficialgman, line 447: Its worth clearly documenting that "update_app" is only used for "real" updates, not refreshes anymore.

api
```diff
@@ -1081,51 +1081,24 @@ terminal_manage() { # wrapper for the original terminal_manage function to termi
   terminal_manage_multi "$action $app"
 }
 
-terminal_manage_multi() { #function to install/uninstall/update/refresh multiple apps (or filelists) - uses a terminal and refreshes the app list
-  queue="$1" #one or multiple actions and app names (or filelists of the format 'filelist:path/to/file:path/to/another/file')
```
theofficialgman, line 1085: Actually this documentation was somewhat correct. In https://github.com/Botspot/pi-apps/pull/2340/commits/149996ce31fa44cf871e6156130f2287ad072549 (part of the "use manage daemon for GUI updates" PR) we changed from files in a "filelist" format to individual files with each their own "update-file" action. We still support that and that action is used to update the non-app script (eg: manage, api, gui, etc) so we should properly document it here now.

api
```diff
@@ -2919,16 +2895,19 @@ send_error_report() { #non-interactively send a Pi-Apps error log file to the Bo
...
+  local num_lines="$(grep . <<<"$failure_list" | wc -l)"
+  local i=1 #counter to track which line in the list is current
+  local line
   local button
   
-  for app in $failed_apps ;do
+  for line in $failure_list ;do
+    local action="$(echo "$line" | awk -F';' '{print $1}')"
+    local app="$(echo "$line" | awk -F';' '{print $2}')"
+
```
theofficialgman, line 2910: I think that `diagnose_apps` has had a bug since it was introduced. I went through the history tracing back where it was located (it used to be in manage) and all its iterations and found when it was added as part of https://github.com/Botspot/pi-apps/pull/1580 and it seems like "action" was never set on a per app basis. If it was set at all, it was always inherited from the last time "action" was set for every app in the for loop (so it was likely to be incorrect). So we were almost always misdiagnosing apps.

Botspot: You are correct. I think the bug was a result of this function originally being run once per app, and later it was changed to run for multiple failed apps. Whatever the cause, it is fixed now.

I think the reason why nobody reported this issue is because most people will update a group of apps, or install several. $action is likely to be the same for all apps at a time.

api
Botspot: Not sure if you are suggesting there is current filelist usage in the master branch. I could not find any.

updater
Botspot: You seem to be right. I think this was from the distant past where manage install ran everything through a gui validate apps function.

Botspot: Removed in https://github.com/Botspot/pi-apps/pull/2708/commits/e7ae18a5cb75c9fc522ed832d00dfe03ae11a49c

api
theofficialgman: there isn't any "filelist" usage, it only existed in the PR where "filelist" was added and then refactored into "update-file". The documentation here just wasn't corrected to remove "filelist" and add "update-file". That is what I am saying needs to be done, we need to document that "update-file" exists and is used.

updater
Botspot: Done in https://github.com/Botspot/pi-apps/pull/2708/commits/e7ae18a5cb75c9fc522ed832d00dfe03ae11a49c

theofficialgman:
Just an FYI, I am still working through the code in my head. Mainly making sure that everything works as expected during the transition from the existing form of the scripts to this new form. I haven't actually run/tested anything new.

api
Botspot: fixed in https://github.com/Botspot/pi-apps/pull/2708/commits/0e8a591732867bff9a775e474bc4eae9c0d94e32

theofficialgman:
>     1. My choice to avoid running a manage subprocess to update each app. (See the "#manage can handle this action, but avoid spawning subprocess" comment in the code)
>        I vaguely remember trying to implement this once before, and there was some reason why running the "update" action was safer through a manage subprocess, but I cannot find that now nor do I see how this could cause issues.

There is a reason and the visual change here is a symptom of the modifications affecting runtime speed https://github.com/Botspot/pi-apps/pull/2708#issuecomment-2601293397 . The previous way allowed terminal-run to start as soon as possible with minimal overhead. That is why originally the terminal popped up before the manage yad window (which is why the terminal is behind the yad window). The changes you made to move the creation of the terminal-run instance much later in the process delayed the creation of the window to the point that yad started first (it is still a race condition because the yad window is a background task but it has a better chance of starting first since it was executed slightly before the terminal-run now).

The problem with this is terminal-run is slow to start on some systems (because their chosen terminal takes a long time to open) and the daemon pid won't be established until the terminal-run instance has started.

What you have done is actually break the pid check system entirely. Previously, the PID of the first terminal-run instance which had the daemon running was written to data/manage-daemon/pid and it is what other manage daemon calls would check to decide if they should become the daemon or only send their commands to the existing daemon. With your changes, now the updater script or gui script PID (depending on what called terminal_manage_multi) gets written to data/manage-daemon/pid, not the terminal-run instance that is actually running the daemon.

theofficialgman:
The PID check system prevents users from launching multiple pi-apps instances and creating multiple manage daemons. There is always only one daemon. It also allows users to start updates from the onboot updates and then open pi-apps and continue to add/remove applications all using the same daemon. With the PID system broken, all that won't work anymore.

Botspot:
Previously the entirety of the manage daemon ran in the terminal. I now want manage daemon to run outside of a terminal and only complete actions in the terminal, hence why I call it a "subterminal". This allows for:

- better window ordering
- for the terminal to be dismissed during the summary phase
- no annoying 30-second countdown

I will investigate the PID issue, but the subterminal changes I want to stay.

Botspot:
@theofficialgman I am confused by your statements that the PID was broken in any way.
There is a terminal pid that is part of terminal-run to wait for the terminal to close, but that did not change the manage script's behavior here:
```
  #write my own PID to the pid file
  echo $$ > "${DIRECTORY}/data/manage-daemon/pid"
```
That is line 624

I just don't see how this code path was affected in the way you describe.

theofficialgman:
> @theofficialgman I am confused by your statements that the PID was broken in any way. There is a terminal pid that is part of terminal-run to wait for the terminal to close, but that did not change the manage script's behavior here:
> 
> ```
>   #write my own PID to the pid file
>   echo $$ > "${DIRECTORY}/data/manage-daemon/pid"
> ```
> 
> That is line 624
> 
> I just don't see how this code path was affected in the way you describe.

the PID is set before the manage daemon is even started now. Note that the file is written and 30 lines later the manage daemon is started. The PID that gets set as I said will end up being the PID of the updater or gui script.

Botspot:
20250125_11h29m05s_grim

Botspot:
How exactly could $$ be the PID of gui or updater script?

theofficialgman:
deamon pid isn't the daemon PID anymore, as I said you are running this script directly from the updater/gui, so it still has the updater/gui PID. the PID that gets written needs to be done from your new manage_daemon_terminal_code function from within the terminal-run PID, not the updater/gui script.

Botspot:
> deamon pid isn't the daemon PID anymore, as I said you are running this script directly from the updater/gui, so it still has the updater/gui PID.

That would be true if we are sourcing the manage script and running a function that contains $$. But that is not the case. `"${DIRECTORY}/data/manage-daemon/pid"` will contain the PID of the manage script, which in a few seconds will run the rest of the daemon code in the terminal.

So code runs like this:
```
gui
  > manage
  > sets daemon pid
  > starts terminal
    > manage daemon runs
    > when it finishes, it removes the pid file
  > now this outer manage PID can finish, because of how terminal-run waits for the terminal to close before it exits
```

> the PID that gets written needs to be done from your new manage_daemon_terminal_code function from within the terminal-run PID, not the updater/gui script.

Bad idea. I thought about doing that, but we want the PID to be written as early as possible to avoid race conditions or duplicate processes.

theofficialgman:
do you see the problem now?

image

theofficialgman:
as you can see in the above, the manage-daemon PID is not written to the PID file but instead the manage script PID run from the updater script. That is not what was done before and is wrong for the reasons stated here https://github.com/Botspot/pi-apps/pull/2708#issuecomment-2614035813

Botspot:
Let's hop on discord vc

manage
```diff
@@ -353,6 +358,226 @@ $app"
...
+    export gman_message='to theofficialgman (notable Pi-Apps contributor)'
+    summary_window_click_function() {
+      if [ "$5" == "$botspot_message" ];then
+        xdg-open 'https://github.com/sponsors/botspot' &
+        
+        #remember that donation link was clicked - opts out of any popups introduced in the future
+        #"$(echo 'botspot donation click' | sha1sum | awk '{print $1}')"
+        if ! grep -q "8a780b08ac20332b897ecb399b25f4622bb31341" "${DIRECTORY}/data/runonce_hashes" ;then
+          echo "8a780b08ac20332b897ecb399b25f4622bb31341" >> "${DIRECTORY}/data/runonce_hashes"
+        fi
```
theofficialgman, line 523: is this dead code?

Botspot: no, the commented code just documents what the otherwise random looking hash represents.

Botspot: That was actually one of my most recent changes. The idea is if in the future I add some kind of donation message elsewhere with a popup, it won't appear for everybody who already clicked donate in the past.

theofficialgman:
PXL_20250125_190451831
There is one problem I have encountered when attempting to install applications after the update has been installed but pi-apps hasn't been restarted. Looking into what code path is responsible now. Luckily it's just a visual issue because once the installs a done the 30 second timeout appears on the terminal in the back.

theofficialgman:
Right.... the gui calls `terminal_manage` (which wraps to `terminal_manage_multi`) which is still the old version, hence the old window still pops up. Nothing we can do about that with our architecture unless we move back to the old method.

theofficialgman:
@Botspot given the above, do you have any issues with merging? I am fine to merge as is. If we want to prevent scenarios like this from occurring in the future we need to re-source the api anytime it changes from when the gui was first loaded (aka: we would have to version it and check the version).

theofficialgman:
also when you merge, I suggest squashing it.

Issue #2726, opened 2025-03-02
Ruffle (compiles from source)


matu6968:
### What is the name of the app?

Ruffle

### Where is the app hosted?

This will compile from source at https://github.com/ruffle-rs/ruffle due to no prebuilt ARM binaries available for desktop and the compile time is short (around 16 minutes on overclocked Pi 5 8GB), though if there will be any GitHub actions to make the binary from source on a ARM64 runner instead of doing it on the host, then this would eliminate making a binary on the host.

### About the app

Ruffle is a open source Adobe Flash Player emulator aimed at having the best compatibility as the official Adobe Flash Player (which is proprietary, EOL and has many unpatched security vulnerabilities)

**Ruffle is a constant work in progress, so except daily nigthly updates and bugs/broken features until a stable release in the future.**

Because no prebuilt is available for ARM64 nor ARMhf, this will compile the latest branch from source which as of writing this issue is [nightly-2025-03-02](https://github.com/ruffle-rs/ruffle/releases).

**Additional note: On the Pi, the Vulkan backend (aka default on Ruffle) is unstable and tends to cause graphical glitches and crashes to certain .swf content, to fix it, swap to the OpenGL backend and restart the app.**

### Upload file or Add PR Link

[Ruffle.zip](https://github.com/user-attachments/files/19044182/Ruffle.zip)

Tested on Ubuntu and Pi OS (64 bit only, on Pi OS it would install rustc and cargo via rustup due to these packages on Debian Bookworm being too old)

### Confirmations

- [x] I have confirmed that this app is legal and not piracy.
- [x] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi-Apps app list.
- [x] I have read and followed the [creating an app for Pi-Apps documentation](https://pi-apps.io/wiki/development/Creating-an-app/).
- [x] I have confirmed that this app can run on all [Pi-Apps supported systems](https://pi-apps.io/install/).

Botspot:
Looks cool. At the moment I think this would make a fine addition to Pi-Apps, unless gman can find some fatal issue with it, which does happen on occasion. :)

This one seems pretty involved, and I see a number of script areas for improvement or simplification, and you seem to know  Github fairly well. Would you mind converting this to a Pull Request to allow edits from maintainers?

matu6968:
[...]

PR #2727, opened 2025-03-03
Convert Ruffle app submission to PR

matu6968:
Requested by Botspot via issue #2726


apps/Ruffle/install
```diff
@@ -0,0 +1,129 @@
+#!/bin/bash
+
+version=nightly-2023-03-02 # dummy variable to make Pi-Apps update the app every new version
```
Botspot, line 3: This is not automatic, it needs a github actions script made, such as the ones you can find in .github/workflows/updates folder.
Documentation here: https://pi-apps.io/wiki/development/DOCUMENTATION/#automatic-app-updaters

apps/Ruffle/install
```diff
@@ -0,0 +1,129 @@
...
+else
+# Compare versions
+if [ "$MAJOR" -gt "$REQUIRED_MAJOR" ] || { [ "$MAJOR" -eq "$REQUIRED_MAJOR" ] && [ "$MINOR" -ge "$REQUIRED_MINOR" ]; }; then
+    echo "Rust compiler is OK (version $RUSTC_VERSION), continuing"
+    rust_too_old=0
+else
+    echo "Rust compiler is too old (found version $RUSTC_VERSION, required 1.78.0 or later), installing latest version via rustup"
+    rust_too_old=1
+fi
+fi
```
Botspot, line 33: This whole section could be simplified by using `sort`
See `package_is_new_enough`
```bash
package_is_new_enough() { #check if the $1 package has an available version greater than or equal to $2
  local package="$1"
  [ -z "$package" ] && error "package_is_new_enough(): no package specified!"
  
  local compare_version="$2"
  [ -z "$package" ] && error "package_is_new_enough(): no comparison version number specified!"
  
  #determine the latest available version for the specified package
  local package_version="$(package_latest_version "$package")"
  
  #if version value not found, return 1 now
  if [ -z "$package_version" ];then
    return 1
  fi
  
  #given both the package_version and compare_version, see if the greater of the two is the available package's version
  if [ "$(echo "$package_version"$'\n'"$compare_version" | sort -rV | head -n1)" == "$package_version" ];then
    #if so, indicate success
    return 0
  else
    return 1
  fi
}
```

apps/Ruffle/install
```diff
@@ -0,0 +1,129 @@
...
+
+# Check if there is enough RAM available
+# Get free memory in MB
+free_mem=$(free -m | awk '/^Mem:/{print $7}')
+
+# Set the threshold (3GB = 3096MB)
+threshold=3096
+
+if (( free_mem < threshold )); then
+     warning "Warning: Free RAM is below 3GB ($free_mem MB). Compilation may fail unless swap/zram is available."
```
Botspot, line 57: Can compilation be made to run on only 1 core if the RAM is below the threshold? We do that with a lot of apps by wrapping the `nproc` command with a function.
```bash
nproc() { #Reduce the number of compile threads on low-RAM systems
  #The Pi02 and Pi3A+ have 4 cores but 500MB of RAM. This function reduces the number of threads to use if RAM is low.
  
  #Estimation of how much memory is available for starting new applications, without swapping.
  local available="$(free | grep 'Mem:' | awk '{print $7}')"
  
  if [ "$available" -gt $((2000*1024)) ] || [ "$GITHUB_ACTIONS" == "true" ];then
    #available memory > 2000MB, use normal number of threads
    command nproc
  elif [ "$available" -gt $((1500*1024)) ];then
    #1500MB < available memory <= 2000MB, use 3 threads
    echo 3
    warning "Your system has less than 2000MB of available RAM, so this will compile with only 3 threads."
  elif [ "$available" -gt $((1000*1024)) ];then
    #1000MB < available memory <= 1500MB, use 2 threads
    echo 2
    warning "Your system has less than 1500MB of available RAM, so this will compile with only 2 threads."
  else
    #available memory <= 1000MB, use 1 thread
    echo 1
    warning "Your system has less than 1000MB of available RAM, so this will compile with only 1 thread."
  fi
}
```

apps/Ruffle/install
```diff
@@ -0,0 +1,129 @@
...
+Type=Application
+Categories=Game;Utility;
+MimeType=application/x-shockwave-flash;" | sudo tee /usr/share/applications/ruffle.desktop >/dev/null
+
+# Make user associate .swf files to Ruffle
+description="Make MIME association for Ruffle?
+
+You will be able to launch .swf files without manually going into Ruffle and loading a Adobe Flash file within Ruffle's UI itself."
+userinput_func "$description" "Yes, associate Ruffle as the default application for .swf files." "No, don't associate Ruffle as the default application for .swf files."
+  if [ "$output" == "Yes, associate Ruffle as the default application for .swf files." ]; then
```
Botspot, line 108: This question will be asked during every update. For that reason, I try to pick sane defaults and I don't tend to ask about MIME attributions but just go ahead and do it. Why would a user not want this?

apps/Ruffle/install
```diff
@@ -0,0 +1,129 @@
...
+    case "$choice" in
+        [Yy]*) echo "Continuing...";;
+        [Nn]*) echo "Exiting."; exit 1;;
+        *) echo "Invalid input. Exiting."; exit 1;;
+    esac
+fi
+
+status "Compiling Ruffle, this will take a bit"
+cd ruffle
+if cargo build --release --package=ruffle_desktop; then status_green "Compile success"; else error "Failed to compile Ruffle!"; fi
```
Botspot, line 70: This can be rewritten as:
```bash
cargo build --release --package=ruffle_desktop || error "Failed to compile Ruffle!"
status_green "Compile success"
```

apps/Ruffle/install
Botspot: Nvm, I saw line 77 now.

apps/Ruffle/install
matu6968: These suggestions are now applied and should be up in a moment.

apps/Ruffle/install
theofficialgman: Let me update that documentation. While "webVer" is still supported, it isn't the suggested variable name anymore to make it easier on developers.

.github/workflows/updates/Ruffle.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+TAG=$(curl -s "https://api.github.com/repos/ruffle-rs/ruffle/releases/latest" | jq -r '.tag_name')
+webVer="nightly-$(date -d "$(echo $TAG | grep -oE '[0-9]{4}-[0-9]{2}-[0-9]{2}')" +'%Y-%m-%d')"
```
theofficialgman, line 4: I encourage you to read the documentation that @Botspot pointed out again. We have a builtin function for prereleases (`get_prerelease`) which would return (at this moment) `nightly-2025-03-06` for your repo. Additionally, you do not set the `all_url` necessary for your script.

To add to this, we generally discourage the use of non-stable releases and really discourage nightly releases as they generally provide trivial changes that users would likely find annoying rather than helpful. Are there plans to switch to a standard release cadence at some point in the future?

apps/Ruffle/install
```diff
@@ -0,0 +1,86 @@
...
+    . "$HOME/.cargo/env"
+    status "Installing dependencies without a Rust toolchain"
+    install_packages libasound2-dev libxcb-shape0-dev libxcb-xfixes0-dev libgtk-3-dev libudev-dev libxcb-xinput-dev libxcb-xkb-dev libxcb-cursor-dev default-jre-headless cmake g++ || exit 1
+else
+    echo "Rust too old variable is at 0, installing Rust toolchain and dependencies via distro's repositories"
+    install_packages cargo rustc libasound2-dev libxcb-shape0-dev libxcb-xfixes0-dev libgtk-3-dev libudev-dev libxcb-xinput-dev libxcb-xkb-dev libxcb-cursor-dev default-jre-headless cmake g++ || exit 1
+fi
+
+# Download Ruffle source code and compile it
+git_clone https://github.com/ruffle-rs/ruffle || error 'Failed to clone Ruffle repository!'
```
theofficialgman, line 28: Pi-Apps scripts should be written such that they are repeatable regardless of when they are run. As such, a stable tag (as defined in the `version` variable at the top of this script) should be used, rather than the latest commit from the repo. This allows for our auto-updater to do its job by only updating an application if it can successfully build on our supported targets. This prevents shipping code to users that will not compile while still allowing us to perform frequent updates without constant supervision on the maintainers part.

.github/workflows/updates/Ruffle.sh
matu6968: > I encourage you to read the documentation that @Botspot pointed out again. We have a builtin function for prereleases (`get_prerelease_raw`) which would return (at this moment) `nightly-2025-03-06` for your repo. Additionally, you do not set the `all_url` necessary for your script.
> 
I am going to look further how app updates are being perfomed on Pi-Apps and update it accordingly, i took the inspiration of the update script logic from [Flow's update script](https://github.com/botspot/pi-apps/blob/master/.github/workflows/updates/Flow.sh) where it still used the legacy `webVer` variable and not the newer `version` and `all_url` variables.

> To add to this, we generally discourage the use of non-stable releases and really discourage nightly releases as they generally provide trivial changes that users would likely find annoying rather than helpful. Are there plans to switch to a standard release cadence at some point in the future?

As mentioned in the app's description, Ruffle is a [constant work in progress](https://github.com/ruffle-rs/ruffle/wiki/Frequently-Asked-Questions-For-Users#when-will-the-game-i-want-to-play-work) with only currently nightly releases being pushed out and that should be held up to Ruffle's developers for a discussion about release candidate/stable variant estimates (which since the [ActionScript API implemenations are mostly complete](https://ruffle.rs/compatibility) with minor features being missing this could likely take but maybe earlier another few years to be polished enough for stable production use) 

It is better that new Ruffle nightly releases tend to be more stable and functional then old releases where core functionality was missing (Ruffle mentions it in their FAQ that you can play most games on the recent nightly releases without having major issues) and issues are being constantly reported about Flash content issues which get solved eventually in new releases

apps/Ruffle/install
matu6968: > Pi-Apps scripts should be written such that they are repeatable regardless of when they are run. As such, a stable tag (as defined in the `version` variable at the top of this script) should be used, rather than the latest commit from the repo. This allows for our auto-updater to do its job by only updating an application if it can successfully build on our supported targets. This prevents shipping code to users that will not compile while still allowing us to perform frequent updates without constant supervision on the maintainers part.

Ruffle makes nightly snapshots of their code, so this can be solved by downloading the latest snapshot of the nightly release from the release tab. I will start implementing the nightly release snapshot download logic.

apps/Ruffle/install
```diff
@@ -0,0 +1,110 @@
...
+    echo "Reproducible build check, no commits found for $YESTERDAY, falling back to last available commit."
+    LATEST_COMMIT=$(git rev-list -n 1 --before="$TODAY 00:00:00" "$BRANCH")
+fi
+
+echo "Reproducible build check, checking out commit: $LATEST_COMMIT from $YESTERDAY"
+git checkout "$LATEST_COMMIT" || error "Failed to checkout latest commit!"
+status "Compiling Ruffle, this will take a bit"
+cd ~/ruffle
+cargo build --release --package=ruffle_desktop -j$(nproc) || error "Failed to compile Ruffle!"
+status_green "Compile success"
```
theofficialgman, line 57: Over-complicated. Use the `https://github.com/ruffle-rs/ruffle/releases/download/${version}/ruffle-${version}-reproducible-source.zip` like you do in the updater. Download zip, extract zip, build (same as before), install (same as before), done.

matu6968: > Over-complicated. Use the `https://github.com/ruffle-rs/ruffle/releases/download/${version}/ruffle-${version}-reproducible-source.zip` like you do in the updater. Download zip, extract zip, build (same as before), install (same as before), done.

I picked that approach because it used a snapshoted version while keeping git info like commit hash, commit ID and commit date. If i use the reproducible source method then builds are for sure can be reproduced but at a cost of warnings during compiling: 
```
warning: ruffle_desktop@0.1.0: not within a suitable 'git' worktree!
warning: ruffle_desktop@0.1.0: VERGEN_GIT_COMMIT_DATE set to default
warning: ruffle_desktop@0.1.0: VERGEN_GIT_COMMIT_TIMESTAMP set to default
warning: ruffle_desktop@0.1.0: VERGEN_GIT_SHA set to default
```
For now sticking with downloading `https://github.com/ruffle-rs/ruffle/releases/download/${version}/ruffle-${version}-reproducible-source.zip` then extracting the archive and compiling it like suggested.

theofficialgman: > I picked that approach because it used a snapshoted version while keeping git info like commit hash, commit ID and commit date. If i use the reproducible source method then builds are for sure can be reproduced but at a cost of warnings during compiling:

Just checkout the tag during the git clone like other scripts
https://github.com/Botspot/pi-apps/blob/45444baaa64e76f64b8f95b7b843abf50c0e4a10/apps/Linux%20Wifi%20Hotspot/install#L10
and update the automatic updater to match (webVer and version are fine in the automatic updater, just make sure touse the same name) https://github.com/Botspot/pi-apps/blob/45444baaa64e76f64b8f95b7b843abf50c0e4a10/.github/workflows/updates/Linux%20Wifi%20Hotspot.sh#L4

matu6968: Another suggestion later, moved to git tags since i just discovered you could clone a git repo with a specific tag set on them, hopefully now this should be the last change I will have to make here for a app submission

theofficialgman: > Another suggestion later, moved to git tags since i just discovered you could clone a git repo with a specific tag set on them, hopefully now this should be the last change I will have to make here for a app submission

Yes that's what I just showed you above...

apps/Ruffle/install
```diff
@@ -0,0 +1,90 @@
...
+  exit 0
+fi
+sudo mkdir /opt/ruffle || error "Failed to make Ruffle folder in /opt!" 
+sudo cp ruffle /opt/ruffle || error "Failed to copy Ruffle binary to /opt/ruffle!" 
+sudo cp "$(dirname "$0")/icon-64.png" /opt/ruffle || error "Failed to copy Ruffle icon to /opt/ruffle!" 
+
+# Make command asociations
+status "Making terminal command..."
+echo '#!/bin/bash
+/opt/ruffle/ruffle "$@"' | sudo tee /usr/local/bin/ruffle >/dev/null
```
theofficialgman, line 55: Be consistent with global installation. Don't mix and match /usr/local and /usr (see next comment)

apps/Ruffle/install
```diff
@@ -0,0 +1,90 @@
...
+status "Making menu launcher..."
+echo "[Desktop Entry]
+Name=Ruffle
+Comment=Open source Flash Player emulator written in Rust
+Exec=ruffle %f
+Icon=/opt/ruffle/icon-64.png
+Terminal=false
+Type=Application
+Categories=Game;Utility;
+MimeType=application/x-shockwave-flash;" | sudo tee /usr/share/applications/ruffle.desktop >/dev/null
```
theofficialgman, line 68: like done here ^

theofficialgman: installation shouldn't be done to `/usr/share` except for apt managed files. So this (and all other instances that use `/usr`) need to use `/usr/local/`

Botspot:
Ready for merge?

matu6968:
> Ready for merge?

It seems to be polished enough for theofficialgman to be happy with it (other then the version being a bit out of date which will be automatically updated by the GitHub action) so it means ready to merge

PR #2758, opened 2025-04-08
Add proper security policy and code of conduct to pi-apps

matu6968:
This a night and day contrast to #2545 and #2429 where the contributor didn't dare to follow the Pi-Apps structure, unlike me where i analysed how the structure works and what related spam issues/PR's to reference it from.


theofficialgman:
@matu6968 I cannot speak for @Botspot but I will say I really appreciate the thought and detail you put into these policies. I have read them and do think they accurately describe what I believe to be our standards. My only nitpick would be with the description I will highlight below (and similar clauses sprinkled throughout)

> regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation

To be clear, I agree with these clauses but I don't want to explicitly highlight any particular "category" in the first sentence of the code of conduct. I would prefer a more flexible (and realistic because lets be honest everyone makes mistakes and the mistakes aren't always because they have bad intentions) code of conduct like used for the dolphin emulator https://github.com/dolphin-emu/dolphin/blob/master/CODE_OF_CONDUCT.md

Don't feel obligated to make any changes right now. Lets wait to hear @Botspot's opinion. I am not sure if @Botspot even wants to entertain adding security and code of conduct policies (even though I think the contents of what you have written are probably majority agreed upon by all parties here).

Botspot:
I like this. A year ago I likely would not have seen much value in it, but if nothing else this adds a bit of professionalism to the project, and also gives us a written list of expectations to refer to, if we encounter any more problematic users. I think at this project size, it now makes sense to add a high quality code of conduct and security policy, which these definitely are. Good job.

@matu6968 I think @theofficialgman has good points about reframing the beginning part. I agree that the "here's the characteristics we pledge to not discriminate on" as a first sentence seems a bit out of place. The Dolphin emulator puts that list in one of the rules, which seems more fitting.

matu6968:
> @matu6968 I think @theofficialgman has good points about reframing the beginning part. I agree that the "here's the characteristics we pledge to not discriminate on" as a first sentence seems a bit out of place. The Dolphin emulator puts that list in one of the rules, which seems more fitting.

Should be now fixed at this point.

matu6968:
Currently waiting for merge...

Issue #2787, opened 2025-05-13
Prusa slicer not working


iugin2:
### Confirmations

- [x] I searched for similar bug reports in https://github.com/Botspot/pi-apps/issues and found none were relevant.
- [x] I tried updating Pi-Apps and the issue is still not fixed. [How to update Pi-Apps?](https://pi-apps.io/wiki/getting-started/updating-pi-apps/)
- [x] IMPORTANT: I promise to respond to comments on this issue and write a good bug report https://www.softwaretestinghelp.com/how-to-write-good-bug-report/

### What happened?

Prusa Slicer doesn't launch

### Description

I installed Prusa Slicer 2.9.1 from Pi app (via install button) and doesn't work.
I tried the terminal comand you indicate but this error appears,

### What are your system specs (run the following command in your terminal)?

```bash
OS: Debian GNU/Linux 12 (bookworm)
OS architecture: 64-bit
Last updated Pi-Apps on: 05/13/2025
Latest Pi-Apps version: 05/13/2025
Kernel: aarch64 6.12.25+rpt-rpi-v8
Device model: Raspberry Pi 400 Rev 1.0
SOC identifier: bcm2711
Machine-id (hashed): 3e9f12ae418a29baace58b94c77a6839a269087a
Serial-number (hashed): e7fbc4792e6a7fd91d2b361f720e47f8ae978bb8
Cpu name: Cortex-A72
Ram size: 3.79 GB
Raspberry Pi OS image version: 2025-05-13
Language: en_US.UTF-8
```

### (Recommended) Error log? Terminal output? Debug messages?

```bash
MESA_GL_VERSION_OVERRIDE=3.3 /opt/PrusaSlicer.AppImage
/tmp/.mount_PrusaSIeq6Nn/usr/bin/prusa-slicer: error while loading shared libraries: libwebkit2gtk-4.0.so.37: cannot open shared object file: No such file or directory
```

theofficialgman:
Its a pretty odd requirement for an appimage to require a specifically versioned system library like this. The whole point of appimages is you include enough libraries in them to NOT be so dependent in system libraries that change between OS releases.

[...]

PR #2795, opened 2025-05-29
check if adding prusaslicer lib breaks any supported systems

Botspot:
Fixes https://github.com/Botspot/pi-apps/issues/2787


theofficialgman:
see https://github.com/Botspot/pi-apps/issues/2787#issuecomment-2926193150

Issue #2807, opened 2025-07-10
App Submission: yaScrcpy GUI (Install Script)


gabreek:
### What is the name of the app?

yaScrcpy GUI

### Where is the app hosted?

https://github.com/gabreek/qt-yascrcpygui

### About the app

Main Purpose
  yaScrcpy GUI is a user-friendly graphical interface for the powerful Android screen
  mirroring tool, scrcpy. It allows users to manage and launch scrcpy sessions without
  needing to use the command line, making it more accessible for all users, especially those
  who prefer a visual workflow.

  How it Works
  The application is built with Python and the PySide6 (Qt) framework. The provided
  installation script handles the entire setup process:
   1. It first checks if the scrcpy dependency is already installed on the system.
   2. It then clones the source code from GitHub.
   3. It builds a standalone executable using PyInstaller within an isolated Python virtual
      environment to avoid conflicts with system packages.
   4. Finally, it installs the binary, an application icon, and a .desktop shortcut for the
      system's application menu.

  Pi-Apps Eligibility
  This app fits the Pi-Apps rubric because it is open-source, provides a useful function for
  Raspberry Pi users (interacting with Android devices), and enhances an existing, popular
  command-line utility (scrcpy) with an accessible GUI. It has no special kernel or hardware
  requirements.

### Upload file or Add PR Link

The installation files are structured and available in my project's repository at the 
  following link:

https://github.com/gabreek/qt-yascrcpygui/tree/master/pi-apps/yaScrcpy%20GUI

i also uploaded zipped apps install folder
[yaScrcpy GUI.zip](https://github.com/user-attachments/files/21171692/yaScrcpy.GUI.zip)

### Confirmations

- [x] I have confirmed that this app is legal and not piracy.
- [x] I have confirmed that this app has never been discussed in https://github.com/Botspot/pi-apps/issues and https://github.com/Botspot/pi-apps/pulls, and it is not in the Pi-Apps app list.
- [x] I have read and followed the [creating an app for Pi-Apps documentation](https://pi-apps.io/wiki/development/Creating-an-app/).
- [x] I have confirmed that this app can run on all [Pi-Apps supported systems](https://pi-apps.io/install/).

Botspot:
This seems like a good fit for Pi-Apps.

I suggest:
[...]

PR #2810, opened 2025-07-17
Add yaScrcpy GUI

gabreek:
What is the name of the app?
yaScrcpy GUI

Where is the app hosted?
https://github.com/gabreek/qt-yascrcpygui

About the app
Main Purpose
yaScrcpy GUI is a user-friendly graphical interface for the powerful Android screen
mirroring tool, scrcpy. It allows users to manage and launch scrcpy sessions without
needing to use the command line, making it more accessible for all users, especially those
who prefer a visual workflow.

How it Works
The application is built with Python and the PySide6 (Qt) framework. The provided
installation script handles the entire setup process:

It first checks if the scrcpy dependency is already installed on the system.
It then clones the source code from GitHub.
It builds a standalone executable using PyInstaller within an isolated Python virtual
environment to avoid conflicts with system packages.
Finally, it installs the binary, an application icon, and a .desktop shortcut for the
system's application menu.
Pi-Apps Eligibility
This app fits the Pi-Apps rubric because it is open-source, provides a useful function for
Raspberry Pi users (interacting with Android devices), and enhances an existing, popular
command-line utility (scrcpy) with an accessible GUI. It has no special kernel or hardware
requirements.

https://github.com/Botspot/pi-apps/issues/2807


theofficialgman:
Pretty sure this won't work on arm32 due to no whl for pyside6 on it. https://pypi.org/project/PySide6/

gabreek:
Okay, I hadn't checked that. I've renamed the install script.

apps/yaScrcpy GUI/uninstall
```diff
@@ -0,0 +1,40 @@
...
+    sudo xdg-icon-resource forceupdate --mode system || true
+
+    # The dependencies (git, python3-pip, python3-venv) are common system packages
+    # and will not be removed to avoid breaking other applications.
+
+    status "Uninstallation of ${APP_NAME} completed successfully!"
+}
+
+# Run the main function
+main "$@"
```
Botspot, line 40: What's the purpose of putting this into a main() function here??

apps/yaScrcpy GUI/uninstall
```diff
@@ -0,0 +1,40 @@
...
+    sudo rm -f "${BIN_PATH}"
+
+    status "Removing application icon..."
+    sudo rm -f "${ICON_PATH}"
+
+    status "Removing application shortcut..."
+    sudo rm -f "${DESKTOP_FILE}"
+
+    status "Updating icon cache..."
+    # Use || true to prevent the script from failing on non-critical cache update errors
```
Botspot, line 29: `|| true` is only necessary if the script has `set -e` enabled, causing the script to exit if any comnmands exit with a nonzero exit code. Pi-Apps does not use `set -e`, meaning it does not care about nonzero exit codes, making `|| true` unnecessary where failing is OK, but `|| error` very important where failing is not OK.

PR #2834, opened 2025-09-17
Add FreeCAD

ooovlad:
Add **FreeCAD** to pi-apps.
Available only on 64-bit ARM systems.

- Installation script uses `https://github.com/FreeCAD/FreeCAD/releases/latest` to download latest AppImage (falls back to version 1.0.2 if it fails to detect AppImage for aarch64 Linux in `latest`).
- Then it downloads `.desktop` and mimetype `.xml` files from that repository to put to corresponding directories.
- Finally, it gets `.svg` icon from Wikimedia or falls back to `icon-64.png` if unable to download.

### DIFF
--- a/apps/FreeCAD/install-64
+++ b/apps/FreeCAD/install-64
@@ -4,7 +4,7 @@ version=1.0.2
 
 wget -O /tmp/FreeCAD.AppImage "https://github.com/FreeCAD/FreeCAD/releases/download/${version}/FreeCAD_${version}-conda-Linux-aarch64-py311.AppImage" || exit 1
 
-sudo mv /tmp/FreeCAD.AppImage /opt/FreeCAD.AppImage || error "failed to move AppImage to /opt"
+sudo mv /tmp/FreeCAD.AppImage /opt/FreeCAD.AppImage || error "Failed to move AppImage to /opt"
 
 status -n "Installing AppImage to /opt, making it executable, adding symlink... "
 sudo chmod +x /opt/FreeCAD.AppImage || error "Failed to mark as executable!"
@@ -14,23 +14,21 @@ sudo ln -s /opt/FreeCAD.AppImage /usr/local/bin/FreeCAD || error "Failed to make
 status_green Done
 
 
-### DOWNLOAD OTHER ASSETS
+### DOWNLOAD AND INSTALL OTHER ASSETS
 
 # Download .desktop file.
 status "Downloading .desktop file..."
-wget -O /tmp/org.freecad.FreeCAD.desktop "https://github.com/FreeCAD/FreeCAD/raw/refs/heads/main/src/XDGData/org.freecad.FreeCAD.desktop" || exit 1
+wget -O /tmp/org.freecad.FreeCAD.desktop "https://github.com/FreeCAD/FreeCAD/raw/refs/tags/$version/src/XDGData/org.freecad.FreeCAD.desktop" || exit 1
 sudo mv /tmp/org.freecad.FreeCAD.desktop /usr/share/applications/org.freecad.FreeCAD.desktop || error 'Unable to move .desktop file from tmp'
 
 # Download and install icon
 status "Downloading icon..."
-wget -O /tmp/org.freecad.FreeCAD.svg "https://github.com/FreeCAD/FreeCAD/raw/refs/heads/main/src/Gui/Icons/freecad.svg" || exit 1
+wget -O /tmp/org.freecad.FreeCAD.svg "https://github.com/FreeCAD/FreeCAD/raw/refs/tags/$version/src/Gui/Icons/freecad.svg" || exit 1
 sudo mv /tmp/org.freecad.FreeCAD.svg /usr/share/pixmaps/org.freecad.FreeCAD.svg || error 'Failed to install the icon.'
 
 # Download mimetype xml file
 status "Downloading mimetype file..."
-wget -O /tmp/freecad.xml "https://github.com/FreeCAD/FreeCAD/raw/refs/heads/main/src/XDGData/org.freecad.FreeCAD.xml" || exit 1
-
-### PROCESS ASSETS
+wget -O /tmp/freecad.xml "https://github.com/FreeCAD/FreeCAD/raw/refs/tags/$version/src/XDGData/org.freecad.FreeCAD.xml" || exit 1
 
 # Make .FCSTd mimetype
 status -n "Creating .fcstd mimetype... "
@@ -41,6 +39,8 @@ if ! sudo update-mime-database /usr/share/mime; then
   exit 1
 fi
 status_green Done
+
+# Write mimetype association to ~/.config/mimeapps.list
 if ! grep -qF 'org.freecad.FreeCAD.desktop' ~/.config/mimeapps.list; then
   status -n "Associating the .fcstd mimetype with FreeCAD... "
   if ! grep -qF '[Added Associations]' ~/.config/mimeapps.list; then
--- a/apps/FreeCAD/uninstall
+++ b/apps/FreeCAD/uninstall
@@ -5,10 +5,13 @@ purge_packages || exit 1
 
 # Remove AppImage itself and all additions.
 status -n 'Removing AppImage and links... '
-sudo rm -f /opt/FreeCAD.AppImage /usr/local/bin/FreeCAD /usr/share/applications/org.freecad.FreeCAD.desktop /usr/share/pixmaps/org.freecad.FreeCAD.svg /usr/share/pixmaps/org.freecad.FreeCAD.png /usr/share/mime/packages/freecad.xml || error 'Unable to execute sudo rm -f <files>'
+sudo rm -f /opt/FreeCAD.AppImage /usr/local/bin/FreeCAD /usr/share/applications/org.freecad.FreeCAD.desktop /usr/share/pixmaps/org.freecad.FreeCAD.svg /usr/share/mime/packages/freecad.xml || error 'Unable to execute sudo rm -f <files>'
 status_green Executed
 
-# Clear mimetype association.
+# Clear mimetype association, then remove mimeapps.list if it's empty or only contains the header.
 if [ -f ~/.config/mimeapps.list ]; then
   sed -i '\#application/x-extension-fcstd=org\.freecad\.FreeCAD\.desktop;#d' ~/.config/mimeapps.list || error 'Unable to remove mimetype association!'
+  if awk '!/^$/ && $0 != "[Added Associations]" {exit 1}' ~/.config/mimeapps.list; then
+    rm -f ~/.config/mimeapps.list
+  fi
 fi


Botspot:
Let's change the script to hardcode a `version=` line, rather than grabbing the latest version. Then a github auto-updater script can be added to this repo that will check for new versions, edit the `version=` line, causing all pi-apps clients to show this app as updatable.

Botspot:
In the above commits, I did the following:
- Simplified the scripts
  - Downloaded the SVG from github rather than wikipedia
  - Removed the big mimetypes logic blob in `uninstall`. TBH, I did not take the time to try to understand it.
  - Many minor code format changes, to follow past convention and the KISS principle. For example I see little need for a tmpdir - just dropping the files into /tmp is fine.
- Downloaded the freecad icon from the github rather than wikipedia.
- Added the github actions updater
- Fixed the app icons to use transparency, and fixed pi-apps to generate icons with transparency from now on

@ooovlad, you clearly know what you are doing when it comes to scripting. Please review the changes and check if my commits could cause any regressions or edge cases.

ooovlad:
@Botspot
>   * Removed the big mimetypes logic blob in `uninstall`. TBH, I did not take the time to try to understand it.

##### This mentioned ["logic blob"](https://github.com/Botspot/pi-apps/pull/2834/commits/ddf99b32d2bc5e09725d3beb55c1c463b8eff71f#diff-c3eafbdf8427d2ccb07ea9cd0280270421d2d6dad56290e20bb803d32ca1ed2e) was doing the following:

1. If file `~/.config/mimeapps.list` exists do:
2. If this file has `org.freecad.FreeCAD.desktop` entry in it, remove mimetype association from file (backup `mimeapps.list` to `/tmp` before making changes, remove backup after the changes are made, or restore on error);
3. If this file doesn't have have anything other than blank lines and `[Added Associations]` line, remove this file.

##### Why this was made?

`~/.config/mimeapps.list` doesn't exist on Raspberry Pi OS by default, so my intention was not to leave in system empty file that does nothing. I also noticed that other installations like `OpenSCAD` use this file the same way, this is why file content checks are made. (BTW, OpenSCAD doesn't check this file on uninstall. It leaves the file in system.)

**I propose to return this code back to uninstall script.**

---

_Reply on other changes is in progress..._

ooovlad:
I see your change to code that downloads AppImage. I understand your intention to use automatic updates script and I like it. However, lines of code for finding download URL should not be removed, because `$version` is not the only thing that may change in future. As you can see, your link `https://github.com/FreeCAD/FreeCAD/releases/download/${version}/FreeCAD_${version}-conda-Linux-aarch64-py311.AppImage` contains `py311` and `conda`, which may be changed or removed in future, hence making install64 script to fail.

Now I'm going to change `install64` slightly, to make it back more future-proof.

ooovlad:
All changes I wanted are made. Waiting for your response on [this](https://github.com/Botspot/pi-apps/pull/2834#issuecomment-3304098628)...

And one more thing.
I used temp folder because I trapped it with `trap 'rm -rf "$tmp"' EXIT` so it automatically deletes itself if you press `Ctrl + C` during installation process. Without `trap` partially downloaded file is left in `/tmp` directory. Using temp folder allows me to use a single trap on $tmp folder rather than trapping every single file.

Botspot:
> All changes I wanted are made. Waiting for your response on [this](https://github.com/Botspot/pi-apps/pull/2834#issuecomment-3304098628)...
> 
> And one more thing. I used temp folder because I trapped it with `trap 'rm -rf "$tmp"' EXIT` so it automatically deletes itself if you press `Ctrl + C` during installation process. Without `trap` partially downloaded file is left in `/tmp` directory. Using temp folder allows me to use a single trap on $tmp folder rather than trapping every single file.

Yes I understand that, but none of the other pi-apps app scripts bother to do this. tmp is cleared out on a reboot, and it's not stored in RAM (on Pi OS at least), so orphan files left behind are not a problem. The only practical risk is lingering files causing failures on a second run. The install script should be verified to work if it runs multiple times in a row.

Botspot:
> @Botspot
> 
> > * Removed the big mimetypes logic blob in `uninstall`. TBH, I did not take the time to try to understand it.
> 
> ##### This mentioned ["logic blob"](https://github.com/Botspot/pi-apps/pull/2834/commits/ddf99b32d2bc5e09725d3beb55c1c463b8eff71f#diff-c3eafbdf8427d2ccb07ea9cd0280270421d2d6dad56290e20bb803d32ca1ed2e) was doing the following:
> 1. If file `~/.config/mimeapps.list` exists do:
> 2. If this file has `org.freecad.FreeCAD.desktop` entry in it, remove mimetype association from file (backup `mimeapps.list` to `/tmp` before making changes, remove backup after the changes are made, or restore on error);

Is there a reason to not use `sed -i` for this? `sed -i` makes a temporary file to edit, then only if successful does it overwrite the original file with the newly edited temporary file.

> 3. If this file doesn't have have anything other than blank lines and `[Added Associations]` line, remove this file.
> 
> ##### Why this was made?
> `~/.config/mimeapps.list` doesn't exist on Raspberry Pi OS by default, so my intention was not to leave in system empty file that does nothing. I also noticed that other installations like `OpenSCAD` use this file the same way, this is why file content checks are made. (BTW, OpenSCAD doesn't check this file on uninstall. It leaves the file in system.)

OpenSCAD's only mime association was for its own dedicated filetype. No other software uses that filetype, so I chose to not touch the mimeapps.list on uninstall. I will readily admit that is a bad practice.

> **I propose to return this code back to uninstall script.**

If you can make the code readable, it's fine to add it back. Just be aware none of the other apps go to such lengths to handle situations such as the file being empty. My philosophy is if empty file=no problem and more complication=more risk, then don't bother removing it. I will not force that philosophy on you, but the code does need to be readable and safe in all circumstances.

Botspot:
> I see your change to code that downloads AppImage. I understand your intention to use automatic updates script and I like it. However, lines of code for finding download URL should not be removed, because `$version` is not the only thing that may change in future. As you can see, your link `https://github.com/FreeCAD/FreeCAD/releases/download/${version}/FreeCAD_${version}-conda-Linux-aarch64-py311.AppImage` contains `py311` and `conda`, which may be changed or removed in future, hence making install64 script to fail.
> 
> Now I'm going to change `install64` slightly, to make it back more future-proof.

Absolutely, the link format probably will change, resulting in later download URLs being invalid. That is fine. It happens here on a regular basis, but it is not problematic in the way you might think. Our Github actions script detects it, [warns us](https://github.com/Botspot/pi-apps/issues/2539), and refuses to update the script in such cases. We receive the warning, correct the script and re-test everything, and then push the changes. Everyone stays on the last working version until we get around to pushing a manual update. It's safer this way.

For example, a URL change could signify that other changes may be necessary. I remember one time when the maintainers changed the archive format from tar.bz2 to tar.xz, so `tar` needed to have its flags changed. We would rather keep everyone on a known working version, rather than have automation robust enough that all users are upgraded to a newer version that needs changes to the implementation to function.

As this PR stands, install-64 is no more future-proof than my version I pushed. Even with your changes, the github actions bot will keep this on the last version that follows this URL naming scheme. That is because you only put your future-proofing code into the install-64 script, not also in the github actions script. As a result, the runner will try the new version, test the new version in the old URL naming scheme, find it to be invalid, and move on to the next app.
I would encourage you to read up on how this system works. https://pi-apps.io/wiki/development/DOCUMENTATION/#automatic-app-updaters

ooovlad:
@Botspot

> If you can make the code readable, it's fine to add it back.

Done. Simplified it to 3 lines. I genuinly agree with you about previous clunkiness of that part.

Regarding updates, I also agree that this script should retain the same logic as other scripts have. Your point is now clear for me.

Botspot:
> Done. Simplified it to 3 lines. I genuinly agree with you about previous clunkiness of that part.
> 
> Regarding updates, I also agree that this script should retain the same logic as other scripts have. Your point is now clear for me.

Thank you as well. Overall your PR was of unusually high quality. I hope we can collaborate on more app additions in the future.

apps/FreeCAD/install-64
```diff
@@ -0,0 +1,53 @@
...
+sudo mv /tmp/org.freecad.FreeCAD.desktop /usr/share/applications/org.freecad.FreeCAD.desktop || error 'Unable to move .desktop file from tmp'
+
+# Download and install icon
+status "Downloading icon..."
+wget -O /tmp/org.freecad.FreeCAD.svg "https://github.com/FreeCAD/FreeCAD/raw/refs/tags/$version/src/Gui/Icons/freecad.svg" || exit 1
+sudo mv /tmp/org.freecad.FreeCAD.svg /usr/share/pixmaps/org.freecad.FreeCAD.svg || error 'Failed to install the icon.'
+
+# Download mimetype xml file
+status "Downloading mimetype file..."
+wget -O /tmp/freecad.xml "https://github.com/FreeCAD/FreeCAD/raw/refs/tags/$version/src/XDGData/org.freecad.FreeCAD.xml" || exit 1
```
theofficialgman, line 31: @Botspot @ooovlad no need to download any of these, they are all available in the appimage itself. if you want to install them to the system, you can simply extract and move them to the correct place

```
./FreeCAD_weekly-2025.09.17-Linux-x86_64-py311.AppImage --appimage-extract usr/share/mime/packages/org.freecad.FreeCAD.xml
./FreeCAD_weekly-2025.09.17-Linux-x86_64-py311.AppImage --appimage-extract org.freecad.FreeCAD.svg
./FreeCAD_weekly-2025.09.17-Linux-x86_64-py311.AppImage --appimage-extract org.freecad.FreeCAD.desktop
```


```bash
$ tree squashfs-root
squashfs-root
├── org.freecad.FreeCAD.desktop
├── org.freecad.FreeCAD.svg
└── usr
    └── share
        └── mime
            └── packages
                └── org.freecad.FreeCAD.xml

5 directories, 3 files
```

theofficialgman:
> Thank you as well. Overall your PR was of unusually high quality. I hope we can collaborate on more app additions in the future.

I'd like to second this. Well done. We can cleanup my review notes in a future update.

Botspot:
Look what I found in `runonce_entries`...

```bash
# remove deprecated apps (only when running in a visible terminal)
...
  remove_deprecated_app "FreeCAD"
```
We used to have a FreeCAD app and then removed it. This code was still here removing the new app too on new pi-apps installs. 😆

Botspot:
> > Thank you as well. Overall your PR was of unusually high quality. I hope we can collaborate on more app additions in the future.
> 
> I'd like to second this. Well done. We can cleanup my review notes in a future update.

I went ahead and did it now, before a lot of users get the chance to install it and then receive an update.

ooovlad:
@Botspot Why this commit 134b823c74fed9f3506ca089a99795c4cff025d7 caused `pi-apps` on my Raspberry Pi to download FreeCAD once again and install it? Why were these changes considered as "update", even though `version` have not been changed?

Botspot:
> @Botspot Why this commit [134b823](https://github.com/Botspot/pi-apps/commit/134b823c74fed9f3506ca089a99795c4cff025d7) caused `pi-apps` on my Raspberry Pi to download FreeCAD once again and install it? Why were these changes considered as "update", even though `version` have not been changed?

See https://pi-apps.io/wiki/development/DOCUMENTATION/#the-updater-script

> This script handles Pi-Apps updates, both for apps and for files. Apps do not have version numbers, so updating involves comparing the installation-script of the existing app-folder with the installation-script of the new app-folder.

This is normal. Suppose we need to made some non-version-related bugfix to the script. All pi-apps clients should and will reinstall the app to get the latest changes. The entire app install script is compared to the new version. If there is any variation in the new version, the app will be reinstalled rather than merely refreshed.

PR #2836, opened 2025-09-19
simplify userinput_func

Botspot:
Much simpler implementation, takes up less width on the screen for 2 choices, UX is better for 3+ choices. (clicking anywhere on the choice works, rather than having to click exactly in the boundaries of a radio button)

I cannot foresee any regressions caused by this. But this is @theofficialgman's function, so requesting review just to be safe.

Before:
image

After:
image

### DIFF
--- a/api
+++ b/api
@@ -1744,8 +1744,8 @@ ${category}"
   fi
 }
 
-userinput_func() { # userinput function to display yad prompts to the user
-  local test="$1"
+userinput_func() { # userinput function to display yad/cli prompts to the user
+  local text="$1"
   [ -z "$text" ] && error "userinput_func(): requires a description"
   shift
   [ -z "$1" ] && error "userinput_func(): requires at least one output selection option"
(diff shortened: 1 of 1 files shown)


theofficialgman:
This would almost certainly break every dialog in the Minecraft Java Server installation script (what the list mode was tuned for) https://github.com/Botspot/pi-apps/blob/master/apps/Minecraft%20Java%20Server/install . Under most uses of userinput_func in that script, each dialog can have dozens of choices (hence why a scrolling list is necessary, buttons would simply go off screen).

I am not opposed to using the screen size calculation code that was in the script to determine how many buttons can be fit on the screen before it needs to change to a radio button list (for most screen sizes it will probably be somewhere between 3 and 10).

Botspot:
Fixed
image

image

Botspot:
Now the only possible issue I can think of is that maybe an older yad version (the one on your Ubuntu system that needed a hardcoded text color) would not support these flags.

theofficialgman:
I had no idea that the `--form` option existed and that it was so extendable.

theofficialgman:
> Fixed image

You removed `--fixed` from the non-scrolling list option. As you probably noticed, yad's window sizing is not great without `--fixed`. I think its worth adding back (at least for the non-scrolling option) unless you noticed other issues with it?

Botspot:
> You removed `--fixed` from the non-scrolling list option. As you probably noticed, yad's window sizing is not great without `--fixed`. I think its worth adding back (at least for the non-scrolling option) unless you noticed other issues with it?

I removed it thinking that it would be nice for users to resize the window if they saw fit, though you are right that with this yad version it causes unused empty space. Will add it back.

PR #2840, opened 2025-09-28
Update Java version to Temurin 21 for Unciv

touhidurrr:
https://github.com/yairm210/Unciv/pull/13976


touhidurrr:
this has been deemed to be overkill.

PR #2930, opened 2026-03-23
Use pi-apps tmpdir to avoid tmpfs size limit on debian trixie

Botspot:
Pi OS Trixie seems to have followed Debian in making /tmp a tmpfs by default. This tmpfs is governed by systemd tmp.mount and seems to be allocated to 50% of total RAM size. This is fine on SBCs with 2+ GB of ram, but someone's Pi3 failed to install Wine due to an "out of space" error (which was diagnosed by `log_diagnose` as "please free up disk space", thus preventing this and all other similar errors from being sent)

image

This changes most code paths that rely on /tmp, to instead use $PI_APPS_TMP, which is one of either /tmp or /var/pi-apps-tmp if the mergefs is active.

If these changes prove too substantial, a possible alternative may be to forcibly `sudo systemctl mask tmp.mount` and try to undo the /tmp mount on a running system, but I can't think off a safe and elegant way to do that without risking collateral damage.

### DIFF
--- a/api
+++ b/api
@@ -467,7 +467,7 @@ repo_rm() { #remove the local apt repository
   rm -rf /var/cache/pi-apps/pi-apps-local-packages || sudo rm -rf /var/cache/pi-apps/pi-apps-local-packages || error "repo_rm(): failed to remove the local repository: /var/cache/pi-apps/pi-apps-local-packages"
   
   #Also remove broken symbolic link to /var/cache/pi-apps/pi-apps-local-packages/./Packages
-  sudo rm -f /var/lib/apt/lists/_tmp_pi-apps-local-packages_._Packages
+  sudo rm -f /var/lib/apt/lists/_var_cache_pi-apps_pi-apps-local-packages_._Packages
 }
 
 app_to_pkgname() { #given an app-name, convert it to a unique, valid package-name that starts with 'pi-apps-'
@@ -705,7 +705,7 @@ Package: $package_name" > /var/cache/pi-apps/$package_name/DEBIAN/control
       local exitcode=$?
       status "Apt finished."
       
-      if [ "$using_local_packages" == 1 ] && [ ! -f /var/lib/apt/lists/_tmp_pi-apps-local-packages_._Packages ] && [ $i != 5 ];then
+      if [ "$using_local_packages" == 1 ] && [ ! -e /var/lib/apt/lists/_var_cache_pi-apps_pi-apps-local-packages_._Packages ] && [ $i != 5 ];then
         #another apt update process deleted apt's knowledge of the pi-apps local repo. Warn the user and try again, up to 5 tries.
         i=$((i+1))
         warning "Local packages failed to install because another apt update process erased apt's knowledge of the pi-apps local repository.\nTrying again... (attempt $i of 5)"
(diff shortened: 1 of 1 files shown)


Botspot:
@theofficialgman Interested to know if you see any risks or alternatives to this

theofficialgman:
lets refer to the linux foundation's filesystem hierarchy standard https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html and our knowledge of how other applications handle files like this (eg: apt `.deb` file downloads before they get installed).

In general, our use of `/tmp` for storing temporary "small" files is valid and I don't think that should change (so no need to modify 91 files).

Apt deals with files that are "larger" (all `.deb` files downloaded) by placing them in `/var/cache/apt` (and nested) folder which can get cleaned up automatically by a number of scripts/programs depending on how the system is configured. If you refer to the filesystem hierarchy standard above, this makes sense to me, as `/var/cache` is intended to be used by programs "as a result of time-consuming I/O or calculation". folders in this directory seems to be owned by multiple users, we could create a folder in this directory for pi-apps owned by whatever user we want that has read/write permissions by all users.

example of some on one of my servers:
```
/var/cache$ ls -la
total 112
drwxr-xr-x 22 root          root           4096 May 25  2025 .
drwxr-xr-x 14 root          root           4096 May  6  2024 ..
drwxr-xr-x  2 root          root           4096 Nov 28  2022 adduser
drwxr-xr-x  9 root          root           4096 Mar 26  2025 apparmor
drwxr-xr-x  3 root          root           4096 Apr 14  2023 app-info
drwxr-xr-x  3 root          root           4096 Mar 28 15:43 apt
drwxr-xr-x  2 colord        colord         4096 May  6  2024 colord
drwxr-xr-x  2 root          root           4096 Jun 21  2023 cracklib
drwxrwx---  3 root          lp             4096 Mar 24 06:01 cups
drwxr-xr-x  3 root          root           4096 Mar 27 06:49 debconf
drwxr-xr-x  2 root          root           4096 Apr 14  2023 dictionaries-common
drwxr-xr-x  2 root          root          24576 May 10  2025 fontconfig
drwxr-xr-x  2 root          root           4096 Mar 28 17:23 fwupd
drwxr-xr-x  2 fwupd-refresh fwupd-refresh  4096 Jul 15  2023 fwupdmgr
drwx------  2 root          root           4096 Mar 27 06:48 ldconfig
drwxr-xr-x 44 man           man            4096 Mar 28 11:24 man
-rw-r--r--  1 root          root            216 Mar 28 16:52 motd-news
drwxr-xr-x  3 root          root           4096 Apr 14  2023 PackageKit
drwxr-x---  2 pollinate     daemon         4096 May 25  2025 pollinate
drwx------  2 root          root           4096 Apr 14  2023 private
drwxr-xr-x  2 root          root           4096 Mar 31  2023 samba
drwxr-xr-x  5 root          root           4096 Mar 28 07:04 snapd
drwxr-xr-x  3 root          root           4096 Apr 14  2023 swcatalog
```

I would then suggest using that directory for "I/O intensive tasks" like any temporary downloads in `install_packages` and selectively for any scripts that download "large" files. "large" being defined by the least common denominator as "500MB" (50% of RAM of our minimum supported system). this way only a small handful of scripts need modifying and that number will always stay small (since very few scripts download 500MB+ worth of stuff).

PR #2931, opened 2026-03-24
Update Notepad++ and DDNet versions

ferranox:
- Update Notepad++ from 8.5.3 to 8.9.3
- Update DDNet from 18.6 to 19.8

I was going to update Flameshot but they no longer support Linux ARM64 architecture with their latest update.

### DIFF
--- a/apps/Flameshot/install-64
+++ b/apps/Flameshot/install-64
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-version=13.3.0
+version=12.1.0
 install_packages https://github.com/flameshot-org/flameshot/releases/download/v${version}/flameshot-${version}-1.debian-10.arm64.deb || exit 1
 
 echo '#!/bin/bash


Botspot:
Thanks for this PR.

We at pi-apps have a [github actions bot](https://pi-apps.io/wiki/development/DOCUMENTATION/#automatic-app-updaters) which should keep most of the apps with `version=` variables updated automatically. I wonder if the ARM64 runners would have that same issue now.

Both of these apps have that, and yet they are out of date. DDNet is an easy answer - the updater was [disabled](https://github.com/Botspot/pi-apps/commit/ca4b31a88dd98f5996307467a61ae3b83c31a1d2) 3 years ago because compiling in the virtual environment was crashing the runner.

However Notepad ++ is much more perplexing. I see in the actions run that the script identifies that v8.9.3 is available, yet it is seemingly never tested or put into either failed_apps or updated_apps.
image
~~The update script isn't disabled, and notepad++ is not mentioned anywhere in #2539, so I'm not sure what is going on.~~

Looks like a new error in the runner that has disabled all app updates. Could you take a look at it @theofficialgman?
image

theofficialgman:
> Looks like a new error in the runner that has disabled all app updates. Could you take a look at it @theofficialgman?

@Botspot Those errors are non-fatal (they don't block any updates). Those errors just happen if there are still active reads/writes/locks on the chroot which causes the directory unmount to fail. Its a non-issues since it just bails out while still allowing any updates to proceed.

The reason Notepad++ wasn't getting updated is because its update script is wrong. it has an `armhf_url` when there is only an `install` script. so the updater just skipped doing anything since there was no matching file to operate on. There should always be a "current version" listed afterwards when it finds a file to operate on and in this case there wasn't. I have corrected this here https://github.com/Botspot/pi-apps/commit/74efa7b92c0bfd5d60e5a1b0b2581a7ad4430a82

image

PR #2935, opened 2026-03-27
Use wrapper script instead of symlink to create executable in /usr/local/bin.

ferranox:
I think this is a better and cleaner approach to the problem I mentioned earlier. Instead of copying the executable in the the system path /usr/local/bin, this creates an wrapper script in /usr/local/bin that launches the app in /usr/games/tetris/tetris. This means that the apps that are run by the terminal command and from the app menu are exactly the same rather than being copies of the same app. The uninstall script can stay the same because the file in /usr/local/bin is called the same thing. Thanks.

### DIFF
--- a/apps/Tetris CLI/install-32
+++ b/apps/Tetris CLI/install-32
@@ -1,4 +1,10 @@
 #!/bin/bash
 
 install_packages https://github.com/Pi-Apps-Coders/files/raw/main/tetris-cli_2.0-1_armhf.deb || exit 1
-sudo ln -s /usr/games/tetris/tetris /usr/local/bin/tetris-cli
+
+sudo tee /usr/local/bin/tetris-cli > /dev/null << 'EOF'
+#!/bin/bash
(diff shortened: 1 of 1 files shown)


Botspot:
I'm assuming you closed this PR because you realized, but:
`ln -s` makes a symlink, not a copy of the binary, which is in effect a wrapper script already.

ferranox:
Yeah sorry - I didn't realise that that was the shortcut command and not copy because I just saw a new executable in /usr/local/bin and though it would be better to make a wrapper script. I guess it pretty much does the same thing though so you could still change it if you want.

Botspot:
Nah it's fine as-is - in fact a symlink is even more elegant than a wrapper script because it avoids spawning a bash process. (1 PID is better than 2 🤣 )

ferranox:
That makes sense - thanks and sorry for opening this in the first place lol. Out of interest are you considering adding Anki or kitty to Pi-Apps as I have not heard from you on that?

PR #2938, opened 2026-03-27
Add kitty app

ferranox:
This just adds the kitty app discussed in issues. I have given the user two options: 'No, do not change anything.' or 'Yes, set as default terminal system-wide (recommended).' as you asked. The scripts are otherwise the same.

I didn't really know whether I should also add kitty to /etc/categories in the pull request so I played it safe and didn't, but you can add it if you want to - it would (quite obviously) go into the Terminals category.

Thanks, Alex :)

### DIFF
--- a/apps/kitty/uninstall
+++ b/apps/kitty/uninstall
@@ -9,12 +9,20 @@ fi
 
 sudo rm -rf "$INSTALL_DIR" || error "Failed to remove $INSTALL_DIR"
 
-sudo rm -f /usr/local/bin/kitty || true
-sudo rm -f /usr/local/bin/kitten || true
+sudo rm -f /usr/local/bin/kitty
+sudo rm -f /usr/local/bin/kitten
 
-rm -f ~/.local/share/applications/kitty.desktop || true
-rm -f ~/.local/share/applications/kitty-open.desktop || true
+sudo rm -f /usr/local/share/applications/kitty.desktop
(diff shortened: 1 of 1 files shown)


Botspot:
> I didn't really know whether I should also add kitty to /etc/categories in the pull request so I played it safe and didn't, but you can add it if you want to - it would (quite obviously) go into the Terminals category.

Yes this needs a etc/categories entry

I'm about to leave on some errands, but once I get back I plan to make an auto-updater script to keep the version variable up to date. If you've got nothing to do, feel free to give it a shot yourself if you want. Otherwise don't bother

ferranox:
I have never made a workflow script before. It seems pretty simple but please do correct it if it is wrong (it probably is).

ferranox:
Again I'm not completely sure on this but I'm pretty confident this is right. Again please let me know if I'm not supposed to do this or I'm also supposed to do something else.

ferranox:
Also I noticed you haven't made an auto-updater script for En Croissant. I could also give that a shot but I probably would get it wrong. Let me know if you want me to do this or if you can do it yourself or something else.

Botspot:
> Also I noticed you haven't made an auto-updater script for En Croissant. I could also give that a shot but I probably would get it wrong.

I'll give it my best shot later. Actually we'll be merging this PR without testing the updater script, because the updater workflow is somewhat broken right now. No apps have been getting auto-updated for a few days now due to it 🤣

PR #2944, opened 2026-03-31
Add Joplin

jose1711:
Joplin is a popular open-source note-taking and to-do application with Markdown support, notebook organization, and sync capabilities.

**Why build from source?**
Joplin does not provide official ARM Linux binaries (arm64 build is provided only for the server part). Building requires fpm to produce the .deb package but Joplin's bundled fpm is x86_64-only. The build therefore uses a Bullseye chroot to provide a clean, isolated environment with system fpm via gem install, avoiding pollution of the host system with Ruby and build dependencies.

**Requirements**
  - ~15 GB free disk space during build (chroot + node_modules); the installed package is ~700 MB
  - Build time: ~1 hour on Raspberry Pi 4

**Testing**
Tested on 64-bit (arm64, Trixie). The 32-bit code path is included but untested - feedback welcome.

**Credits**
Based on the build steps shared by MichaelDavidHarry, adapted for Pi-Apps.


theofficialgman:
@jose1711 please try to get this supported by upstream first. its free to build on arm64 linux runners on github actions (what they are already using for official releases) and generally pretty easy to extend existing CI to do so

theofficialgman:
@jose1711 any progress on upstream efforts?

PR #2957, opened 2026-04-12
Add Flathub app??

ferranox:
This is one of the best Linux app stores (as well as Pi-Apps of course XD) in existence. I'd imagine you probably use it already as it is very popular and has a lot of apps for ARM64 architecture (a couple 32 bit as well but not a lot - the app itself can still install on 32 bit architecture and I have mentioned this in the app description).  I think that, because of this, Flathub should definitely be on Pi-Apps to present users with the option to download Flathub apps.

The only problem is that Flathub doesn't have a GUI for Raspberry Pi OS. I have also mentioned this in the description and provided a link to the website in which you can browse the apps that can be installed via Flathub (so it is still technically appropriate for everyday GUI Raspberry Pi users which is the criteria in the creating an app documentation - it's just that the app that doesn't have a GUI in and of itself).

The thing is, on Ubuntu, you can add a plugin (and they do recommend to do so in the official install instructions for Ubuntu on the Flathub website) for GNOME Software to add the catalogue of Flathub apps in that and it gives a GUI for installation. GNOME software isn't installed by default on Raspberry Pi OS but I'm guessing it could be installed with `sudo apt install gnome-software` (or `install_packages gnome-software` in a Pi-Apps script). I haven't tested this because I don't use Raspberry Pi OS but I assume you could. I could potentially give the user an option to install GNOME software which would give them a GUI for the app but this is a bit odd to install a different app than the one that they tried to install but it could be the best way forward. I have not done this in the pull request at the moment but I have allowed edits by maintainers if you think this would be the best option.

Another great option in my opinion would be to not add Flathub at all but add GNOME Software because it is an excellent piece of software on its own and then give the user to install the Flathub plugin during installation. If you think this would be the best option then let me know and I can close this pr and open a new one for gnome software (it would be a package app but isn't if you want to give the user the option to add the Flathub plugin).

Let me know what you think - if you don't want it on Pi-Apps at all then that would be fine but I think it would be a great thing to have.

Thanks, Alex


Botspot:
Pi-Apps has Plasma Discover, which is preconfigured to use flathub.
image


Gnome Software is actually on pi-apps right now, but it is hidden on raspberry pi OS since 2023 due to it conflicting with the preinstalled pi-packages program. See https://github.com/Botspot/pi-apps/commit/6e008290ba7e3bc45aaee10758ed3de35b51c122

Botspot:
Closing this PR, as it does not align with the eligibility rubric, as installing and configuring flathub is not directly useful for end users.

PR #2962, opened 2026-04-24
Update En Croissant install

ferranox:
Before this update, En Croissant built itself locally, which not only took ages but also didn't work on machines with Pi OS Trixie because it was too large for the /tmp folder stored in ram. I have changed the script completely so that it downloads prebuilt files from a GitHub repo I made and just installs them locally. This also needs less dependencies as half of them before were for building, and it takes a fraction of the time. I have attempted an updater script but I cannot confirm it will work. If you think there are any mistakes just let me know.


Botspot:
Looks good. This successfully installed on bookworm 32 and 64-bit, and trixie 64-bit, but failed on trixie 32-bit. We don't officially support trixie 32-bit, but could you look at this to see if it's an easy fix?
image
Here's the test run link: https://github.com/Botspot/pi-apps/actions/runs/25023816142/job/73290344187

ferranox:
Okay this definitely seems more like an issue with your tester for Trixie armhf. Looking at your screenshot, you can see that the issue is that it can't find these dependencies: `libatk1.0-0`, `libglib2.0-0`, and `libgtk-3-0`. These are absolutely essential for a GTK app like En Croissant. The app will literally not even open without them so removing them is not an option.

The bigger issue is that this error is NOT normal. One line in your failed installation reads `Depends: libgtk-3-0 but it is not installable` which literally means that APT literally cannot find any version of this package for your system which is **very** wrong because GTK3 _should exist everywhere_.

By far the most likely problem would be that your Trixie armhf repo setup is incomplete or wrong. Another option is that Trixie armhf is incomplete - some packages simply aren’t built yet for armhf. This does happen sometimes because Debian prioritises amd64 and arm64 before everything else, but it is very unlikely.

You should probably try using the actual OS rather than the tester because it seems to me that it doesn't have these packages which would be available in the actual OS. At any rate it is not a problem with the scripts or something I can fix.

Thanks, Alex

ferranox:
Okay I just tried installing it on a fresh install of Trixie 32 on my Pi 5 and it had the same error so it must be something weird going on with trixie. I think we should probably just make the changes anyway considering 32 bit trixie is no longer supported and I don't think that the people who use 32 bit trixie would if they wanted it to play chess.

Botspot:
These types of issues were why I decided to drop trixie 32-bit support. Merging now 👍

PR #2972, opened 2026-05-16
Add RQuickShare Pi

EladBG-code:
## Summary

Adds RQuickShare Pi as an install-based ARM64 desktop app.

RQuickShare Pi is a Raspberry Pi OS ARM64-focused Quick Share desktop app for sending and receiving files with Android Quick Share devices. It uses a GUI, tray behavior, Bluetooth discovery, and mDNS networking.

## App Type

Install-based app with `install-64` only. ARM64-only apps are allowed per the 2026 update in the app eligibility rubric.

## Files

- `apps/RQuickShare Pi/icon-64.png`
- `apps/RQuickShare Pi/icon-24.png`
- `apps/RQuickShare Pi/website`
- `apps/RQuickShare Pi/description`
- `apps/RQuickShare Pi/credits`
- `apps/RQuickShare Pi/install-64`
- `apps/RQuickShare Pi/uninstall`
- `etc/categories`

## Install Behavior

The install script uses `install_packages` to install:

- `bluez`
- `avahi-daemon`
- the public ARM64 `.deb` release from GitHub

It then tries to enable/start Bluetooth and Avahi so discovery works after install.

## Uninstall Behavior

The uninstall script stops the app if running, removes stale autostart launchers, and calls `purge_packages`. It leaves user settings and received files untouched.

## Validation

- `bash -n 'apps/RQuickShare Pi/install-64' 'apps/RQuickShare Pi/uninstall'`
- `shellcheck 'apps/RQuickShare Pi/install-64' 'apps/RQuickShare Pi/uninstall'`
- `identify 'apps/RQuickShare Pi/icon-64.png' 'apps/RQuickShare Pi/icon-24.png'`
- Pi-Apps API detected `script_name=install-64`
- Release `.deb` URL responds successfully


EladBG-code:
Small update: I refreshed the v0.0.1-alpha release asset at the same URL after testing on Raspberry Pi OS. The refreshed build fixes a WebKitGTK first-frame rendering issue on Pi where the app could appear visually corrupted until the window was resized. The Pi-Apps install script URL did not change.

apps/RQuickShare Pi/uninstall
```diff
@@ -0,0 +1,16 @@
...
+if command -v systemctl >/dev/null 2>&1; then
+  systemctl --user stop rquickshare-pi-dev.service 2>/dev/null || true
+fi
+
+if command -v pkill >/dev/null 2>&1; then
+  pkill -x rquickshare-pi 2>/dev/null || true
+fi
+
+rm -f "$HOME/.config/autostart/RQuickShare Pi.desktop"
+rm -f "$HOME/.config/autostart/rquickshare-pi.desktop"
```
Botspot: What step in the installation process created these files?

Botspot:
What advantages does this have over a pi-apps install script which clones the upstream rquickshare repo from source and compiles it from source? Does your repo make any specific workarounds or usability improvements compared to upstream, or is the arm64 deb release the only value your repo provides?

apps/RQuickShare Pi/uninstall
EladBG-code: Good catch. The Pi-Apps install script does not create those files directly.

The current app can create `$HOME/.config/autostart/RQuickShare Pi.desktop` after first launch when its in-app `Start on boot` setting is enabled. That autostart file runs `/usr/bin/rquickshare-pi --rquickshare-pi-autostart`, so if the package is later removed and the autostart file is left behind, Raspberry Pi OS will try to launch a removed binary on login.

I pushed an update to tighten this uninstall script: removed the dev-service cleanup and the legacy autostart filename, and left only the current app-created autostart file with an explanatory comment.

EladBG-code:
Thanks, fair question. The ARM64 `.deb` is not the only value here. A Pi-Apps script that clones **upstream** RQuickShare and builds it would not include the Pi-specific changes in this fork.

Concrete differences in this fork:

- Real target validation: the app is built, installed, launched, and tested on a Raspberry Pi 5 running Raspberry Pi OS 64-bit/ARM64. The release asset is the exact ARM64 Debian package tested on the target hardware.
- Raspberry Pi WebKitGTK workaround: the Linux binary sets `WEBKIT_DISABLE_COMPOSITING_MODE=1` before Tauri creates the WebKit webview. This fixes a Raspberry Pi OS rendering bug where the first visible window could appear as corrupted horizontal stripes until resized.
- Android-to-Pi receive discovery fixes:
  - The BLE listener scans broadly and handles Samsung Quick Share manufacturer-data hints, because on the tested Samsung S26 Ultra the send sheet did not always expose the expected service data while searching.
  - The BLE advertiser publishes Nearby/Quick Share FE2C service data and falls back from `Broadcast` to `Peripheral` when BlueZ on the Pi rejects broadcast advertising. This fallback was added because the Pi produced `Failed to register advertisement` in real testing.
  - Advertising follows the app visibility state, so `Always visible` actually controls whether the Pi is discoverable.
- Pi desktop integration: AppIndicator tray support, Pi-friendly tray icon handling, start-hidden/start-on-boot behavior, and single-instance handling so autostart does not pop a window unexpectedly.
- Runtime diagnostics: the app checks and prompts for Bluetooth service, Avahi/mDNS, rfkill state, and adapter power issues that commonly block discovery on Raspberry Pi OS.
- User-facing Pi docs: the README/wiki document Raspberry Pi OS dependencies, Samsung's `Share with Apple devices` caveat, logs, and troubleshooting commands.

Building from source inside Pi-Apps would also require installing Rust, Node.js, pnpm, protobuf, GTK/WebKit dev packages, and then doing a full Tauri release build on the user's Pi during install. That is slow and more failure-prone than installing the tested `.deb`. Source remains public and GPL-3.0; the `.deb` is just the tested install artifact for Pi users.

EladBG-code:
Also, on a personal note: I really appreciate the work you have put into Pi-Apps. It is one of the things that makes Raspberry Pi OS feel approachable for normal desktop users, not just people who are already comfortable building software from source.

That is the spirit I am aiming for with RQuickShare Pi too: take something useful, test it on real Pi hardware, package it cleanly, and make it easier for Raspberry Pi users to actually use. I would be glad if this project can eventually fit well into the Pi-Apps catalog.

EladBG-code:
One more bit of context on the motivation: before this fork, I could not find a practical Raspberry Pi OS ARM64 path for using Android Quick Share on a Pi as a normal desktop user.

Upstream RQuickShare is the foundation this is based on, but it did not provide a Pi-focused, hardware-tested ARM64 package, Raspberry Pi OS runtime checks, or the Pi/Samsung receive-discovery fixes this fork now carries. The goal here is not just to make a `.deb`; it is to make Android Quick Share usable on a real Raspberry Pi desktop without expecting each user to debug Bluetooth, mDNS, WebKitGTK, and Tauri build issues themselves.

I had to recode a lot of the original code to match the Pi's architecture so the average user can't just use the original repository (rquickshare) as it doesn't really have a version for the Pi...

EladBG-code:
Plus my version has some QOL improvements over the upstream because mine has several like the ability to change the name of the Pi which is visible to nearby QuickShare devices and some graphical improvements in the rendering of the software itself - more stable, less prone to crashes and just an overall version for Pi (which we didn't have before for our processor architecture).

I am planning to add more things in the futrure to make it even more optimized for Pi but as this is version alpha I just thought to myself - I had to release it to the public so they would also enjoy it and eventually I'll test it on the 32bit version as well.

Botspot:
It seems like your fork's stated changes boil down to 2 potential benefits and 2 drawbacks.

Benefits:

- Your fork hosts a precompiled arm64 binary.
- Your fork has some tweaks to make bluetooth work with your phone. (unclear if those changes are needed to fix an issue with raspberry pi hardware, or with your specific phone's hardware)

Drawbacks:

- Your fork is unlikely to be kept up to date with upstream RQuickShare, which could include critical bugfixes, vulnerability patches, or support for newer android devices and protocols.
- Your fork's changes to the bluetooth scanning protocol could introduce regressions for other android devices, or could break the implementation on other ARM devices that Pi-Apps is commonly used on.

Without paraphrasing, what **_exactly_** are your bugfixes doing? How do they work? Do you know if they fix an issue with raspberry pi hardware, or with the quick share protocol on the specific samsung phone and android version you were testing this on? What is the feasibility of getting those fixes merged upstream, so that you are not having to pull upstream changes and ensure that your diverging changes continue to work?

EladBG-code:
Thanks. I agree this needs a more exact answer than my earlier summaries.

The fork is not just "upstream plus a prebuilt .deb", but I also do not want to overstate the scope. Some changes are Raspberry Pi OS / BlueZ / WebKitGTK specific, and some are compatibility behavior observed with Samsung Quick Share devices. Here is the exact breakdown from the current public source.

### Raspberry Pi / Raspberry Pi OS specific fixes

1. **WebKitGTK first-frame rendering workaround**

   On Raspberry Pi OS, the Tauri/WebKitGTK window could open as corrupted horizontal stripes until the window was manually resized. The app now sets `WEBKIT_DISABLE_COMPOSITING_MODE=1` before Tauri creates the webview, and the Linux show-from-tray path nudges the window size by 1px and restores it to force a repaint.

   This is not a Quick Share protocol change. It is a Raspberry Pi OS WebKitGTK/compositor workaround found by testing on real Pi hardware.

2. **BlueZ BLE advertising fallback**

   The fork adds a Linux BLE advertiser for the Nearby/Quick Share `FE2C` service data. On the Pi, BlueZ rejected the preferred `Broadcast` advertisement path during testing with errors around advertisement registration. The code retries as a `Peripheral` advertisement, marks it discoverable, supplies a local name, and keeps advertising tied to the app visibility state.

   This is mainly a Linux/BlueZ adapter behavior fix. It was validated on Raspberry Pi 5 with Raspberry Pi OS 64-bit. I am not claiming it is validated on every ARM board Pi-Apps can run on.

3. **mDNS address/interface handling**

   The fork makes mDNS advertising more deterministic on Linux desktops by choosing a physical LAN IPv4 address, preferring Wi-Fi before wired, filtering virtual interfaces such as Docker/VPN/tap/tun/veth-style interfaces, and rebuilding the service when the user changes the visible device name.

   This is meant to avoid Android discovering an mDNS service that resolves to the wrong or unusable local address. This is a good upstream candidate because it is not Pi-branded behavior by itself.

### Quick Share compatibility changes

4. **Android/Samsung discovery trigger**

   Upstream listened for BLE events filtered to the `FE2C` service UUID. On the Samsung S26 Ultra I tested with, the Android send sheet did not always expose that service-data event while searching. Sometimes the observable event was Samsung manufacturer data instead.

   The fork broadens the scan, but it does not accept every BLE event. It only uses either:

   - `FE2C` service-data advertisements, or
   - Samsung manufacturer data, company id `117`, as a hint that a Samsung Quick Share sender is actively searching.

   That hint only causes the app to register/resend the mDNS service, throttled to once every 5 seconds. It does not auto-accept transfers, does not connect over BLE, and does not change the actual transfer authentication path.

   This part is based on Samsung Quick Share behavior observed on my test devices. I would not describe this part as proven to be a Raspberry Pi hardware issue.

5. **Outbound completion / disconnect behavior**

   When sending from the Pi to other devices, some receivers copy the file bytes successfully but still show a failure if the sender closes too quickly after the final payload. I reproduced that with the official Windows Quick Share receiver and with Android receivers.

   The fork now handles payload ACK/control packets, disconnection frames, and safe-to-disconnect negotiation. For Windows laptop targets it sets `safe_to_disconnect_version` and waits for the receiver path instead of immediately closing. For Android phone targets it waits longer for receiver completion before disconnecting.

   This is why the Windows receiver stopped showing "Can't complete transfer" while the file was already present in Downloads.

6. **File metadata compatibility**

   Some receivers are picky about file type metadata. SVG/generic files were a real example: the file bytes arrived, but the receiver UI still reported failure. The fork sends safer metadata for those cases, for example treating SVG/generic Android-bound files as document-like payloads with `application/octet-stream` instead of forcing image metadata that the receiver rejects.

### Risk and scope

Your regression concern is fair. I would describe the Pi-Apps entry as an alpha Raspberry Pi OS ARM64 app, not a generic ARM Quick Share implementation.

To reduce Pi-Apps risk, I am happy to tighten the Pi-Apps install script further so it only installs on Raspberry Pi OS 64-bit / Raspberry Pi hardware, instead of merely checking `arm64`. That would avoid implying support for other ARM boards where I have not tested the BlueZ and desktop behavior.

The Samsung manufacturer-data path is also intentionally narrow and throttled, but if you prefer I can make it stricter before this is accepted, for example by parsing the manufacturer payload shape more carefully instead of only checking Samsung company id `117`.

### Upstream feasibility

I do think several of these should be proposed upstream, but not as one large Pi-branded patch. The realistic upstream path is to split them into small PRs/issues:

- BlueZ advertisement fallback on Linux.
- More deterministic mDNS address/interface selection.
- Payload ACK / safe-disconnect handling for receivers that complain after successful byte transfer.
- Endpoint-info/device-name cleanup.
- Samsung discovery behavior as a separate issue or experimental PR, because the evidence I have is from Samsung devices rather than all Android Quick Share implementations.

So the long-term plan should not be permanent divergence from upstream. The fork exists because these changes needed real Raspberry Pi OS ARM64 testing and a tested `.deb` for normal Pi users. I will keep tracking upstream and try to upstream the general fixes where they make sense.

theofficialgman:
ugh... can we not use webkit for simple apps like this? this is so unnecessarily ram and processing heavy.

would be much better to have a native app implementation (like QNearbyShare https://github.com/vicr123/QNearbyShare) but it needs to be updated.

theofficialgman:
Which is coincidentally exactly what Packet is https://github.com/nozwock/packet/tree/main (can be installed from the official flathub build)
a maintained project that uses a maintained fork of the rust backend from QNearbyShare while using a native GTK UI

theofficialgman:
Tested Packet out on Nintendo Switch to my android phone and works well.

Closing this PR in favor of that given the issues @Botspot has already outlined. If you have features you would like to add to Packet, I encourage you to contribute upstream.

EladBG-code:
Thanks for looking at alternatives. I want to clarify a few points for the record, because I do not think Packet existing makes this submission technically equivalent or removes the value of RQuickShare Pi.

First, I agree with the general GTK point: a native GTK UI is lighter than a Tauri/WebKitGTK UI. I am not going to argue otherwise. RQuickShare Pi currently inherits the upstream RQuickShare architecture, and replacing the frontend stack with GTK would be a significant rewrite rather than a small Pi-Apps packaging change. If demand for the project grows, I am open to moving more of the UI toward a lighter native implementation, but the first goal here was to make the existing RQuickShare codebase actually usable, packaged, and tested on Raspberry Pi OS ARM64.

Second, Packet is a good project, but it is not the same submission. Packet's own README describes it as a partial Quick Share implementation, and its current Pi-Apps entry is a Flatpak install. RQuickShare Pi is a native ARM64 Debian package targeted specifically at Raspberry Pi OS desktop users, with Pi-focused behavior around BlueZ, Avahi/mDNS, tray/start-hidden desktop integration, autostart, runtime diagnostics, WebKitGTK-on-Pi rendering workarounds, and user-facing Raspberry Pi troubleshooting docs. Those are not just cosmetic differences.

Third, testing Packet on a Nintendo Switch to an Android phone is useful evidence that Packet works in that path, but it does not validate Raspberry Pi OS 64-bit desktop behavior, Raspberry Pi Bluetooth/BlueZ behavior, Pi-Apps install/uninstall behavior, or the exact Android/Windows interoperability issues this fork has been working through. RQuickShare Pi has been tested on real Raspberry Pi 5 hardware with Raspberry Pi OS ARM64, Android Quick Share devices, and Windows Quick Share receivers.

Finally, I am not against upstream work. The general fixes should be split and proposed upstream where they make sense: mDNS/interface selection, safer transfer completion/disconnect handling, and Linux/BlueZ behavior are all candidates. But saying “contribute upstream” does not solve the immediate Pi user problem by itself, especially when some of the work is Raspberry Pi OS specific packaging, runtime checks, and desktop integration.

If Pi-Apps prefers to include Packet as one Quick Share option, that is reasonable. I just do not think RQuickShare Pi should be dismissed solely because another Quick Share client exists or because a future native GTK rewrite would be lighter. The projects make different tradeoffs, and this one exists specifically to provide a tested Raspberry Pi OS ARM64 path.

theofficialgman:
@EladBG-code Please test packet and report back with any deficiencies you have. If you test the software, you will see that is already has support for most of the features you report that RQuickShare Pi has. The official readme for Packet just needs some updates, it is well supported and implements more quickshare than the upstream RQuickShare project even does (again, it uses a fork of RQuickShare as its backend).

Your responses sound a lot like AI slop and I'm inclined to believe that you haven't even tested Packet or RQuickShare to understand what either of their limitations are.

EladBG-code:
I personally go over everything I respond with it might sound "Ai" to you because I use Grammarly but that's it.

I will go over Packet but regardless I don't know if there's any legitimacy for even getting FlatHub or any Gnome software for that matter for such light computer boards such as the Raspberry Pi...

Regarding the AI comment I'm gonna let that slide because I'm sure you meant no harm but please be more considerate next time you respond especially because I did work my way on this project and if the current architecture is too heavy to your liking I can EASILY replace the UI toolkit you've mentioned.

PR #2982, opened 2026-05-31
VARA FM: New app

joltcan:
Adds **VARA FM**, the shareware FM software modem for amateur radio messaging (companion to the existing VARA HF app).

**Implementation**
- Installs Wine via the Pi-Apps Wine installer (x86/x64), then `vb6run pdh_nt4 sound=alsa` via winetricks.
- Downloads the installer from winlink.org and extracts the program files with `innoextract` (no GUI installer needed), then registers the VB6 OCX/DLLs with `regsvr32`. Same proven approach as VARA HF.
- The version is detected dynamically from the winlink.org directory listing, so it tracks upstream releases automatically rather than being hardcoded.

**Dependencies installed:** `innoextract`, `libsdl2-2.0-0`, `libxkbregistry0` (plus Wine).

Includes `install`, `uninstall`, `description`, `credits`, `website`, and icons.


Crilum:
The reason for using a hardcoded version variable, is so that pi-apps can trigger updates to the app. (There is a github action that periodically checks for new app versions). With your current implementation the user won't receive updates to the app... (unless the app updates itself)
You should create an update file in `.github/workflows/updates/`.

take a look at one of those files for a reference (e.g. [`.github/workflows/updates/PyChess.sh`](https://github.com/Botspot/pi-apps/blob/master/.github/workflows/updates/PyChess.sh))

Crilum:
For example:

```sh
#!/bin/bash

webVer="$(wget -qO- 'https://downloads.winlink.org/VARA%20Products/' | grep -oP 'VARA%20FM%20\Kv\d+\.\d+\.\d+' | sort -V | tail -1)"

all_url="https://downloads.winlink.org/VARA%20Products/VARA%20FM%20${webVer}%20setup.zip"

```

PR #2983, opened 2026-05-31
Winlink Express: New app

joltcan:
Adds **Winlink Express** (formerly RMS Express), the standard ham radio client for the Winlink radio email network. Works together with software modems such as VARA FM/HF and ARDOP.

**Implementation**
- Installs Wine via the Pi-Apps Wine installer (x86/x64), then `vb6run pdh_nt4 sound=alsa` via winetricks (skipped if a VARA modem already configured the prefix).
- The version is detected dynamically from the winlink.org directory listing.
- Runs the official Inno Setup installer silently through Wine under `xvfb-run` (`/VERYSILENT /NORESTART`), so install-time setup (registry, file placement) is handled by the real installer.

**Dependencies installed:** `xvfb` (provides `xvfb-run`, required to run the GUI installer headlessly), plus Wine.

Includes `install`, `uninstall`, `description`, `credits`, `website`, and icons.


Crilum:
This needs the same changes as your last PR, the app will not update dynamically.

Also.. is this really the app icon?

joltcan:
Ah, sorry, had some issues with my git and figuring out in what order to apply stuff after the merge. Ill sort it out.

Also yes, this is the actual icon :D 

(The software is some terrible win32 legacy crap that somehow become global standard).

PR #2994, opened 2026-06-16
Minecraft Java Server: Various improvements

samsoninja:
Minecraft 26.1 and newer require Java 25; this adds it as an option. Tested and working on a Raspberry Pi 4, Raspberry Pi OS Trixie.

### DIFF
--- a/apps/Minecraft Java Server/install
+++ b/apps/Minecraft Java Server/install
@@ -55,15 +55,15 @@ case "$server_category" in
     wget -O "$server_jar" "https://maven.minecraftforge.net/net/minecraftforge/forge/$installer_version/forge-$installer_version-installer.jar" || exit 1
     ;;
   "Paper")
-    available_versions=$(wget -O- https://api.papermc.io/v2/projects/paper/ | jq -r '.versions | .[]')
+    available_versions=$(wget -O- https://fill.papermc.io/v3/projects/paper/versions | jq -r '.versions.[].version.id')
     description="What version of $server_category do you want to run?"
     table=($available_versions)
     userinput_func "$description" "${table[@]}"
     server_version="$output"
     status "Selected $server_category version $server_version"
-    version_info=$(wget -O- https://api.papermc.io/v2/projects/paper/versions/$server_version/builds | jq -r '.builds | [last]')
-    server_jar="$(echo "$version_info" | jq -r '.[].downloads.application.name')"
-    wget -O "$server_jar" "https://api.papermc.io/v2/projects/paper/versions/$server_version/builds/$(echo "$version_info" | jq -r '.[].build')/downloads/$server_jar" || exit 1
+    version_info=$(wget -O- https://fill.papermc.io/v3/projects/paper/versions/$server_version/builds/latest | jq -r '.downloads."server:default"')
+    server_jar="$(echo "$version_info" | jq -r '.name')"
+    wget -O "$server_jar" "$(echo "$version_info" | jq -r '.url')" || exit 1
     ;;
   "Purpur")
     available_versions=$(wget -O- https://api.purpurmc.org/v2/purpur/ | jq -r '.versions | .[]')
--- a/apps/Minecraft Java Server/install
+++ b/apps/Minecraft Java Server/install
@@ -19,7 +19,7 @@ mkdir -p "$HOME/Minecraft-Java-Server"
 cd "$HOME/Minecraft-Java-Server" || error "Could not enter Minecraft-Java-Server folder."
 
 description="What type of Minecraft server\ndo you want to run?"
-table=("Vanilla" "Fabric" "Forge" "Paper" "Purpur")
+table=("Vanilla" "Fabric" "Forge" "Paper" "Folia" "Purpur")
 userinput_func "$description" "${table[@]}"
 server_category="$output"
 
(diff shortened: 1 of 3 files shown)


LuuOW (commented):
Technical audit: code patterns and architectural layout verified for system reliability.

samsoninja:
List of new changes: Use new Paper API so new versions are supported; Add Folia, a Paper fork with multithreading; Fix spelling errors in JVM dialog

samsoninja:
New feature: use a custom directory when installing (note that this will generate a file at $HOME/Documents/mcserdir.txt so that uninstalling and updating works as expected)

samsoninja:
I've tested this pr and it's now ready to merge, unless someone wants to add another new feature... there may be some edge cases that I didn't test; @Botspot @theofficialgman y'all might want to test it yourselves

apps/Minecraft Java Server/install
```diff
@@ -15,11 +15,26 @@ fi
...
+  # make sure the user wants to install the server in the saved directory, as mcserdir.txt persists on uninstall
+  text="Using the $directory directory.\nIf you would like to use a different directory,\nexit the script and delete the $HOME/Documents/mcserdir.txt file."
+  userinput_func "$text" "Continue with the $directory directory"
+else
+  # allow user to select directory
+  directory=$(yad --center --wrap --text "Choose directory:\nDefault is $directory.\nMake sure it ends with Minecraft-Java-Server if you don't want it to fail." --entry --entry-text="$directory" --button="Submit":0)
+  # store directory in $HOME/Documents so it can be used later
+  echo "$directory" > mcserdir.txt
+fi
+mkdir -p "$directory" || error "$directory is invalid or privileged."
```
theofficialgman: this is a very poor way of keeping the location of a custom directory, it also breaks updates from the old system if the user doesn't understand that they need to select the Minecraft-Java-Server directory. also the comments seem to make no sense as there is no Minecraft-Java-Server directory anymore. please don't vibecode this.... if you really want to allow custom directories, keep the configuration file in the standard `.config` folder in the users home directory

theofficialgman (changes requested):
Please note that all AI assisted development should be indicated should have the linux kernel standard in the commit message like the following example:

```
Assisted-by: Claude:claude-sonnet-4-6
```

apps/Minecraft Java Server/install
samsoninja: I swear I'm not vibecoding... I can fix that but I might just remove the custom directory feature because the person who wanted it wanted to use a custom disk and I don't think that works

samsoninja:
Ok @theofficialgman I've tested my latest changes and it's working on my RPi 4 with Paper and Folia. It's your script, so I want your feedback, but please be constructive this time :). Still 100% AI free.

apps/Minecraft Java Server/install
```diff
@@ -273,7 +298,8 @@ EOF
 screen_location=$(which screen)
 sleep_location=$(which sleep)
 
-sudo tee "/etc/systemd/system/minecraft-server.service" <<EOF
+# this is equivalent to using tee "/etc/systemd/system/minecraft-server.service" and then systemctl daemon-reload
+sudo systemctl edit --stdin --full --force minecraft-server <<EOF
```
theofficialgman: Either a new/old feature of systemd or AI hallucination.

```
systemctl: unrecognized option '--stdin'
```

Please revert back to what this was before. No issue with it.

apps/Minecraft Java Server/install
```diff
@@ -292,10 +318,6 @@ WantedBy=multi-user.target
 EOF
 echo
 
-# update services
-sudo systemctl daemon-reload
-sudo systemctl disable minecraft-server
-
```
theofficialgman, line 298: needs to be re-added. see above

apps/Minecraft Java Server/install
```diff
@@ -194,13 +211,21 @@ case "$java_selection" in
...
+  jvm_args=$(yad --center --wrap --text "Would you like to input any JVM Arguments?\nA default list is prepopulated in the text box below.\n\nThe most important arguments, -Xmx and -Xms, are at the beginning and define the max RAM and initial RAM, respectively.\nYou should specify a value slightly less than your total RAM in bytes, kilobytes(k), megabytes(m), or gigabytes(g).\nHint, use your keyboard Home key to see the start of this list." --entry --entry-text="-Xmx1200m -Xms1200m -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:MaxGCPauseMillis=200 -XX:+UnlockExperimentalVMOptions -XX:+DisableExplicitGC -XX:+AlwaysPreTouch -XX:G1NewSizePercent=30 -XX:G1MaxNewSizePercent=40 -XX:G1HeapRegionSize=8M -XX:G1ReservePercent=20 -XX:G1HeapWastePercent=5 -XX:G1MixedGCCountTarget=4 -XX:InitiatingHeapOccupancyPercent=15 -XX:G1MixedGCLiveThresholdPercent=90 -XX:G1RSetUpdatingPauseTimePercent=5 -XX:SurvivorRatio=32 -XX:+PerfDisableSharedMem -XX:MaxTenuringThreshold=1" --button="Validate JVM Arguments":0)
```
theofficialgman: unless you have zram/swap, suggesting that the user make the Xmx ONLY slightly less than their actual RAM will almost guarantee that the OOM (out of memory) killer will kill the java process. I would recommend setting it to 75% of RAM for a safe general usecase. certain situations could push that further but for the general user we would rather the server run slightly less well (due to less ram) than crash outright (due to out of memory killer).

theofficialgman (changes requested):
see above comments

apps/Minecraft Java Server/install
theofficialgman: I believe you. I was simply referencing the messages on discord regarding this PR from birdman who mentioned vibecoding. Thought you might have taken up the suggestion.

image

apps/Minecraft Java Server/install
samsoninja: hmm I guess it's sorta new

apps/Minecraft Java Server/install
samsoninja: I said slightly but I didn't really mean slightly... I'll change that

samsoninja:
@theofficialgman I made your requested changes... no new changes so it should be good to go

PR #3002, opened 2026-06-29
Updated Install for Java Server

BirdmanAustin-dev:
Changes!
Adds custom install directory support for Minecraft Java Server. Users can now install to the default `~/Minecraft-Java-Server` folder or choose another mounted/custom folder, including an external SSD. The selected path is saved and reused for future updates.

The installer now updates `start-server.sh`, the desktop launcher, and `minecraft-server.service` to use the selected server folder instead of assuming the home directory.

#3003 that it is the companion uninstaller update for #3002.

PR #3003, opened 2026-06-29
Update uninstall

BirdmanAustin-dev:
Changes!
Updates the uninstaller so it detects the saved Minecraft server folder instead of assuming `~/Minecraft-Java-Server`.

It removes the service, desktop launcher, and non-preserved server files from the selected install folder while still preserving important data folders like worlds, mods, plugins, and versions.

#3003 that it is the companion uninstaller update for #3002.

### DIFF
--- a/apps/Minecraft Java Server/uninstall
+++ b/apps/Minecraft Java Server/uninstall
@@ -2,19 +2,83 @@
 
 shopt -s extglob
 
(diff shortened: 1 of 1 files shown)


BirdmanAustin-dev:
Assisted by GPT 5.5 High--I don't know exactly where to put that, but was looking and saw I might need it.

PR #3025, opened 2026-07-30
Update Tor Browser nightly version

amonkarsidhant:
The Tor Browser nightly version was pinned to 2026.07.19 which now returns 404.\n\nUpdated to 2026.07.30 (latest available, verified 200 OK).\n\nThe installer uses a nightly build for arm64 while the 32-bit version uses the stable SourceForge release (13.0.9) which is unaffected.

### DIFF
--- a/apps/Tor/install-64
+++ b/apps/Tor/install-64
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-version=2026.07.19
+version=2026.07.30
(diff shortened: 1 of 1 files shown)


Botspot:
This should be handled by the auto-updater github action. If the auto-updater is failing then that needs fixing. This pr only fixes the surface issue (it's a bandaid fix) rather than the upstream source of the issue.

PR #3027, opened 2026-07-30
Add missing error handling to install_packages calls

amonkarsidhant:
Eight install scripts had `install_packages` calls without `|| exit 1`. If package installation fails, these scripts would continue running with missing dependencies, producing confusing downstream errors.

Files fixed:
- Caprine/install-32, install-64
- OpenSCAD/install-64
- Obsidian/install-64
- Clam Antivirus/install
- Alacritty Terminal/install-32, install-64
- Shellbeats/install

### DIFF
--- a/apps/Alacritty Terminal/install-32
+++ b/apps/Alacritty Terminal/install-32
@@ -1,2 +1,2 @@
 #!/bin/bash
-install_packages "https://github.com/Pi-Apps-Coders/files/raw/main/alacritty_0.11.0-1_armhf.deb" || exit 1
+install_packages "https://github.com/Pi-Apps-Coders/files/raw/main/alacritty_0.11.0-1_armhf.deb"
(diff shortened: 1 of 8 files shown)


Botspot:
Please remove Alacritty and Caprine from this PR. As those `install_packages` commands are the final line in the script, their exit code becomes the exit code of the script, thus making `|| exit 1` redundant.

While one could argue for the changes to be kept for safety or standardization reasons, I don't feel this is worth causing a script change which would trigger a reinstalled app for all users.

Botspot:
OpenSCAD is the same situation - install_packages is the final command

amonkarsidhant:
Removed OpenSCAD from this PR as well -  is the final command in that script too, making  redundant. The remaining changes (Clam Antivirus, Obsidian, Shellbeats) are valid since  is not the final command in those scripts.

PR #3028, opened 2026-07-30
Add error handling to Pi Power Tools wget-and-pipe-to-bash

amonkarsidhant:
The install script downloaded and piped an external script to bash with no error checking. If the download failed, the pipe would silently pass empty data to bash, making the script appear to succeed while doing nothing.\n\nChanged to capture the download in a variable first with proper error handling, then pass to bash.


Botspot:
Please review the change I made to this pr

PR #3031, opened 2026-08-10
Add PiSTG

stghecker:


Botspot:
Has this app been tested on other devices? Pi-Apps supports most arm linux devices, not exclusively raspberry pi boards.

Also while the app looks cool and useful, it appears to fail a criterion in our [app eligibility guidelines](https://github.com/Botspot/pi-apps/issues/185)
- The app needs to be useful for people who use their Raspberry Pi as an everyday, desktop computer.
  - Apps that are meant to run on a headless Pi will not be accepted. (Rosetta@home, pihole, etc)
  - Apps that are not designed for everyday GUI users won't be accepted. (command-line file servers, webservers, etc)

Now after reviewing the criterion again, I can spot an edge case where an app is headless on the pi, but has a gui webserver for management, thus possibly making it "designed for everyday GUI users". However, this would probably fail the "useful for people who use their Raspberry Pi as an everyday, desktop computer." under most use-cases. Interested in your thoughts on this.

Also I'm concerned about general usefulness. It's hard to find a reason to use this compared to something like RustDesk, which is FOSS, allows incoming connections via local IP address or persistent paired name via free p2p forwarding servers, and offers remote desktop, command-line, and file browser connection modes. Raspberry Pi Connect is another contender, with web-based remote desktop and ssh.

theofficialgman:
Agree with botspot. This fragile software is not needed. His suggestions hold up as well. Cockpit is also a good tool in this vane and available in the apt repos of every system https://cockpit-project.org/

stghecker:
Hi. I just saw the pull request... I'm very happy you actually responded to me. It's something very important to me. I agree with you in some points. Anyways, thank you for all you did, I really appreciate your work with pi-apps. I'm planning to create a new app. See ya! 👋

PR #3047, opened 2026-09-15
Add Fan Controller app

danysus:
A Controller for the Raspberry Pi 5 Active Cooler


theofficialgman:
The fan is a standard interface and already controlled automatically.

PR #3052, opened 2026-09-19
CI: test arbitrary code execution

theofficialgman:


theofficialgman:
tested safe against arbitrary code execution. the workflow is run from the master branch of Botspot/pi-apps, not the PR's workflow, so its safe to run unconditionally. closing.

